PR #4, opened 2017-01-18
anthony-bruscantini

bruscantini:


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,28 @@
 // Find the maximum
 function maxOfTwoNumbers (first, second) {
-
+  if (first > second) return first;
```
f3r: Good job!!

Also, you could consider this:
```
return (first > second ? first : second);
```

PR #13, opened 2017-01-31
huh

Buraburaite:
Daniel Soraluz, Oleh Kolinko, and Javier Cueto got to the bonus content, and successfully checked all valid horizontal combinations (result was between 40 and 50 million). Everything was deleted by an inexperienced mac user navigating in Chrome (Javier, me).


Not-A-Sith-Lord:
lol

PR #33, opened 2017-04-04
bryan-al-noumair

BryanAl-noumair:


starter-code/functions-and-arrays.js
```diff
@@ -106,7 +151,13 @@ console.log(hasDog);
...
 // Counting Repetion
 function howManyTimes (words, word) {
-
+  var count =0;
+   words.forEach(function(element){
+    if(element===word){
+      count += 1;
+    }
+  });
+  return count;
```
tawebbcn: good job, try to get to the bonus part if you have some time

PR #34, opened 2017-04-04
BCN-Functions and Arrays

edtrias:


starter-code/functions-and-arrays.js
```diff
@@ -1,6 +1,10 @@
 // Find the maximum
-function maxOfTwoNumbers (first, second) {
 
+function maxOfTwoNumbers (first, second) {
+  var numList = [first, second];
+  var maxNum = numList.sort().reverse(); //sort the numbers in order and then reverse it
+  maxNum = maxNum[0]; //select the index[0] because it's sorted in descendant order
+  return maxNum;
 }
```
tawebbcn: It seems to complicated. You don't need al those methods. 

  return first > second ? first : second;
}

starter-code/functions-and-arrays.js
```diff
@@ -85,6 +114,12 @@ console.log(uniqued);
 // Finding Elements
 function doesWordExist (wordsArray, word) {
 
+  if (wordsArray.includes(word)) { //includes checks if the word it's within the array
+   return true;
+ } else {
+   return false;
+ }
+
```
tawebbcn: #checked

PR #36, opened 2017-04-05
File with completed exercises: 1-7

michellebro:
Didn't finish but the ones I did do I made sure that I could do them/understand them.


starter-code/functionsandarrays.js
```diff
@@ -0,0 +1,168 @@
...
+sumOfElements = sumOfElements + averageNumbersnumbers[i]}
+
+var averageSum = (sumOfElements / numberOfElements);
+console.log(averageSum);
+
+}
+
+var average = averageNumbers(numbers);
+console.log(average);
+// 6*/
```
tawebbcn: for this part, try to be sure that you are declaring and using the variables on the correct place. The logic is good

starter-code/functionsandarrays.js
```diff
@@ -0,0 +1,168 @@
...
+}
+
+
+var hasMatter = doesWordExist(words, "matter");
+console.log(hasMatter);
+// true
+
+var hasDog = doesWordExist(words, "dog");
+console.log(hasDog);
+// false*/
```
tawebbcn: try to keep working on it, do not hesitate on coming to us to ask for any doubts you may have

PR #69, opened 2017-06-13
BCN: Miguel and Arnau

Mi6u3l:


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,25 @@
...
 
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
 function findLongestWord (words) {
-
+  var longest = 0;
+  for(var i = 0; i<words.length; i++) {
+    var wordLength = words[i].length;
```
tawebbcn: you don't have to declare var wordLength: 

"     if (longest < words[i].length) {
       longest = words[i].length;
    }"

starter-code/functions-and-arrays.js
```diff
@@ -106,7 +132,13 @@ console.log(hasDog);
...
+  var counter = 0;
+  for(var i = 0; i<words.length; i++) {
+    if (words[i] === word) {
+      counter++;
+    }
+  }
+  return counter;
 }
 
 var words = [
```
tawebbcn: #checked. Good job.

PR #72, opened 2017-06-13
AlexFontanellas & Stephanos Theodotou Day 2 Pair Prog.

stephanos7:


starter-code/functions-and-arrays.js
```diff
@@ -123,39 +167,70 @@ var words = [
...
+  }
+  return maxValue;
+
+
+
+  }
 var maxProduct = greatestProduct(matrix);
 console.log(maxProduct);
+
+//57,148,146
```
tawebbcn: # checked.

PR #75, opened 2017-06-13
BCN Rachel & Patri - second day exercise

patriki:


starter-code/functions-and-arrays.js
```diff
@@ -23,27 +40,44 @@ var words = [
...
+
+
 // Calculating a Sum
 function sumArray (array) {
-
+  var sum = 0;
+  for (i = 0; i < array.length; i++){
+    sum += array[i];
+  }
+  return sum;
```
tawebbcn: you could try to use the reduce method:

function sumArray (array) {
  return array.reduce(function (sum, number){
    return sum + number;
  }, 0);
}

starter-code/functions-and-arrays.js
```diff
@@ -129,6 +192,9 @@ console.log(howManyMatter);
 var howManyDog = howManyTimes(words, "dog");
 console.log(howManyDog);
 
+
+
+
```
tawebbcn: Good job! Try the bonus part if you have time :)

PR #77, opened 2017-06-13
arrays-functions Albert+Camille

camdw:


starter-code/functions-and-arrays.js
```diff
@@ -20,21 +20,45 @@ var words = [
...
+
+function findLongestWord (words) {
+  var j = "";
+    for(var i=0; i < words.length; i++){
+        if( words[i].length > j.length) {
+            j = words[i];
+          }
+    }
+
+  return j
```
tawebbcn: try to give variables a name that makes sense, in this case, something like "longestWord"

starter-code/functions-and-arrays.js
```diff
@@ -20,21 +20,45 @@ var words = [
...
 
+
 // Calculate the Average
 function averageNumbers (array) {
-
+  var j = 0;
+  for(var i = 0; i < numbers.length; i++) {
+    j = j + numbers[i];
+  }
+  return j/numbers.length
```
tawebbcn: Your solution is good, but since you created a function to calculate the sum of an array, you could call this function and divide the result by the array.length

starter-code/functions-and-arrays.js
```diff
@@ -158,69 +158,65 @@ var hasDog = doesWordExist(words, "dog");
...
+//   [04,52,08,83,97,35,99,16,07,97,57,32,16,26,26,79,33,27,98,66],
+//   [88,36,68,87,57,62,20,72,03,46,33,67,46,55,12,32,63,93,53,69],
+//   [04,42,16,73,38,25,39,11,24,94,72,18,08,46,29,32,40,62,76,36],
+//   [20,69,36,41,72,30,23,88,34,62,99,69,82,67,59,85,74,04,36,16],
+//   [20,73,35,29,78,31,90,01,74,31,49,71,48,86,81,16,23,57,05,54],
+//   [01,70,54,71,83,51,54,69,16,92,33,48,61,43,52,01,89,19,67,48],
+// ];
+//
+// var maxProduct = greatestProduct(matrix);
+// console.log(maxProduct);
```
tawebbcn: Good job! Try to go for the bonus if you guys have time ;)

PR #79, opened 2017-06-13
bcn PP work Brad and Alex Buchanan all except bonus..for now

bradford-le:
bcn day 2 pp work Brad Le and Alex Buchanan

### DIFF
--- a/starter-code/functions-and-arrays.js
+++ b/starter-code/functions-and-arrays.js
@@ -1,16 +1,26 @@
 // Find the maximum
 function maxOfTwoNumbers (first, second) {
-
+  if(first > second) {
+    return first;
+  } else {
+    return second;
 }
-
+}
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
 function findLongestWord (words) {
-
+  var wordstore= words[0];
+  for(i=0; i<words.length;i+=1) {
+  if (wordstore.length < words[i].length) {
+    wordstore = words[i];
+    console.log("wordstore for greater loop: " + wordstore );
+  } else {
+    console.log("wordstore for else loop: " + wordstore );
+  }}
+return wordstore;
 }
-
 var words = [
   "mystery",
   "brother",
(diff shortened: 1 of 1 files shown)


bradford-le:
Brad Le and Alex Buchanan

starter-code/functions-and-arrays.js
```diff
@@ -1,16 +1,26 @@
...
+  var wordstore= words[0];
+  for(i=0; i<words.length;i+=1) {
+  if (wordstore.length < words[i].length) {
+    wordstore = words[i];
+    console.log("wordstore for greater loop: " + wordstore );
+  } else {
+    console.log("wordstore for else loop: " + wordstore );
+  }}
+return wordstore;
 }
```
tawebbcn: your solution is ok but you could try this with a forEach loop:

  var longest = "";

  words.forEach (function (word) {
    if (word.length > longest.length) {
      longest = word;
    }
  });

  return longest;
}

starter-code/functions-and-arrays.js
```diff
@@ -23,27 +33,40 @@ var words = [
...
 console.log(total);
 
 // Calculate the Average
+var sumTotal = 0;
 function averageNumbers (array) {
-
+for(i=0; i<array.length;i+=1){
+   sumTotal += array[i];
+ }
+ return sumTotal/=array.length;
```
tawebbcn: You could call the function you did before to calculate the sum and then divide it for the array.length, you can also use the reduce method for calculating the sum.
Your solution is good though :)

starter-code/functions-and-arrays.js
```diff
@@ -129,9 +175,15 @@ console.log(howManyMatter);
...
 // Bonus Quest
 function greatestProduct (matrix) {
-
+  var maxproduct;
+  var centervalue;
+  for(i=0;i<matrix.length;i+=1){
+    centervalue = Math.max.apply(Math, matrix);
+  }
+  return maxproduct;
 }
```
tawebbcn: Good job! Try to go for the bonus! ;)

PR #80, opened 2017-06-13
BCN-Jonathyn (JB) Tellez, Julius Souminen

JB-Tellez:
JB and Julius' pair programming results.

### DIFF
--- a/starter-code/functions-and-arrays.js
+++ b/starter-code/functions-and-arrays.js
@@ -1,5 +1,11 @@
 // Find the maximum
 function maxOfTwoNumbers (first, second) {
+  
+  if (first > second) {
+    return first;
+  } else {
+    return second;
+  }
 
 }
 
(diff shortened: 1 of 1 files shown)


starter-code/functions-and-arrays.js
```diff
@@ -34,6 +54,12 @@ console.log(total);
 
 // Calculate the Average
 function averageNumbers (array) {
+  
+
+  var sum = sumArray(array);
+  
+  return sum / array.length;
```
tawebbcn: You can skip the var declaration and do something like:

starter-code/functions-and-arrays.js
```diff
@@ -158,4 +318,4 @@ var matrix = [
 ];
 
 var maxProduct = greatestProduct(matrix);
-console.log(maxProduct);
+console.log(maxProduct);
```
tawebbcn: Good job!

PR #92, opened 2017-06-27
Paris - Mathieu and François

MacSlymo:


starter-code/functions-and-arrays.js
```diff
@@ -20,9 +16,23 @@ var words = [
...
   "crackpot"
 ];
+
+var lgth = 0;
+
+function findLongestWord (words) {
+ for (i=0; i<words.length; i++){
+   if (words[i].length > lgth){
+     lgth = words[i].length;
+     longest = words[i];
```
posva: remember to declare this variable inside the function

starter-code/functions-and-arrays.js
```diff
@@ -20,9 +16,23 @@ var words = [
   "orchard",
   "crackpot"
 ];
+
+var lgth = 0;
+
+function findLongestWord (words) {
+ for (i=0; i<words.length; i++){
+   if (words[i].length > lgth){
+     lgth = words[i].length;
```
posva: You don't need to store the length of the longest word because you can access it with `longest.length`

starter-code/functions-and-arrays.js
```diff
@@ -20,9 +16,23 @@ var words = [
   "orchard",
   "crackpot"
 ];
+
+var lgth = 0;
+
+function findLongestWord (words) {
+ for (i=0; i<words.length; i++){
```
posva: don't forget the `var` before the `i`

PR #93, opened 2017-06-27
Paris vincent

vincentmoulines:


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,18 @@
...
 
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
 function findLongestWord (words) {
-
+  var longest = words[0]
+  words.forEach(function(word){
+    longest = (word.length > longest.length) ? word : longest
```
posva: in this case the ternary is not necessary because we don't need the `else` statement

starter-code/functions-and-arrays.js
```diff
@@ -25,7 +29,11 @@ console.log(longest);
 
 // Calculating a Sum
 function sumArray (array) {
-
+  var sum = 0;
+  array.forEach(function(elmt){
```
posva: it's ok to use generic names for generic functions but try to make variable name readable

starter-code/functions-and-arrays.js
```diff
@@ -34,16 +42,25 @@ console.log(total);
...
 }
 
 var numbers = [2, 6, 9, 10, 7, 4, 1, 9];
 var average = averageNumbers(numbers);
 console.log(average);
 
+// Calculating a Sum words
+function sumArrayWords (array) {
+  var sum = 0;
+  array.forEach(function(elmt){
```
posva: here you should have used `word` as the value in the function
it's ok to copy paste 😉 but make sure to adapt the content

starter-code/functions-and-arrays.js
```diff
@@ -63,7 +80,13 @@ console.log(averageLength);
 
 // Unique Arrays
 function uniquifyArray (array) {
-
+  result = [];
```
posva: where is the var ? 😨

starter-code/functions-and-arrays.js
```diff
@@ -84,7 +107,12 @@ console.log(uniqued);
 
 // Finding Elements
 function doesWordExist (wordsArray, word) {
-
+  if(wordsArray.indexOf(word)  === -1){
+    return false
+  } else {
+    return true
+  }
+  return (wordsArray.indexOf(word)  === -1);
```
posva: This will never get executed

starter-code/functions-and-arrays.js
```diff
@@ -84,7 +107,12 @@ console.log(uniqued);
 
 // Finding Elements
 function doesWordExist (wordsArray, word) {
-
+  if(wordsArray.indexOf(word)  === -1){
```
posva: directly do 
```js
return wordsArray.indexOf(word) > -1
```
because the result is already a boolean

starter-code/functions-and-arrays.js
```diff
@@ -106,7 +134,13 @@ console.log(hasDog);
 
 // Counting Repetion
 function howManyTimes (words, word) {
-
+  var iteration = 0;
```
posva: `count` would be a more appropriate name 🙂

PR #110, opened 2017-08-15
BCN - Mark & Anca

markvertra:
@ancadelmar


tawebbcn:
Good job! :+1: 

Tips:
 - forEach (use it!)
 - Variable's name
 - Never divide by 0
 - Bonus can be improved

PR #112, opened 2017-08-15
BCN - Philip Wisner & Frances Morales

philipwisner:
@frances0688


tawebbcn:
Good job! :+1: 

Tips:
 - Longest word exercise!!! (var inside function, return and length)
 - forEach (use it!)
 - Never divide by 0
 - Try bonus when you feel confident

PR #113, opened 2017-08-16
BCN will-thorp gabriele-dolfi

willjthorp:


tawebbcn:
Good job! :+1: 

Tips:
 - DRY (don't repeat yourself): Reuse code!
 - forEach (use it!)
 - Longest word exercise: you need to return the word.
 - Bonus can be improved

PR #114, opened 2017-09-05
PAR - David PINHAS & Nicolas BARTHEZ - Iteration 1

nicolasBarthez:
Done

### DIFF
--- a/starter-code/functions-and-arrays.js
+++ b/starter-code/functions-and-arrays.js
@@ -1,6 +1,9 @@
 // Find the maximum
 function maxOfTwoNumbers (first, second) {
-
+  if (first>=second) {
+    return first;
+  }
+    return second ;
 }
 
 var largest = maxOfTwoNumbers(2, 6);
@@ -8,7 +11,15 @@ console.log(largest);
 
 // Finding Longest Word
 function findLongestWord (words) {
-
+  var result = words[0];
+  for(var i=1; i<words.length;i++) {
+    if (result.length > words[i].length) {
+      result = result;
+    } else {
+      result = words[i];
+    }
+  }
+  return result;
 }
 
 var words = [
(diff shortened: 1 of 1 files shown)


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,25 @@
...
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
 function findLongestWord (words) {
-
+  var result = words[0];
+  for(var i=1; i<words.length;i++) {
+    if (result.length > words[i].length) {
+      result = result;
```
yacinehmito: You don't really need that line.

starter-code/functions-and-arrays.js
```diff
@@ -34,7 +49,11 @@ console.log(total);
 
 // Calculate the Average
 function averageNumbers (array) {
```
yacinehmito: You can reuse `sumArray`.

starter-code/functions-and-arrays.js
```diff
@@ -43,7 +62,11 @@ console.log(average);
 
 // Array of Strings
 function averageWordLength (array) {
```
yacinehmito: This is the fastest solution. Another would be to create an array of word lengths and then apply `averageNumbers`on it.

starter-code/functions-and-arrays.js
```diff
@@ -63,7 +86,15 @@ console.log(averageLength);
 
 // Unique Arrays
 function uniquifyArray (array) {
-
+  var newTab =[];
+  for (i=0;i<array.length ; i++){
+    if (newTab.indexOf(array[i])===-1){
+      newTab.push(array[i]);
+    } else {
+      continue;
```
yacinehmito: You don't need that line. `else`is not mandatory.

starter-code/functions-and-arrays.js
```diff
@@ -105,8 +136,17 @@ var hasDog = doesWordExist(words, "dog");
 console.log(hasDog);
 
 // Counting Repetion
-function howManyTimes (words, word) {
-
+function howManyTimes (array, word) {
+  var newTab =[word];
```
yacinehmito: This will work fine but `newTab`will always have the same word in it. You could instead use an integer and increment it accordingly; it's simpler.

PR #116, opened 2017-09-05
PAR - Davoult Thibaut - Lefremova Liudmyla

thibautdavoult:


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,24 @@
 // Find the maximum
 function maxOfTwoNumbers (first, second) {
-
+  Math.max(first, second);
```
yacinehmito: This is cheating 😛 
Nice catch though.

starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,24 @@
...
 }
 
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
 function findLongestWord (words) {
```
yacinehmito: Original answer; I like it, although it is not the most performant. Try to do it in a single loop.

starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,24 @@
...
-
+  var newArray = [];
+  var longestWord;
+  words.forEach(function(element, i){
+    var newElement = element.length;
+    newArray.push(newElement);
+    });
+  var maxArray = Math.max(...newArray);
```
yacinehmito: You're little wizards already!

starter-code/functions-and-arrays.js
```diff
@@ -43,7 +58,13 @@ console.log(average);
 
 // Array of Strings
 function averageWordLength (array) {
```
yacinehmito: This is the solution I expected. Reusing code is almost always the best move. Keep it up!

starter-code/functions-and-arrays.js
```diff
@@ -63,7 +84,14 @@ console.log(averageLength);
 
 // Unique Arrays
 function uniquifyArray (array) {
-
+  var uniqueArray = [];
+  array.forEach(function(element) {
+    var uniqueElement;
+    var index = element.indexOf();
```
yacinehmito: It seems you got stuck on this. You need to use `indexOf` on the array, not the element.

LiudmilaE: I found the solution in the evening)))
https://repl.it/Kh8w/4

yacinehmito: Awesome! Please add those changes to your repo by doing a new commit and pushing. This will automatically update the PR. Once you're done I will validate your work ;)

PR #117, opened 2017-09-05
stuart-gardner, thomas-beguin

stuartogardner:


starter-code/functions-and-arrays.js
```diff
@@ -61,9 +99,26 @@ var words = [
...
 
+
+
+// ABOVE ALL SEEMS TO BE CORRECT
+
+
+
 // Unique Arrays
+
 function uniquifyArray (array) {
```
yacinehmito: You are trying to remove duplicates from the array. This approach is very hard. An easier one would be to create a new array in which you add elements you haven't seen before.

PR #118, opened 2017-09-05
PAR - Hubert Saint Olive & Jad Challita - Iteration 2, done.

jadch:


starter-code/functions-and-arrays.js
```diff
@@ -130,8 +183,127 @@ var howManyDog = howManyTimes(words, "dog");
...
+        var product = matrix[row + 1][column] * matrix[row - 1][column] * matrix[row][column - 1] * matrix[row][column + 1]
+        if (product > max_product) {
+          max_product = product
+        }
+      }
+  }}
+  return max_product
+}*/
+
 function greatestProduct (matrix) {
```
yacinehmito: This is the most elegant solution you could have ended up with. Very nice work!
Notice however that if a cell holds a zero, the associated condition inside the `if`s will be false. To be perfectly correct, you can check for `undefined` by doing for example `if(matrix[row - 1] === undefined)`instead of `if(matrix[row - 1])`.

starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,28 @@
...
+  }
 }
 
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
+
 // Finding Longest Word
 function findLongestWord (words) {
+  var max_length = 0
```
yacinehmito: This is a very good solution. However you don't need `max_length` because you can always get it from `max_word.length`.

starter-code/functions-and-arrays.js
```diff
@@ -23,27 +37,39 @@ var words = [
...
+  return sumArray(array) / array.length
 }
 
 var numbers = [2, 6, 9, 10, 7, 4, 1, 9];
 var average = averageNumbers(numbers);
 console.log(average);
 
+
 // Array of Strings
 function averageWordLength (array) {
```
yacinehmito: This the solution I expected. Great work!

starter-code/functions-and-arrays.js
```diff
@@ -82,9 +116,16 @@ var words = [
 var uniqued = uniquifyArray(words);
 console.log(uniqued);
 
+
+
 // Finding Elements
 function doesWordExist (wordsArray, word) {
-
+  for (var i = 0; i < wordsArray.length; i++) {
+  if (wordsArray[i] === word) {
```
yacinehmito: Be careful with the indentation. You should use an editor that fixes this for you.

starter-code/functions-and-arrays.js
```diff
@@ -104,9 +145,21 @@ console.log(hasMatter);
 var hasDog = doesWordExist(words, "dog");
 console.log(hasDog);
 
+
+
 // Counting Repetion
 function howManyTimes (words, word) {
```
yacinehmito: This will work fine but it will modify the array `words` in place. By doing `wordsBis = words` you are not copying the array; you are just giving it a new name.

A simpler approach would simply to loop on the array and count the number of times the current element is equal to `word`.

starter-code/functions-and-arrays.js
thibautdavoult: Would it be possible to solve this one with a `forEach`? I'm not sure how that'd work with the `wordsArray[i]` being used here.

Also, would using `array.includes(word)`

yacinehmito: `wordsArray.includes(word)`shoud work. Did you return it?

For the `forEach`solution, please post your attempt here as a comment and I will guide you toward the solution if you didn't succeed.

PR #119, opened 2017-09-05
PAR-Frederic Jacquette & Frederic magnin - Iteration 1

fredjack:


starter-code/functions-and-arrays.js
```diff
@@ -20,67 +20,117 @@ var words = [
   "orchard",
   "crackpot"
 ];
+
+function findLongestWord(words) {
```
yacinehmito: This works fine. However you don't need `num`because you can simply to `long.length`.

starter-code/functions-and-arrays.js
```diff
@@ -20,67 +20,117 @@ var words = [
...
-function averageNumbers (array) {
+var numbers = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(array){
+  var avg = total/array.length;
+  return avg;
 }
 
-var numbers = [2, 6, 9, 10, 7, 4, 1, 9];
+function sumArray(array){
```
yacinehmito: You just copy-pasted the same function as above.

yacinehmito: Don't give it the same name as a previously defined function.

starter-code/functions-and-arrays.js
```diff
@@ -20,67 +20,117 @@ var words = [
...
+   "bring",
+   "sharp",
+   "playground",
+   "poison",
+   "communion",
+   "simple",
+   "bring"
+ ];
+
+ var words2 = [];
```
yacinehmito: This is the solution. However you should but it in the body of the `doesWordExist`function.

PR #120, opened 2017-09-05
PAR - OMBLINE GILLIOT - Özgür - done except bonus

ombline:


starter-code/functions-and-arrays.js
```diff
@@ -63,7 +88,14 @@ console.log(averageLength);
 
 // Unique Arrays
 function uniquifyArray (array) {
```
yacinehmito: This is an original solution. It uses stuff we haven't seen yet. I like it but for the sake of it you could try a simpler one.

starter-code/functions-and-arrays.js
```diff
@@ -84,7 +116,8 @@ console.log(uniqued);
 
 // Finding Elements
 function doesWordExist (wordsArray, word) {
-
+var n =  words.includes(word);
```
yacinehmito: This is the good kind of cheating ;)

PR #121, opened 2017-09-05
PAR - Reshma / Elodie 

Lalo8:


starter-code/functions-and-arrays.js
```diff
@@ -20,142 +20,84 @@ var words = [
...
-  "sharp",
-  "playground",
-  "poison",
-  "communion",
-  "simple",
-  "bring"
-];
-var uniqued = uniquifyArray(words);
-console.log(uniqued);
+function sumArray(array){
```
yacinehmito: You forgot to change the name of the function. Copy-pasting is treacherous; I suggest people to avoid it.

starter-code/functions-and-arrays.js
```diff
@@ -20,142 +20,84 @@ var words = [
...
+var sum=0 ;
+var counter=0;
```
yacinehmito: You don't really need a counter. `i`is already one.

PR #124, opened 2017-09-22
MIA Marisa Bruno

marisabruno:
Exercise without the bonus completed

### DIFF
--- a/more-dom-practice/index.html
+++ b/more-dom-practice/index.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


dsoraluz:
Great job on getting all the way to the Bonus Quest.

This paired programming exercise was meant to really get your brain thinking of how we can solve problems with the use of arrays and functions. Arrays and functions never go away in programming so it is important to understand them and how they can be used to help solve our programming problems.

Math and programming also go hand in hand so you should have gotten to flex your arithmetic muscles with these exercises.

And again paired programming relies heavily on effective communication between all in the group. It is important to learn to communicate both effectively and efficiently.

PR #142, opened 2017-10-17
BCN- ELOI&FELICE

xxxelf:
@eloigil

### DIFF
--- a/starter-code/functions-and-arrays.js
+++ b/starter-code/functions-and-arrays.js
@@ -1,14 +1,18 @@
 // Find the maximum
-function maxOfTwoNumbers (first, second) {
-
-}
+function maxOfTwoNumbers(first, second) {}
 
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
-function findLongestWord (words) {
-
+function findLongestWord(words) {
+  var long = "";
+  words.forEach(function(word) {
+    if (word > long) {
+      long = word;
+    }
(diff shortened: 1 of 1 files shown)


xxxelf:
@eloigil

xxxelf:
@eloigil

starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,18 @@
...
-}
+function maxOfTwoNumbers(first, second) {}
 
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
-function findLongestWord (words) {
-
+function findLongestWord(words) {
```
tawebbcn: Good!

tawebbcn:
/*
Good job!

My two most important improvements are:
- to avoid spaghetti code. you should always try to wrap block of codes into functions. Then you just call the function instead of duplicate the code block multiple times. You call averageNumbers(anotherNumberArray), averageNumbers(oneNumberArray), etc..
- try to always `return` something from a function, if you later want to console log it, you do it after.

example:
*/

// Calculate the Average
```
var result = numbers.reduce(function(accumulator, currentValue) {
  return accumulator + currentValue;
});
var average = result / numbers.length;
console.log(average);
```

// to ->

```
  var result = array.reduce(function(accumulator, currentValue) {
    return accumulator + currentValue;
  });
  return result / array.length;
}

console.log(average);

```

// Isak

PR #144, opened 2017-10-17
BCN - Jean and Brian

jnothomb:
@brianboyCode


tawebbcn:
Very Good Job!

I just found just a small thing to improve. No need to declare an variable if you know will immediately return it. 
```
function maxOfTwoNumbers(first, second) {
  var theBiggest = Math.max(first, second);
  return theBiggest;
}
```

// can be simplified to
```
function maxOfTwoNumbers(first, second) {
  return Math.max(first, second);
}
```

PR #145, opened 2017-10-17
BCN - Clara Ameller - Tiago

ClaraAmeller:
@tiago1999


tawebbcn:
Very good job!
Feedback:
- Don't forget to remove or / comment-out all unnecessary console.log()'s. Like `console.log(sorted[index] + " is repeated!");`
`console.log("New array with no repetitions: ");`
Before submit.

PR #146, opened 2017-10-17
BCN - Constantinos & Misha

cyenis:
Misha @MythicalMichael


tawebbcn:
Good job!

Feedback: 
-You almost got it.

// Finding Elements
```
/*    function doesWordExist (wordsArray, word) {
    wordsarray.forEach(function (word) {
    wordsarray.includes()
    };
    });
    return
    var words = [
      "machine",
      "subset",
      "trouble",
      "starting",
      "matter",
      "eating",
      "truth",
      "disobedience"
    ];
    var hasMatter = doesWordExist(words, "matter");
    console.log(hasMatter);
    console.log(hasDog);
*/
```
Includes are both a String method and an Array Method. In this case, the best way it to use the Array method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) instead of the String method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes). Why it is better using the Array method is you do not need to loop over the array with an ForEach method to solve it.

Solution:
```
// Finding Elements
  if (wordsArray.includes(word)) {
    return true;
  } else {
    return false;
  }
}
var words = [
  "machine",
  "subset",
  "trouble",
  "starting",
  "matter",
  "eating",
  "truth",
  "disobedience",
];
var hasMatter = doesWordExist(words, "matter");
console.log(hasMatter);
console.log(hasDog);
```

PR #147, opened 2017-10-17
BCN - Dafne - Jordi

joordiii:
@dafneolca

### DIFF
--- a/starter-code/functions-and-arrays.js
+++ b/starter-code/functions-and-arrays.js
@@ -1,14 +1,31 @@
 // Find the maximum
-function maxOfTwoNumbers (first, second) {
+maxOfTwoNumbers (3,5);
 
+function maxOfTwoNumbers (first, second) {
+  var result;
+  if (first>second){
+  result=first;
+  }
+  else if (first<second) {
+  result=second;
+  }
+  else {
+    result = 'both numbers are equal';
+  }
+  return result;
 }
 
-var largest = maxOfTwoNumbers(2, 6);
-console.log(largest);
-
 // Finding Longest Word
-function findLongestWord (words) {
-
+function findLongestWord (palabras) {
+  var longestWord='';
+  palabras.forEach(function (word){
+    var wordElement = word;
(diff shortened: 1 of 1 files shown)


tawebbcn:
**Good job!**

**Feedback:**
**You can improve this function, I suggest two solutions. (you only need to find the max, even to they are equal.)**

```
function maxOfTwoNumbers1(first, second) {
  var result;
  if (first > second) {
    result = first;
  } else if (first < second) {
    result = second;
  } else {
    result = "both numbers are equal";
  }
  return result;
}
```

// Better to this
```
function maxOfTwoNumbers2(first, second) {
  if (first > second) {
    return first;
  } else {
    return second;
  }
}
```
// Or this
```

function maxOfTwoNumbers3(first, second) {
  return Math.max(first, second);
}
```

```
console.log("maxOfTwoNumbers1: ", maxOfTwoNumbers1(3, 5));
console.log("maxOfTwoNumbers2: ", maxOfTwoNumbers2(3, 5));
console.log("maxOfTwoNumbers3: ", maxOfTwoNumbers3(3, 5));
```
**Feedback:**
**You can improve this function, I suggest below solutions.**

```
// Finding Longest Word
var words = [
  "mystery",
  "brother",
  "aviator",
  "crocodile",
  "pearl",
  "orchard",
  "crackpot",
];

// Finding Longest Word
function findLongestWord(palabras) {
  var longestWord = "";
  palabras.forEach(function(word) {
    var wordElement = word;
    var wordLength = wordElement.length;
    if (wordLength > longestWord.length) {
      longestWord = word;
    }
  });
  return longestWord;
}

console.log("first solution: ", findLongestWord(words));
```

```
// You can shorter this down like this -->
function findLongestWordNew(palabras) {
  var longestWord = "";
  palabras.forEach(function(word) {
    if (word.length > longestWord.length) {
      longestWord = word;
    }
  });
  return longestWord;
}

console.log("second solution: ", findLongestWordNew(words));

```

PR #149, opened 2017-10-17
BCN - Cristina and Dominik

dlasch089:
@crismoreno

### DIFF
--- a/starter-code/functions-and-arrays.js
+++ b/starter-code/functions-and-arrays.js
@@ -1,17 +1,18 @@
 // Find the maximum
-function maxOfTwoNumbers (first, second) {
-
+function maxOfTwoNumbers(first, second) {
+  if(first > second){
+    return first;
+  }else if(first < second){
+    return second;
+  }else {
+    return "they are the same!";
+  }
 }
 
-var largest = maxOfTwoNumbers(2, 6);
-console.log(largest);
+var largest = maxOfTwoNumbers(9, 1);
+console.log("The largest number is: ",largest);
 
-// Finding Longest Word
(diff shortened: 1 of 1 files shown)


tawebbcn:
**Very good job!**

Tips:
- you can use `!newList.includes(el)` instead of `newList.includes(el) === false`
- Make sure you check if array.length is larger then 0.

```
  var total = 0;
  if (array.length > 0) {
    array.forEach(function(el) {
      total += el;
    });
    return total / array.length;
  }
  return 0;
}

var numbers2 = [2, 6, 9, 10, 7, 4, 1, 9];
var average = averageNumbers(numbers2);
console.log("the average of the array is: ", average);

```

- Use can use includes here.

```
var words = [
  "machine",
  "subset",
  "trouble",
  "starting",
  "matter",
  "eating",
  "truth",
  "disobedience",
];

// Finding Elements
  var exist = false;
  for (var ix = 0; ix < words.length; ix++) {
    if (words[ix] === word) {
      exist = true;
      break;
    }
  }
  return exist;
}

var hasMatter = doesWordExist(words, "subset");
console.log("The word exists: ", hasMatter);

console.log("The word exists: ", hasDog);


```
**Updated code:**

```
// Finding Elements
function doesWordExistUpdated(wordsArray, word) {
  if (wordsArray.includes(word)) {
    return true;
  } else {
    return false;
  }
}

var hasMatter = doesWordExistUpdated(words, "subset");
console.log("The word exists: ", hasMatter);

var hasDog = doesWordExistUpdated(words, "dog");
console.log("The word exists: ", hasDog);

```

PR #150, opened 2017-10-17
BCN - Elvin & David

3lv27:
@otitlam


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,25 @@
 // Find the maximum
 function maxOfTwoNumbers (first, second) {
-
+ if ( first < second) {
+   console.log(second);
+ } else {
+   console.log(first);
```
tawebbcn: You are showing the result in the console instead of returning it. That way, the variable 'largest' in line 10 is going to be undefined (because you are not returning anything). Usually you want your functions to return something, that way you can use that value ahead.

starter-code/functions-and-arrays.js
```diff
@@ -34,7 +49,11 @@ console.log(total);
 
 // Calculate the Average
 function averageNumbers (array) {
-
+  var total = 0;
+  numbers.forEach(function(element){
+    total += element;
```
tawebbcn: Do you have another function that sums all the values in an array? Then, maybe you can use it here instead of repeating code.

starter-code/functions-and-arrays.js
```diff
@@ -64,6 +87,15 @@ console.log(averageLength);
 // Unique Arrays
 function uniquifyArray (array) {
 
+  var uniqueArr = [];
+
+  array.forEach(function(word) {
+    if (uniqueArr.includes(word)) {
```
tawebbcn: This function is returning an empty array because uniqueArr don't include anything at the beginning. Try this:
```
  var uniqueArr = [];
  array.forEach(function(word) {
    if (!uniqueArr.includes(word)) {
      uniqueArr.push(word);
    }
  });
  return uniqueArr;
}
```
You can read this "if (!uniqueArr.includes(word))" as "if uniqueArr DON'T include the word, then...".

starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,25 @@
...
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
-function findLongestWord (words) {
-
+function findLongestWord (word) {
+
+ var longestWord = '';
+ words.forEach(function(word) {
```
tawebbcn: WATCH OUT! You are using the variable words (outside the function) instead of the parameter of the function (word, at line 14). You will see it clearer in the next comment.

starter-code/functions-and-arrays.js
```diff
@@ -24,8 +35,12 @@ var longest = findLongestWord(words);
 console.log(longest);
 
 // Calculating a Sum
-function sumArray (array) {
-
+function sumArray (arr) {
+var total = 0;
+numbers.forEach(function(element){
```
tawebbcn: WATCH OUT! You are using the variable numbers (outside the function) instead of the parameter of the function (arr, at line 38).

starter-code/functions-and-arrays.js
```diff
@@ -34,7 +49,11 @@ console.log(total);
 
 // Calculate the Average
 function averageNumbers (array) {
-
+  var total = 0;
+  numbers.forEach(function(element){
```
tawebbcn: Same problem, use the function's parameters.

PR #151, opened 2017-10-17
BCN - Desiree Lucas Cristian

dvillool:
@gominolas
@lacazeto


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,34 @@
...
 function maxOfTwoNumbers (first, second) {
-
+  if (first>second){
+    console.log ("first");
+  }
+  else if (second>first){
+    console.log ("second");
+  }
+  else {
+    console.log("the numbers are the same");
```
tawebbcn: You are showing the result in the console instead of returning it. That way, the variable 'largest' in line 14 is going to be undefined (because you are not returning anything). Usually you want your functions to return something, that way you can use that value ahead.

starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,34 @@
...
-function findLongestWord (words) {
 
+function findLongestWord (words) {
+    var biggest="";
+    for ( var ix=0; ix < words.length; ix++){
+      if ( biggest.length < words[ix].length){
+        biggest=words[ix];
+      }
+    }
+    for ( var j=0; j<words.lenght; j++){
```
tawebbcn: You don't need this second for. I understand you use it to display more than one result (if there is a tie for the longest word), but the problem again is that you MUST return the result instead of showing it in the console. In this case, if you want to display more than one word, you will need an array. If the function's signature is expecting an output of a value, returning an array could be a problem.

starter-code/functions-and-arrays.js
```diff
@@ -20,12 +40,16 @@ var words = [
...
+
 var longest = findLongestWord(words);
 console.log(longest);
 
 // Calculating a Sum
 function sumArray (array) {
-
+  var totalSum=0;
+  array.forEach(function(number){
+      totalSum+=number;
```
tawebbcn: Good job with the forEach. However, you have forgotten return something.

starter-code/functions-and-arrays.js
```diff
@@ -34,7 +58,11 @@ console.log(total);
 
 // Calculate the Average
 function averageNumbers (array) {
-
+  var totalSum=0;
+  array.forEach(function(number){
```
tawebbcn: Do you have another function that sums all the values in an array? Then, maybe you can use it here instead of repeating code.

starter-code/functions-and-arrays.js
```diff
@@ -43,9 +71,14 @@ console.log(average);
 
 // Array of Strings
 function averageWordLength (array) {
-
+  var total=0;
+  array.forEach(function(number){
```
tawebbcn: Good job. However, be careful with variable's names. Don't call it number if it is a word.

starter-code/functions-and-arrays.js
```diff
@@ -84,7 +122,12 @@ console.log(uniqued);
 
 // Finding Elements
 function doesWordExist (wordsArray, word) {
-
+  var biggest="";
+  for ( var ix=0; ix < words.length; ix++){
+    if ( biggest.length < words[ix].length){
```
tawebbcn: You don't want to find the biggest word, you want to check if that word is in the array.

starter-code/words.html
```diff
@@ -0,0 +1,22 @@
+<script>
```
tawebbcn: What is this? Why are you including a script in another html file?

PR #158, opened 2017-10-18
MIA Joshua Morales & Robert E Gonzalez

JomoraZero:


dsoraluz:
Great job on getting all the way to the array of strings exercise. 

This paired programming exercise was meant to really get your brain thinking of how we can solve problems with the use of arrays and functions. Arrays and functions never go away in programming so it is important to understand them and how they can be used to help solve our programming problems. 

Math and programming also go hand in hand so you should have gotten to flex your arithmetic muscles with these exercises.

And again paired programming relys heavily on effective communication between all in the group. It is important to learn to communicate both effectively and efficiently.

PR #165, opened 2017-10-21
PAR - Vivian & Luca

VivianSolide:


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,24 @@
...
+  }
+  else {
+    console.log("equals!");
+  }
 }
 
-var largest = maxOfTwoNumbers(2, 6);
-console.log(largest);
+var largest = maxOfTwoNumbers(6, 6);
+console.log(largest
```
ta-web-paris: You forgot the parenthesis here

PR #185, opened 2017-11-03
Cristina Bertrand WEB PT1017 MADRID

Cristina-bertrand:
"FEAT: full exercise except the bonus part, and the last function for count have an error that I don't know how to solve it"


starter-code/functions-and-arrays.js
```diff
@@ -1,14 +1,20 @@
...
+return Math.max(first,second);
 }
 
 var largest = maxOfTwoNumbers(2, 6);
 console.log(largest);
 
 // Finding Longest Word
 function findLongestWord (words) {
-
+var longest = "";
```
ta-web-mad: intenta meter una sangría para que sea más legible

PR #194, opened 2018-01-06
find-longest-word-little-fix

HectorBlisS:
The description of the test says: "returns the length of the longest one" but the test is expecting for the string itself, i change the Readme file wich is the quck fix, but may be could be good to change the spec file to not confuse the student.
@f3r @papuarza


papuarza:
You are right! Perfect! Thanks!!

PR #211, opened 2018-01-09
BCN - Filipe

ncoder92:
@DTDK

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,16 @@
+'use strict';
+
 // Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
+  } else {
+    return num2;
+  }
+}
+
 
 // Finding Longest Word
 var words = [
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,16 @@
+'use strict';
+
 // Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
```
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +23,49 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(words) {
+  var longestWord = '';
+  if (words.length > 0) {
```
tawebbcn: This is a way of returning undefined when the array is empty. Another one is:
```
var longestWord = words[0]
```
This way, you asign to your comparator (longestWord) the first array's element. If the array is empty, longestWord will be 'undefined' and the function will return 'undefined'. Using an internal element for comparision solve some tricky problems (e.g. finding a minimum).
```
  var longestWord = words[0];

      longestWord = words[i];
    }
  }
  return longestWord;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +23,49 @@ var words = [
...
+}
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(numbersAvg) {
+  var sum = 0;
+  if(numbersAvg.length > 0){
+  for (var i = 0;i < numbersAvg.length; i++){
```
```
  var sum = 0;
  if(numbersAvg.length > 0) {
    return sumArray(numbersAvg) / numbersAvg.length;
  } else{
    return undefined;
  }
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -60,6 +136,17 @@ var wordsFind = [
   'disobedience'
 ];
 
+function doesWordExist(arrayWords, wordSearch) {
+  function searchWord(element) {
```
tawebbcn: I understand why you have done this, but not get used to define functions inside another functions if you don't have a very good reason for doing it.

PR #212, opened 2018-01-09
BCN - David & Sheng

theRealSheng:
@davidalexteixeira


starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,11 @@
 // Find the maximum
 
+function maxOfTwoNumbers(one, two) {
+  if(one > two) return one;
+  if(two > one) return two;
+  if(one == two) return one;
```
tawebbcn: You don't need three conditions here. If the first number is not greater than the second one, you can return the second one, because either the second one is greater or they are equal (and then doesn't matter wich number you return):
```
function maxOfTwoNumbers(one, two) {
  if (one > two) {
    return one;
  } else {
    return two;
  }
}
```
Or, if you prefer it:
```
function maxOfTwoNumbers(one, two) {
  if (one > two) return one;
  return two;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +17,32 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(array) {
+  
+  if (array == undefined) return 0;
```
tawebbcn: This line is not necessary.

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +17,32 @@ var words = [
...
+  return array.reduce((a,b)=>a+b,0);
+}
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(array) {
+  let len = array.length;
+  return len === 0 || array.includes(NaN) ? undefined : array.reduce((a,b)=>a+b,0)/len;
```
```
return len === 0 || array.includes(NaN) ? undefined : sumArray(array)/len;
```

starter-code/src/functions-and-arrays.js
```diff
@@ -74,6 +126,19 @@ var wordsCount = [
   'disobedience',
   'matter'
 ];
+
+function howManyTimes(array, word) {
```
tawebbcn: Here you are cheating writing the code that passes the tests but doesn't work with every case (try it with an array with the word repeated two times). A possible solution would be:
```
  var appearances = 0;
  if (array.length === 0) {
    return undefined;
  }
    if (array[i] === word) {
      appearances++;
    }
  }
  return appearances;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -98,3 +163,28 @@ var matrix = [
   [20, 73, 35, 29, 78, 31, 90, 1, 74, 31, 49, 71, 48, 86, 81, 16, 23, 57, 5, 54],
   [1, 70, 54, 71, 83, 51, 54, 69, 16, 92, 33, 48, 61, 43, 52, 1, 89, 19, 67, 48]
 ];
+
+function greatestProduct(array){
```
tawebbcn: We will talk about this function when you have time. It's a good start, but has some problems. Anyways, the tests for this one are not enough.

starter-code/src/functions-and-arrays.js
theRealSheng: I will in the future. I just thought I had to rewrite it.

starter-code/src/functions-and-arrays.js
theRealSheng: Yes, I think I was so focus in just fixing the testing situation that missed the whole point of the problem. Fixing it.

PR #213, opened 2018-01-09
BCN Danielle & Jana

jjanastinn:
@danielleleb

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -5,6 +5,8 @@ function maxOfTwoNumbers(value1, value2) {
     return value1;
   } else if (value2 > value1){
     return value2;
+  } else {
+    return value1;
   }
 }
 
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,18 @@
 // Find the maximum
 
+function maxOfTwoNumbers(value1, value2) {
+  if (value1 > value2){
+    return value1;
+  } else if (value2 > value1){
```
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,10 +24,47 @@ var words = [
   'crackpot'
 ];
 
+
+function findLongestWord(words) {
+  var longestWord = '';
```
tawebbcn: This function will fail returning undefined when the array is empty. You can solve this checking if the array is empty and returning undefined then. A better way is asign to your comparator (longestWord) the first array's element. If the array is empty, longestWord will be 'undefined' and the function will return 'undefined'. Using an internal element for comparision solve some tricky problems (e.g. finding a minimum).
```
  var longestWord = words[0];

      longestWord = words[i];
    }
  }
  return longestWord;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,10 +24,47 @@ var words = [
...
+}
+
+numbers.reduce(function sumArray(p,n){
+  return p + n;
+},0);
+
+
+var counter = 0;
+
+function sumArray(p,n) {
```
tawebbcn: This solution is valid, but you need to put the variable counter inside the function. The previous reduce has some problems with the parameters (we will talk about that later). Don't forget to delete the previous sumArray, you can't have two functions with the same name.

PR #214, opened 2018-01-09
BCN Giovanni and Josep

JoniWanKenobi:
@Jsalvat

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -91,6 +91,17 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray(arr){
+  return arr.reduce(function(acc,val){
+    if(acc.indexOf(val)===-1){
+      acc.push(val);
+    }
+    return acc;
+  },[])
+}
+
+console.log('uniquified array is: ',uniquifyArray(wordsUnique));
+
 // Finding Elements
 var wordsFind = [
   'machine',
@@ -103,6 +114,16 @@ var wordsFind = [
   'disobedience'
 ];
 
+function doesWordExist(arr,word){
+  return arr.reduce(function(acc,val){
+    if(val===word){
+      acc=true;
+    }
+    return acc;
+  },false)
+}
+
+console.log('check word exists: ', doesWordExist(wordsFind,"mathinondjsh"));
 // Counting Repetion
 var wordsCount = [
   'machine',
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,5 +1,11 @@
 // Find the maximum
 
+function maxOfTwoNumbers(x,y){
+  return Math.max(x,y);
+}
+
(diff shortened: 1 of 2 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,11 @@
 // Find the maximum
 
+function maxOfTwoNumbers(x,y){
+  return Math.max(x,y);
```
tawebbcn: :dark_sunglasses:

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +17,40 @@ var words = [
   'crackpot'
 ];
 
+
+function findLongestWord(arr){
+  return arr.reduce(function(acc,val){
```
tawebbcn: Good aproach, but there is a problem. This code doesn't work for an empty array (is one of the tests). You can start checking if the array is empty (and then returning undefined) or include an initial value of undefined:
```
function findLongestWord(arr){
  return arr.reduce(function(acc,val){
    if(val.length>acc.length){
      acc=val;
    }
    return acc;
  }, undefined)
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +17,40 @@ var words = [
...
+}
+
+console.log(findLongestWord(words));
+
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(arr){
+  return arr.reduce(function(acc,val){
```
tawebbcn: Same problem as before.

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +17,40 @@ var words = [
...
+    return acc;
+  })
+}
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arr){
+  var sum = sumArray(arr);
```
tawebbcn: Great job reusing the function.

starter-code/src/functions-and-arrays.js
```diff
@@ -33,6 +65,19 @@ var wordsArr = [
   'palace'
 ];
 
+function averageWordLength(arr){
+  var lengthSum= arr.reduce(function(acc,val,index,a){
```
tawebbcn: You don't need index, neither a.

starter-code/src/functions-and-arrays.js
```diff
@@ -60,6 +116,16 @@ var wordsFind = [
   'disobedience'
 ];
 
+function doesWordExist(arr,word){
+  return arr.reduce(function(acc,val){
```
tawebbcn: This is great, but maybe in this case would be more interesting using a for that you can stop as soon as you encounter the value. That way, if the word is at the beginning of the array, you don't iterate over every element.

PR #216, opened 2018-01-09
BCN - Rodrigo and Pere

rodrigodaoud:
@pereamoros


starter-code/src/functions-and-arrays.js
```diff
@@ -1,23 +1,54 @@
...
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+//var avg = sum / numbersAvg.length;
+
+
+
+function averageNumbers(numbersAvg){
+  if(numbersAvg.length === 0 ){
+    return undefined
+  }
+   var sum = numbersAvg.reduce(function(a,b) {
```
```
var sum = sumArray(numbersAvg);
```

PR #217, opened 2018-01-09
BCN Alla Luis

luisacherrera:
@allarom

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -13,11 +13,34 @@ var words = [
 
 // Calculating a Sum
 
-var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -13,11 +13,34 @@ var words = [
 
 // Calculating a Sum
```
tawebbcn: What about the previous exercises?

starter-code/src/functions-and-arrays.js
```diff
@@ -13,11 +13,34 @@ var words = [
...
 // Calculate the Average
 
-var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+//var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+
+function averageNumbers(numbers) {
+
+  var total =0;
+
+  for (var i=0; i < averageNumbers.length; i++) {
```
```
total = sumArray(numbers);
```

PR #218, opened 2018-01-09
BCN - Matt & Martin: JASMINE

mattmegarry:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,27 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2){
+    return num1;
+  } else if (num1 < num2){
```
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,27 @@
...
+function findLongestWord(words){
+ var maxLength = 0;
+ var maxWordIndex = 0;
+  for (var i = 0; i < words.length; i++){
+    if (words[i].length > maxLength) {
+      maxLength = words[i].length;
+      maxWordIndex = i;
+    }
+ }
+ return words[maxWordIndex];
```
tawebbcn: Two things to keep in mind:
 - You don't need two variables. You can store the longest word in a variable called 'longestWord' and access its length with 'longestWord.length'.
 - This function will fail returning undefined when the array is empty. You can solve this checking if the array is empty and returning undefined then. A better way is asign to your comparator (longestWord) the first array's element. If the array is empty, longestWord will be 'undefined' and the function will return 'undefined'. Using an internal element for comparision solve some tricky problems (e.g. finding a minimum).
```
  var longestWord = words[0];

      longestWord = words[i];
    }
  }
  return longestWord;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -12,14 +33,44 @@ var words = [
...
+
+  return total;
+}
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
 // Calculate the Average
+function averageNumbers (numbersAvg){
+  if (numbersAvg.length !== 0){
+  return sumArray(numbersAvg)/numbersAvg.length;
```
tawebbcn: Good job reusing your function.

starter-code/src/functions-and-arrays.js
```diff
@@ -12,14 +33,44 @@ var words = [
...
 // Array of Strings
+function averageWordLength(wordsArr){  
+ var totalCharSum = 0;
+  for (i = 0; i < wordsArr.length; i++){
+    totalCharSum += wordsArr[i].length;
+  }
+  
+  if(wordsArr.length !== 0){
+  return totalCharSum/wordsArr.length;
+  }
```
tawebbcn: Else, you have to return undefined. Better, start checking it.

PR #222, opened 2018-01-09
BCN Stephie James

kissy89:
@jsavory-94

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,23 +1,73 @@
+
 // Find the maximum
 
+function maxOfTwoNumbers (var1, var2) {
+  if (var1 > var2){
+    return var1;
+  } else {
+    return var2;
+  }
+}
+
+var max = maxOfTwoNumbers(10, 6);
+console.log(max);
+
 // Finding Longest Word
+
+function findLongestWord (words) {
+  var longestWord="";
+  if (words.length === 0){
+    return;
+  }
+
+  for(var i=0; i < words.length; i++){
+    if (longestWord.length < words[i].length){
+        longestWord = words[i];
+    }else if(longestWord.length === words[i].length){
+        var morelongestWords = [];
+        morelongestWords.push(words[i] , longestWord);
+    }
+  }
+  return longestWord;
+}
+
 var words = [
-  'mystery',
-  'brother',
-  'aviator',
-  'crocodile',
-  'pearl',
-  'orchard',
-  'crackpot'
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "holidaysarefine",
+  "pearl",
+  "orchard",
+  "crackpot"
 ];
+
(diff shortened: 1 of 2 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,23 +1,73 @@
...
+}
+
+var max = maxOfTwoNumbers(10, 6);
+console.log(max);
+
 // Finding Longest Word
+
+function findLongestWord (words) {
+  var longestWord="";
+  if (words.length === 0){
```
tawebbcn: This is a way of returning undefined when the array is empty (you need to add 'undefined' to the 'return' in line 20, though). Another one is:
```
var longestWord = words[0]
```
This way, you asign to your comparator (longestWord) the first array's element. If the array is empty, longestWord will be 'undefined' and the function will return 'undefined'. Using an internal element for comparision solve some tricky problems (e.g. finding a minimum).
```
  var longestWord = words[0];

      longestWord = words[i];
    }
  }
  return longestWord;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -1,23 +1,73 @@
...
+    return;
+  }
+
+  for(var i=0; i < words.length; i++){
+    if (longestWord.length < words[i].length){
+        longestWord = words[i];
+    }else if(longestWord.length === words[i].length){
+        var morelongestWords = [];
+        morelongestWords.push(words[i] , longestWord);
```
tawebbcn: Good job thinking about possibilities, but this approach has two problems:
 - It only works if there are two longests words. If you find a third one, you will lose the previous results and store only the first and third longest words.
 - You only return the first longest word (line 31), so this piece of code (26-29) never changes the output.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,23 +1,73 @@
...
+
+var allNumbers = sumArray(numbers);
+console.log(allNumbers);
+
+
 // Calculate the Average
 
-var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+function averageNumbers(numbers) {
+  var average = sumArray(numbers)/numbers.length;
```
tawebbcn: Good job reusing your functions.

starter-code/src/functions-and-arrays.js
```diff
@@ -60,6 +110,20 @@ var wordsFind = [
...
 ];
 
+function doesWordExist(words, searchWord){
+  for (var i = 0; i< words.length; i++){
+    if (searchWord === words[i]){
+    return true;
+    } else {
+    return false;
+    }
+    break;
```
tawebbcn: This code only will work properly if the word is in first place. Either way, will return false before iterating all over the array (because line 118). Furthermore, the break does nothing. This is a possible solution:
```
function doesWordExist(words, searchWord) {
    if (words[i] === searchWord) {
      return true;
    }
  }
  return false;
}
```

PR #266, opened 2018-03-20
PAR Lucas & Pierre

lucasm-git:
I used only regular for loops because I didn't quite get how forEach works...

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,6 +1,24 @@
-// Find the maximum
+function maxOfTwoNumbers(nb1, nb2) {
+  if(nb1 > nb2) {
+    return nb1;
+  }
+  else {
+    return nb2;
+  }
+}
 
-// Finding Longest Word
+/******************************************************/
+
+function findLongestWord(array) {
+  var longestWord = array[0];
+  for(var i=0; i < array.length; i++) {
+    if(array[i].length > longestWord.length) {
+      longestWord = array[i];
+    }
+  }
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Basically forEach is a method on the Array prototype. It iterates through each element of an array.
When a for loop is iterating threw in a limited environment.

Check the MDN documentation, it's a really good way to dive a bit into this notions, and if you have any question just ask :) 

Otherwise, good job ! :)

starter-code/src/functions-and-arrays.js
```diff
@@ -74,6 +167,9 @@ var wordsCount = [
   'disobedience',
   'matter'
 ];
+
+/******************************************************/
+
 // Bonus Quest
```
ta-web-paris: You can use this piece of code for the bonus quest :) 

  var big = 0;
  for (var i = 0; i < matrix.length; i++) {
    var left, right, up, down;
    var temp = 1;
    for (var j = 0; j < matrix.length; j++) {
      if (i === 0) {
        up = 1;
      } else if (i === matrix.length - 1) {
        down = 1;
      } else if (j === 0) {
        left = 1;
      } else if (j === matrix.length - 1) {
        right = 1;
      } else {
        left = matrix[i - 1][j];
        right = matrix[i + 1][j];
        up = matrix[i][j - 1];
        down = matrix[i][j + 1];
      }
      temp = left * right * up * down;
      if (temp > big) {
        big = temp;
      }
    }
  }
  return big;
}

// var matrix

console.log(maxProduct);

PR #269, opened 2018-03-20
PAR Brianna & Jeremie

ycrtw93:
We didn't have enough time to finish all exercises. Had some trouble debugging some small details. Also, it looks like we're making the average functions more complicated than they need to be.


ta-web-paris:
There is some beginners mistakes but it's not bad. :clap: 
Do not hesitate to draw your issues before coding.
Help a lot!

PR #270, opened 2018-03-20
PAR Mostafa & Clement

M-Gadd:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,12 @@
-// Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1>=num2) {
+    return num1;
+  } else if (num2>num1) {
+    return num2;
+  }
```
ta-web-paris: You must return an else statement in order to return num1.

starter-code/src/functions-and-arrays.js
```diff
@@ -11,15 +17,68 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(words) {
+  var longestWord = "";
+
+  if (words.length === 0) {
```
ta-web-paris: not reliable I guess...

ta-web-paris:
I'll send you the correction.
But it's pretty well done.
Keep rockin'

PR #274, opened 2018-03-20
PAR Maggie & Louis

mbbaber:
We stopped at uniqueify array exercise.


ta-web-paris:
Good job!
This exercise is very hard but you master the loops, this is the most important.

PR #275, opened 2018-03-20
PAR Leo -W & JB

Jbmouterde:
We didn't had the time to finish the part " Array of Strings" .

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,5 +1,11 @@
 // Find the maximum
-
+function maxOfTwoNumbers (number1, number2){
+  if (number1>number2){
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,15 +17,50 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord (words){
+  var longestWordLength = words[0].length;
+  var longestWord ;
+  words.forEach (function (word){
```
ta-web-paris:
```js
  var longest = words[0];

    if (words[i].length > longest.length) {
      longest = words[i];
    }
  }
  return longest;
}
```
I prefer for loop :)

PR #276, opened 2018-03-20
PAR Sami & Bertrand & Billy

Sami-Az:
we got a little bit confused with console.log and return results
(Finding Longest Word - findLongestWord returns the word with an 1-word array)
we didnt have enough time to go over the rest


ta-web-paris:
console.log is just here for print infos
return give the value
https://www.codecademy.com/en/forum_questions/50107c79dcd8fd0002003498

PR #282, opened 2018-03-20
BCN - Jorge Terol

jorgeterol:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,442 @@
...
-  'pearl',
-  'orchard',
-  'crackpot'
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
```
tawebbcn: You didn't had to change. It's actually preferable to use single quotes in JS and double quotes in HTML

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,442 @@
...
 ];
 
+function findLongestWord(arrayLong) {
+  var longestWord = "";
+
+  for (var i = 0; i <= arrayLong.length - 1; i++) {
```
tawebbcn: You can simplify a bit here: ``` for (var i = 0; i < arrayLong.length; i++)```

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,442 @@
...
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray (arraySum){
+
+ var summedArray = arraySum.reduce(function(a,b){
+    return a+b;
+  })
+ return summedArray;
+}
```
tawebbcn: Nice one! :smile:

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,442 @@
...
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers (arrayAverage){
+
+ return sumArray(arrayAverage)/arrayAverage.length
```
tawebbcn: Re-using functions :+1:

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,442 @@
...
+  "trouble",
+  "starting",
+  "matter",
+  "eating",
+  "truth",
+  "disobedience"
 ];
 
+  function averageWordLength(arrayAverageWordLength){
+    var arrayWithLengthWords = arrayAverageWordLength.map(function(a){
```
tawebbcn: Naming FTW :+1: If you didn't knew map, how would you do it?

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,442 @@
...
+  "matter",
+  "truth",
+  "disobedience",
+  "matter"
 ];
+
+function uniquifyArray (arrayUniquify){
+  var finalArray = [];
+  for (var j=0; j<=arrayUniquify.length;j++){
+    if (arrayUniquify.indexOf(arrayUniquify[j])==j){
```
tawebbcn: Here more logically you would want to do the indexOf in the finalArray. If the indexOf returns -1 then it does not exist, so push it into the array.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,442 @@
...
+    
+  }
+  return false;
+}
+
+function howManyTimes(arrayToCount, wordToFind){
+  var howManyTimes=0;
+  for (var y=0; y<=arrayToCount.length-1; y++){
+    if (arrayToCount[y]==wordToFind){
+      howManyTimes += 1;
```
tawebbcn: You can also use howManyTimes++

PR #285, opened 2018-03-20
BCN-Jenny

Fantasmy:
COuldn't finish it, brain is fried....

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,5 +1,14 @@
 // Find the maximum
 
+function findMaxTwoNumbers (num1,num2){
+  if (num1>num2) {
+    return num1
+  }
+  else if (num1<num2) {
+    return num2;
+  }
+}
+
 // Finding Longest Word
 var words = [
   'mystery',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,13 +20,31 @@ var words = [
...
+  var total=0;
+  for(var i in numbers) { total += numbers[i]; }
 
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+var total=0;
+for(var i in numbersAvg) {total += numbersAvg [i]} {
+  var avg = total/numbersAvg.length;
+}
```
tawebbcn: The for in loop is recommended to be used with objects, here's what you could do:
```
function sumCalculation(arrayNumbers) {
  var total = 0;
  arrayNumbers.forEach(function (number) {
    total += number;
  });
  return total;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -33,6 +60,9 @@ var wordsArr = [
   'palace'
 ];
 
+averageWordLength = wordsArr.join('').length/wordsArr.length;
```
tawebbcn: Smart :+1: :+1:

PR #286, opened 2018-03-20
BCN-RUBEN-AGUAYO

rubenaguayo:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,18 @@
-// Find the maximum
+// Find the maximum{
+
+function max(num1,num2) {
+  if (num1>num2) {
+    console.log(num1);
+  } else if (num2>num1) {
+    console.log(num2);
+  } else {
+    console.log("num1 = num2");
```
tawebbcn: Here the variables won't show because they are inside the double quotes

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +23,42 @@ var words = [
...
+add(numbers);
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function avg(list) {
+  for (var sum=0, i=0; i < numbersAvg.length; i++) {
+    sum += list[i];
+  }
```
tawebbcn: This is a nice scenario where you could use the above add function to avoid repeating code

PR #287, opened 2018-03-21
BCN-MIHAI-COSTANTEA

cosmihai:


starter-code/src/functions-and-arrays.js
```diff
@@ -10,14 +17,44 @@ var words = [
...
 ];
+function findLongestWord(array) {
+  var longestWord;
+  if(array.length!==0){
+    var maxLength=array[0].length;
+    longestWord=array[0];
+    array.forEach(function(word){
+      if(word.length>maxLength){
+        maxLength=word.length;
+        longestWord=word;
```
tawebbcn: Here you could initialize the longestWord var as an empty string and then compare that length with the one with the word you are checking in the array

starter-code/src/functions-and-arrays.js
```diff
@@ -10,14 +17,44 @@ var words = [
...
+  return sum;
+}
 
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+function averageNumbers(array) {
+  var average;
+  if(array.length!==0) {
+    average=sumArray(array)/array.length;
```
tawebbcn: Good work on reusing the function :+1:

starter-code/src/functions-and-arrays.js
```diff
@@ -47,6 +95,29 @@ var wordsUnique = [
...
+  var itemToCompare;
+  for(var i=0; i<array.length; i++){
+    itemToCompare=array[i];
+    
+    for(var j=i+1; j<array.length; j++) {
+      if(itemToCompare===array[j]) {
+        doubleElementIndex.push(array.indexOf(array[j]));
+      }
+    }
+  }
```
tawebbcn: Loops are expensive. Here's a simpler solution
```
  var newArray = [];
  if (array.length === 0) {
    return undefined;
  }
      newArray.push(array[i]);
    }
  }
  return newArray;
}
```

PR #288, opened 2018-03-21
BCN - Sebastian Nugnez

Sebachowa:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,39 @@ var words = [
   'crackpot'
 ];
 
-// Calculating a Sum
+function findLongestWord(words) {
+  var word;
+  var length = 0;
+  words.forEach(function(string){
+    if (string.length > length) {
```
tawebbcn: Why not start with ``` var word = ''``` and then compare the length below? :wink:

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,39 @@ var words = [
...
+};
 
+// Calculating a Sum
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-// Calculate the Average
+function sumArray(array) {
+  return array.reduce(function (acum, current) {
+    return acum + current;
+  }, 0);
```
tawebbcn: Gooood Stuff :+1:, in this case, you didn't need to start the accumulator at 0. In the first run, the accumulator will be the position 0 of the array and the current, the position 1.

starter-code/src/functions-and-arrays.js
```diff
@@ -33,6 +65,15 @@ var wordsArr = [
   'palace'
 ];
 
+function averageWordLength(array) {
+  if (array.length !== 0) {
+    var newArray = array.map(function(string) {
+      return string.length;
+    });
+  return averageNumbers(newArray); //I just taked advantage of the function created in the previous exercise (line 45)
```
tawebbcn: Oh yeah! :+1: :+1:

starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +89,15 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray(array) {
+  if (array.length !== 0) {
+    var uniqArray = array.filter(function(string, index) {
+			return array.indexOf(string) === index;
```
tawebbcn: Nicee :+1:

PR #291, opened 2018-03-21
BCN-Miriam-Aparicio

MiriamAparicio:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,186 @@
...
-  'pearl',
-  'orchard',
-  'crackpot'
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
```
tawebbcn: Single quotes are preferred in JS and Double quotes in HTML

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,186 @@
...
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(arrayOfNumbers) {
+  var sum;
+  if (arrayOfNumbers.length === 0) {
+    return 0;
+  } else {
+    sum = (acumulator, currentValue) => acumulator + currentValue;
+    return arrayOfNumbers.reduce(sum);
```
tawebbcn: Interesting.. :+1:

PR #292, opened 2018-03-21
BCN -Jordi Calzada

jordicalzada:


starter-code/src/functions-and-arrays.js
```diff
@@ -74,6 +122,17 @@ var wordsCount = [
   'disobedience',
   'matter'
 ];
+
+function howManyTimes(array, word){
+  if(array.length === 0){ return false;}
```
tawebbcn: Better keep the structure:
```
if(array.length === 0){
    return false;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -74,6 +122,17 @@ var wordsCount = [
...
+  for(var i=0; i<array.length; i++){
+    if(array[i] == word){
+      var times =+ 1;
```
tawebbcn: better define times out of the for. Now you are declaring it every time.

tawebbcn: It's +=

PR #293, opened 2018-03-21
BCN-Germán-Rodríguez

GermanRodrickson:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +19,52 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(words) {
+  var sumWords = ""
```
tawebbcn: you are missing the semicolons ' ; ' in all the document

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +19,52 @@ var words = [
...
+  for(var i = 0; i < words.length; i++) {
+    if(words[i].length > sumWords.length) {
+      sumWords = words[i]
+    }
+  }
+  if(sumWords === ""){
+    return undefined
```
tawebbcn: :+1:

PR #294, opened 2018-03-21
BCN - Lennie

lenniez:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,16 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1>num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
```
tawebbcn: Is this moment, mean that `num2` is greater or equal than `num1` so we can just put `return num2`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,15 +21,48 @@ var words = [
...
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-// Array of Strings
+function averageNumbers (numbersAvg) {
+  var total = [0];
+  for (var j = 0; j < numbersAvg.length; j++) {
+    total += numbersAvg[j];
+  }
```
tawebbcn: You can reuse your `sumArray`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,15 +21,48 @@ var words = [
...
+  if (total > 0) {
+    return total/numbersAvg.length;
+  } else if (total == 0) {
```
tawebbcn: `total === 0`

tawebbcn: You don't need this. If `total > 0` return the average. Else don't do anything. Javascript will return `undefined` as default

starter-code/src/functions-and-arrays.js
```diff
@@ -33,6 +76,18 @@ var wordsArr = [
   'palace'
 ];
 
+function averageWordLength (wordsArr) {
+  var allLetters = [0];
+  for (var k = 0; k < wordsArr.length; k++) {
+    allLetters += wordsArr[k].length;
+  }
+  if (allLetters == 0) {
```
tawebbcn: You should check this in the beginning of the function and better with something like `wordsArr.length > 0`

starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +103,15 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray (wordsUnique) {
+  var wordsNoDuplicates = Array.from(new Set(wordsUnique));
```
tawebbcn: `Set`! 👍 
But you know how a `Set` works? 👀

starter-code/src/functions-and-arrays.js
```diff
@@ -60,7 +124,13 @@ var wordsFind = [
   'disobedience'
 ];
 
-// Counting Repetion
+var searchTerm = 'trouble';
+
+function doesWordExist (wordsFind, searchTerm) {
+  return (wordsFind.includes(searchTerm));
```
tawebbcn: `includes` 👍

PR #296, opened 2018-03-21
BCN Samantha Ferreira

samanthaferreira:
Only got the first one done

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,8 @@
-// Find the maximum
+// Find the maximu
+
+function maxOfTwoNumbers(num1, num2) {
+  return Math.max(num1, num2)
+}
 
 // Finding Longest Word
 var words = [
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -9,7 +13,28 @@ var words = [
...
-];
+]; 
+function findLongestWord(words){
+  for(var i = 0; i < words.length; i++) {
+		var longest = words[i].length;
+ 		if (words[i].length > longest) {
+ 			longest = words[i]
+		}
+ 	} 
+ 	return longest
```
tawebbcn: indentation

starter-code/src/functions-and-arrays.js
```diff
@@ -9,7 +13,28 @@ var words = [
   'pearl',
   'orchard',
   'crackpot'
-];
+]; 
+function findLongestWord(words){
+  for(var i = 0; i < words.length; i++) {
+		var longest = words[i].length;
```
tawebbcn: You can't assign the var longest inside the for, cause you'll be reasigning it every iteration. I leave you a good example of how it should be done!
```
    var longest = "";
    for(var i = 0; i < words.length; i++) {
        if(words[i].length > longest.length) {
            longest = words[i]
        }
    }
    if(longest === ""){
        return undefined;
    } else {
        return longest;
    }
}
```

PR #297, opened 2018-03-21
BCN - Pere

Interna1ta:
first 4exercices fine, the fifth fails

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,15 @@
 // Find the maximum
+function maxOfTwoNumbers (num1, num2) {
+  var result = '';
+  if (num1>num2) {
+    result = num1;
+  } else if (num1<num2 ){
+    result = num2;
+  } else {
+    result = num1;
+  }
+  return result;
+}
 
 // Finding Longest Word
 var words = [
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,15 @@
 // Find the maximum
+function maxOfTwoNumbers (num1, num2) {
+  var result = '';
+  if (num1>num2) {
+    result = num1;
+  } else if (num1<num2 ){
```
tawebbcn: Is this moment, mean that `num2` is greater or equal than `num1` so we can just put `result = num2`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +22,38 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord (words) {
+  var result = words[0];
+  for (i=1; i<words.length; i++){
```
tawebbcn: `for (var i = 1;  i< words.length; i++){`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +22,38 @@ var words = [
...
+  }
+  return result;
+}
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(numbersAvg) {
+  var resultAvg = sumArray(numbersAvg) / numbersAvg.length
```
tawebbcn: 👍

PR #298, opened 2018-03-21
BCN - Marketa

Hrabenka7:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,35 @@
...
+function maxOfTwoNumbers(a,b) {
+  if(a > b){
+    //console.log(a + "is the larger number.");
+    return a;
+  }else if(a < b){
+    //console.log( b + "is the largest number")
+    return b;
+  }else{
+    //console.log("Numbers are equal.")
+  } return a,b;
```
tawebbcn: 3 things:
- the last return `return a, b` is outside the `else`
- `return a,b` is a interesting statement. It will only return the last expression. The before ones it will be evaluated.
- the logic can be simplified to:
```
function maxOfTwoNumbers (a, b) {
  if (a > b) {
   return a;
  }
  return b;
}
```
Because if the `if` is true, it will return and stop the function. Else it will return the another one.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,35 @@
...
+
+  if (array.length > 0) {
+    for (let i = 0; i < array.length; i++) {
+      if (array[i].length > longestWordLength){
+        longestWordLength = array[i].length;
+        longestWord = array[i];
+      }    
+    } return longestWord;
+  }
+  else
```
tawebbcn: Care with the `{}`.

tawebbcn: This `else return` is not need it since javascript will return `undefined` as default

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +40,43 @@ var words = [
...
+// Calculate the Average
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(array) {
+  var sum = 0;
+  var average = 0;
+  if (array.length > 0) {
+    for (var i = 0; i < array.length; i++) {
+      sum = sum + array[i];
+    }
```
tawebbcn: Here you should reuse you `sumArray` function

PR #299, opened 2018-03-21
BCN-Dauv 

dauvbafna:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,13 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2){
+  if(num1>num2){
+    return num1;
+  }
+    else{
```
tawebbcn: Indentation!

starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,13 @@
...
+      return num2;
```
tawebbcn: Since you have a `return` in the first `if`, if the condition is true, it will stop the function in that moment. So we don't need that else a put only `return num2`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,13 +20,43 @@ var words = [
...
 
+function findLongestWord(words){ 
+  var longestWordTemp = ""; // string to save the current longest word
+for(var i= 0; i<words.length; i++)
+{
+  if(words[i].length > longestWordTemp.length)
+  {
+    longestWordTemp = words[i];
+  }
+}
```
tawebbcn: Indentation!

starter-code/src/functions-and-arrays.js
```diff
@@ -11,13 +20,43 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(words){ 
+  var longestWordTemp = ""; // string to save the current longest word
+for(var i= 0; i<words.length; i++)
```
tawebbcn: Spacing on the condition!

starter-code/src/functions-and-arrays.js
```diff
@@ -47,6 +97,18 @@ var wordsUnique = [
   'simple',
   'bring'
 ];
+function uniquifyArray(wordsUnique)
+{
+  for (var i=0;i<wordsUnique.length;i++) {
+    for (var j=i+1;j<wordsUnique.length;j++) {
+        if (wordsUnique[i]==wordsUnique[j]){ 
+            wordsUnique.splice(j,1);
```
tawebbcn: This is dangerous since `splice` method change the `length` of the array.  So can have problems in the loop.
If you are curious you can ask us about this.

starter-code/src/functions-and-arrays.js
```diff
@@ -59,6 +121,16 @@ var wordsFind = [
   'truth',
   'disobedience'
 ];
+var wordToFind = "trouble";
+function doesWordExist(wordsFind,wordToFind){
+  for (var i=0;i<wordsFind.length;i++) {
+        if (wordsFind[i]==wordToFind){ 
+            return true;
+        }
```
tawebbcn: Indentation!

PR #301, opened 2018-03-21
done

fredrik62:
didnt have time this morning to complete the assignment, will review the methods more this weekend,

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,6 +1,26 @@
 // Find the maximum
 
-// Finding Longest Word
+
+
+/*
+function maxOfTwoNumbers(num1, num2) {
+       if (num1 > num2) {
+         alert("num1 is larger than num2");
+          } else if (num2 > num1) {
+            alert("num2 is larger than num1");
+              } else if (num1 === num2) {
+                 alert("num1 and num2 are equal");
+                  }
+                    }
+
+                        maxOfTwoNumbers(5,19);
+
+*/
+
+
+
+// i will work on this during the weekend i am aware of some of the issues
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +31,60 @@ var words = [
   'crackpot'
 ];
 
-// Calculating a Sum
 
+// Finding Longest Word
+function findLongestWord(str)
+{
+  var array1 = str.match(/\w[a-z]{0,}/gi);
```
tawebbcn: You should not need this since on the instructions said that `findLongestWord` accepts an array as argument

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +31,60 @@ var words = [
...
+
+
+
+
+// Calculating a Sum
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+var sum = 0;
+for (var i = 0; i < numbers.length; i++) {
+  sum += numbers[i];
+}
```
tawebbcn: You should have this inside a function

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +31,60 @@ var words = [
...
 
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+
+function averageNumbers(a) {
+  var sum1 = 0;
+  for (var j = 0; j < a.length; j++) {
+    sum1 += a[j];
```
tawebbcn: You should reuse your `sumArray` function

PR #302, opened 2018-03-21
JS function Fabain 

fscheugenpflug:
not to about it


starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,17 @@
 // Find the maximum
 
+var num1;
+var num2;
+function maxOfTwoNumbers (num1, num2){
```
tawebbcn: You don't have to declare `num1` `num2`. 
`maxOfTwoNumbers (num1, num2)` this parameters (num1, num2) are placeholders. You can name it as you want for using inside the function's scope.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,17 @@
...
+  if(num1>num2){
+    console.log(num1);
+  } else if (num1<num2){
+    console.log(num2);
+  } else{
+    console.log('The numbers are equal!');
+  }
```
tawebbcn: This also can be written like:
```
  if (num1 > num2) {
+    console.log(num1);
+  } else {
+    console.log(num2);
+  }
```
If `num1 > num2` we show `num1`. Else means that `num2` is either greater or equal to `num1` . So we just can show `num2`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +23,37 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord (words){
+  for ( i = 0; i < words.length; i ++){
+    lengthNumber += words[i].length;
+    return Math.max(lengthNumber)
+  }
```
tawebbcn: For this exercise you should:
- create a variable `longestWord` for saving the longest word
- on each for iteration, compare the current word's length and change if longest
- after the `for`, return `longestWord`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +23,37 @@ var words = [
...
+}
+
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray (numbers){
+  var total = 0;
+  for( ix = 1; i < numbers.length - 1; ix++ ) {
```
tawebbcn: If you have `i < numbers.length - 1` you will go throw the loop but missing the last value of the array

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +23,37 @@ var words = [
...
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers (numbers){
+  var total = 0;
+  for( ix = 1; i < numbers.length - 1; ix++ ) {
+  total += Number(numbers[ix])
+  }
```
tawebbcn: Since you have `sumArray` function, you can reused here

starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +90,22 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray (wordsUnique){
+  for ( i = 0; i < wordsUnique.length; i++){
+    for ( j = i+1 ; j < wordsUnique.length - 1; j++){
+      if (wordsUnique[i] == wordsUnique[j]){
+        uniquifyArray.splice (j,j);
```
tawebbcn: How splice works is:
`splice(index, howMuch)` 
- index: array's position
- howMuch: how much elements you want remove

So you should have `uniquifyArray.splice (j,1);`

But this is dangoures since you are changing the `.length` of the array. So the loop will fail in some point. 
We saw how to do it today. If you have question, ask us.

PR #314, opened 2018-04-03
FTWD Miami Osse Laborde

olaborde:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,10 +26,69 @@ var words = [
...
+
+
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+
+function sumArray(){
+
+  var reducer = (accumulator, currentValue) => accumulator + currentValue;
```
MarcosPalaciosIII: this function did not work the way you wanted it to. in order to use reduce you would do something like this
`
sumArray(array) {
return array.reduce(function (currentValue, accumulator) {
return currentValue + accumulator;
  }, 0)
}
`

starter-code/src/functions-and-arrays.js
```diff
@@ -11,10 +26,69 @@ var words = [
...
+ console.log("longest word is: ", findLongestWord(words));
+
+
+
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+
+function sumArray(){
```
MarcosPalaciosIII: this function did not work the way you wanted it to. in order to use reduce you would do something like this
`
sumArray(array) {
return array.reduce(function (currentValue, accumulator) {
return currentValue + accumulator;
  }, 0)
}
`

MarcosPalaciosIII: this function did not work the way you wanted it to. in order to use reduce you would do something like this
`
sumArray(array) {
return array.reduce(function (currentValue, accumulator) {
return currentValue + accumulator;
  }, 0)
}
`

MarcosPalaciosIII:
you were on the right track but just needed to start by just doing a return 0 if the array is empty. When working with jasmine, pay close attention to the instructions of what jasmine wants. They tend to lead in the right direction but I know they can be confusing. Good job keep working hard

PR #315, opened 2018-04-03
FTWD -Miami- Rawan Alfaiz

rawanalfaiz:


starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +100,13 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray(wordsUnique){
+  var uniqueNames = [];
+
```
MarcosPalaciosIII: you were on the right path here. You just ha to finish the function with something like this 
`
  array.map (function (word) {
    if (newArray.indexOf(word) == -1)
      newArray.push(word);
  });

  return newArray;
}
`

MarcosPalaciosIII:
you were on the right track and looks like with a little more time you could have finished the lab. Keep up the hard work.

PR #317, opened 2018-04-03
mia-jens-functions and arrays

jkai7:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,10 +19,27 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(wordsArray){
```
ianizaguirre: Nope. You are not supposed to hardcode this find longest word function with something like "foo"? 
This function was supposed to find the longest word from a list.

Here is the solution to this part, so you can compare the steps needed:

```js
function findLongestWord(wordsArr) {
  var longest = wordsArr[0];
  var candidate;

  for (var i = 1; i < wordsArr.length; i++) {
    candidate = wordsArr[i];

    if (candidate.length > longest.length) {
      longest = candidate;
    }
  }
  return longest;
}

var longest = findLongestWord(words);
console.log(longest);
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,10 +19,27 @@ var words = [
...
+ }
+ return longestWord;
+}
+findLongestWord("foo");
+
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(){
```
ianizaguirre: Empty Function? You never completed this step. Here is the solution for this step, so you can check it out and make sense of how it works:

```js
// Calculating a Sum
function sumArray(array) {
  var sum = 0;
    sum += array[x];
  }
  return sum;
}

console.log(total);
```

ianizaguirre:
You only completed about 20% of this assignment. If you want help understand how to solve these problems, Slack me, and we can work on them together.

Try to apply this feedback on your next project.
If you have any questions on my feedback, just send me a message on Slack. 
Aside from that, good job and keep up the good work! 😁

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #319, opened 2018-04-03
FTWD Miami: Ryan Martin

ryancraigmartin:
Got up to the Unique Arrays challenge.

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,12 @@
 // Find the maximum
+function maxOfTwoNumbers(numOne, numTwo){
+  if(numOne > numTwo){
+    return numOne;
+  }
+  else{
+    return numTwo;
+  }
+}
 
 // Finding Longest Word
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -33,68 +70,79 @@ var wordsArr = [
...
-  'starting',
-  'matter',
-  'eating',
-  'matter',
-  'truth',
-  'disobedience',
-  'matter'
-];
-// Bonus Quest
+// // Unique Arrays
```
MarcosPalaciosIII: for finding a unique array you can try something like this 
`
  var newArray = [];

  array.map (function (word) {
    if (newArray.indexOf(word) == -1)
      newArray.push(word);
  });

  return newArray;
}
`

MarcosPalaciosIII:
you were on the right path and seems with a little more time you could have finished. Just keep up the good work and remember you can always google search things you do not understand or need help with if one of us is not available to help you.

PR #322, opened 2018-04-03
MIA // Eleyni // lab-javascript-functions-and-arrays

erodrig1010:
Functions and arrays exercise, initial commit.

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -14,55 +14,68 @@ function maxOfTwoNumbers(numberOne, numberTwo) {
 
 // Finding Longest Word
 
-function findLongestWord() {
-  
-  var words = [
-    'mystery',
-    'brother',
-    'aviator',
-    'crocodile',
-    'pearl',
-    'orchard',
-    'crackpot'
-  ];
-  
+var words = [
+  'mystery',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,23 +1,68 @@
...
+  var longest= "";
+  
+  for (i=0; i<words.length; i++){
+      if (words[i].length > longest.length){
+        longest= words[i];
+      }
+  }
+  
+  words.forEach(function(words){
+  });
```
sandrabosk: We really don't need these 2 lines.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,23 +1,68 @@
...
 
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+var sum = numbers.reduce(sumArray, 0);
+
+function sumArray(a, b) {
+  return a + b;
+}
```
sandrabosk: Can you make sum function without using .reduce()? Can you think of solution using for or forEach loop? :)

sandrabosk:
Thank you very much for committing. This is very good.

PR #323, opened 2018-04-03
MIA Claudia Sanchez

clausan121:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,10 +26,27 @@ var words = [
...
+
+
+
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(oneNumber, secondNumber){
+  return oneNumber+secondNumber; 
+ }
```
sandrabosk: This solution would be correct when you add two numbers, but as you named your function sumArray, we expect an array of numbers. In that case we would change it to something like this:
```
function sumArray(array) {
  var sum = 0;
    sum += array[x];
  }
  return sum;
}
```

PR #324, opened 2018-04-03
Miami Diego Leal functions&arrays lab

DiegoLealB:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,43 @@ var words = [
   'crackpot'
 ];
 
-// Calculating a Sum
+function findLongestWord(arr1){
```
ianizaguirre: A few problems with this function you wrote.
First issue:
You wrote 
```js
  if (longestWord = ''){
    longestWord = undefined;
  }
```
You NEED either 2 equal signs or three equal signs but you wrote one equal sign.
```js
  if (longestWord === ''){ // <------- You need 3 equals in the condition
    longestWord = undefined;
  }
```
---------
Next Issue, this if statement does not do anything in the code.... So if we remove it, then we get this, which still works:
```js
function findLongestWord(arr1){
  var longestWord = '';

    for (var i =0; i < arr1.length; i++){
      if (longestWord.length < arr1[i].length){
        longestWord = arr1[i];
      }
   }
  return longestWord;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,43 @@ var words = [
...
+      longestWord = arr1[i];
+    }
+  }}
+  return longestWord;
+}
 
+// Calculating a Sum
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray([]){
```
ianizaguirre: You would not pass an empty array. This is the correct method, so you can compare:
```js
// Calculating a Sum
function sumArray(array) {
  var sum = 0;
    sum += array[x];
  }
  return sum;
}

console.log(total);
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,43 @@ var words = [
...
+  sumCount += number;
+    });
+  return sumCount;
+}
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers([]){
```
ianizaguirre: Hmm? I wonder why you think you would need to keep passing empty arrays as arguments for functions? This is incorrect. If you need help clearing this up, Slack me and we can sit down and work through it.

Here is the correct method, so you can compare:
```js
// Calculate the Average
  if (array.length === 0) {
    return undefined;
  }
  console.log(sumArray(array), array.length);
}

var average = averageNumbers(numbersAvg);
console.log(average);
```

PR #329, opened 2018-04-04
MIA | Sean - Zamora

seanzamora:
Completed JS | Functions & Arrays Assignment.


ManuelCabrera01:
dude you did exactly what  Jasmine was expecting so vary good job. and i see that your are implementing es6 code (arrow function and  hight hierarchy method ) which is great

PR #331, opened 2018-04-04
MIA-Ruben Vargas

Ambition101:


MarcosPalaciosIII:
you are heading in the right direction. Always remember that if your having trouble figuring something out that you can look it up on google and just try and make sure you know what the code your using does. overall you did well. keep it up.

PR #338, opened 2018-04-06
MIA-PT-18-MariahDunn

teepott99:
Not 100% but definitely feel like I'm getting a better feel of functions.

Let's discuss!


colbyallenc:
I agree. You're definitely getting better! The code looks good. Lets chat at some point about the one's you missed. I can offer some help.

PR #342, opened 2018-04-06
MIA-PTWD-NicholasKreps-LabJavascriptFunctionsAndArrays

sperkn:


colbyallenc:
Great approach to the functions you were able to solve. Work at DRY code. 
A good example for the third test: 
// Calculating a Sum
function sumArray(array) {
  var sum = 0;
    sum += array[x];
  }
  return sum;
}
console.log(total);

PR #343, opened 2018-04-06
MIA PT - DiLillo - jsFunctionsArrays

tommydilillo:


ManuelCabrera01:
the counting repetition exercise is similar to the previous but just adding a counter that increase its value if a word match with the parameter ... vary good job.!!

PR #345, opened 2018-04-07
MIA-PT-Francisco-Odon JS Functions and Arrays

fodon11:


colbyallenc:
Seems like functions that you were able to complete are not passing. If you would like some help, I can share with you the solution code and explain how they work.

PR #386, opened 2018-05-22
BER - John Fish - done

j31:


starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +99,12 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray(arr) {
+  return (arr.length == 0 ? undefined : Array.from(new Set(arr)))
```
ta-web-ber: Very complex solution :tada: 
I had problems in my final project with a function and I finally used Array.from(new Set(arr))

PR #387, opened 2018-05-22
BER - Ting and Steffi - pair-programming exercise day 2

steffriedrichs:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,40 @@ var words = [
...
+  return a[index]; 
+}
+
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+function sumArray(a){
+  if(a[0]!=null){
+    const reducer = (accumulator, currentValue) => accumulator + currentValue;
+    return a.reduce(reducer);
```
ta-web-ber: Good job!! You are using ECMAscript 6 (we will see it on week 4). If you feel comfortable, use it. If not, you can use syntax we are using at class.
You can use reduce directly:
```function sumArray(a){
  if(a[0]!=null){
    return a.reduce((accumulator, currentValue) => accumulator + currentValue);
  }
  return 0;
}

PR #391, opened 2018-05-22
BER - Silvio 

silvio-galli:
Not completed the Bonus exercise

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,8 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  return Math.max(num1, num2);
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +15,39 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(list) {
```
ta-web-ber: Good solution!!! :tada: But sometimes a for loop is more efficient. For example:
```
  if (arr.length === 0) {
    return undefined;
  }
  var longestWord = "";
  for (i = 0; i < arr.length; i++) {
      longestWord = arr[i];
    }
  }
  return longestWord;
}

PR #392, opened 2018-05-22
BER-Jens/spela

je0705:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +19,60 @@ var words = [
...
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers (arr2) {
+
+  var sum = 0;
+  var counter = 0;
+
+  if (arr2.length > 0){
+  for (var i = 0; i<arr2.length;i++) {
```
ta-web-ber: You can use the previous sumArray function :)

starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +120,21 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray(arr4) {
```
ta-web-ber: With this filter (we will see it tomorrow), basically:
 We iterate over the array and check if the first position of this element in the array is equal to the current position. For duplicate elements, these two positions are different.

PR #393, opened 2018-05-22
BER - Hannes

hannesreinsch:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,14 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2){
+  if(num1 > num2){
+    return num1;
+  }else if(num1 < num2){
```
ta-web-ber:
```js
else {
  return num2
}
```
It's faster :)

starter-code/src/functions-and-arrays.js
```diff
@@ -11,6 +21,19 @@ var words = [
   'crackpot'
 ];
 
+var lgth = 0;
```
ta-web-ber:
```js
function findLongestWord(wordsArr) {
  var longest = wordsArr[0];
  var candidate;

  for (var i = 1; i < wordsArr.length; i++) {
    candidate = wordsArr[i];

    if (candidate.length > longest.length) {
      longest = candidate;
    }
  }
  return longest;
}
```

ta-web-ber:
Don't forge to write the pull request like this please :) 
_BER - Hannes_

PR #409, opened 2018-05-29
PAR Ghaith & Diane

Dianedel:
@GhaithHan

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,14 @@
 // Find the maximum
+function maxOfTwoNumbers (a, b) {
+  if (a > b) {
+    return a;
+  } else if (b > a) {
+    return b;
+  } else{
+  //  return "They are equal";
+  return b
+  }
+}
 
 // Finding Longest Word
 var words = [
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +21,43 @@ var words = [
   'crackpot'
 ];
 
+
+function findLongestWord (words) {
+  var maxLengthWord = words[0].length;
```
dudydude: You should assign an empty value to this variable, per exempl : 

```js

var maxLengthWord = ""

```

It's better, because words (your array) can be empty, so you won't be able to refer to the index 0 of the Array, if you use this, it will remove some error from jasmine :)

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +21,43 @@ var words = [
...
+   for (var i = 0 ; i < words.length ; i++) {
+      if (words[i].length > maxLengthWord) {
```
dudydude: To stick to the expectation of the test, you should add an if that manage to return undifined if the array is empty

ta-web-paris:
As I told you you have to stick to requirement of the jasmine (typo, ect..), to remove some error that are still there, you have to deal with some special case (empty array per exemple).

Otherwise your code is good ! 👍 

If you have any question, let me know :)

PR #410, opened 2018-05-29
PAR Jana Manelle

janamr:
@ManelleG


ta-web-paris:
Your code is good! If you want to resolve all the error, you have to take into account the special cases (empty array, ect..) as expected by jasmine :)

If you have any question regarding this exercise let me know !

PR #411, opened 2018-05-29
PAR Nadia & Jihad & Meri

mpligioso:
@eloji9
@nadiahachemi

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,11 @@
 // Find the maximum
+function maxOfTwoNumbers(a,b){
+  if (a > b){
+    return a;
+  } else {
+  return b;
+  }
+};
 
 // Finding Longest Word
 var words = [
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,48 @@ var words = [
   'crackpot'
 ];
 
+function findLongestWord(array) {
+  var longestword = array[0];
```
ta-web-paris: You shouldn't initiate a variable like this, if the array is empty (which is the case in some test) it will return an error (as there will be no element at index 0).

Instead you can initiate it like this : 

```js
var longestWord = ""

```

ta-web-paris: Also if you want to resolve the error you need to deal with some special cases like an empty array, so you can add another if statement at the beginning of the function, and last point, in your code as you are returning longestword as soon as the condition is verified, it won't work (the function will stop as soon as it reach a word that is longer than the value inside the longest variable and the loop will stop), you need to reassign the value & to return it after the loop like this : 

```js 

 if (array.length == 0) {
    return undefined;
  }
  var longestword = "";
  for (index = 0; index < array.length; index++) {
    if (array[index].length > longestword.length) {
      longestword = array[index];
    } else continue;
  }

```

ta-web-paris:
Sometimes the test driven development (like this one) can be strange (your code is working but as it doesn't match the requirement of the test file you still have errors), so be careful and pay attention to the message written in the jasmine file :) 

Otherwise good job! Keep up the good work 👍 
(and if you have any question let me know 😄 )

PR #414, opened 2018-05-29
PAR Valérie Linarès & Shanshan Cao

moonchie:
@valerielinares

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,9 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2){
+  return Math.max(num1, num2);
+};
+
+maxOfTwoNumbers(2,3);
 
 // Finding Longest Word
 var words = [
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +16,45 @@ var words = [
...
+
+findLongestWord(words);
 
+
+// Calculating a Sum
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+var sum = 0;
+function sumArray(target){
+  var result = target.reduce(function(accumulator, currentValue, currentIndex, array) {
```
ta-web-paris: To resolve the error in jasmine just add this two lines of code : 

```js

  if (target.length == 0) {
    return sum;
  }

```

Because the jasmine want's you to deal with special case like this (empty array, ect..)

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +16,45 @@ var words = [
...
+
+sumArray(numbers)
+
+
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(target){
+  sumArray(target);  //get the sum
```
ta-web-paris: Same here you should add a condition, so when target is empty you return undefined

PR #423, opened 2018-06-12
BCN - Javier + Khalil

Javirum:
@lilhak4


starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,15 @@
 // Find the maximum
 
+var numbers = [];
+function maxOfTwoNumbers(num1,num2) {
+  console.log(Math.max(num1,num2));
```
tawebbcn: remember to remove console logs

PR #424, opened 2018-06-12
BCN - almost done - Sebastian, Julio, Julien 

juliocarlosvelez:
@se66e
@julienchilhagopian

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,6 +1,27 @@
 // Find the maximum
-
+function maxOfTwoNumbers (num1 , num2) {
+  if (num1 > num2) {
+    return num1;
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -12,6 +33,16 @@ var words = [
...
+function sumArray (array) {
+  if (array.length === 0) {
+    return 0;
+  } else if (array.length > 0) {
+    array.reduce( function (accumulate, currentVal) {
+      return accumulate + currentVal;
+    }, 0);
+  }
+}
+
```
tawebbcn:
```
function sumArray(array) {
  var sum = 0;
    sum += array[x];
  }
  return sum;
}
```

PR #425, opened 2018-06-12
done Montse and Gabriel

gabogarciam:
done Montse and Gabriel

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,6 +1,29 @@
 // Find the maximum
-
+function maxOfTwoNumbers(num1, num2) {
+    if (num1 > num2) {
+      return num1;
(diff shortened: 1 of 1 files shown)


gabogarciam:
@MontseBadia

starter-code/src/functions-and-arrays.js
```diff
@@ -12,6 +35,23 @@ var words = [
...
+    sum += element;
+  });
+  return sum;
+};
+
+function averageNumbers(array) {
+  if (!array.length) {
+    return;
+  } else if (array.length === 1) {
+    return array[0];
```
tawebbcn: You don't need this step ```sumArray()``` of 1 element in the array divided by ```array.length = 1``` will be the same as next step. Same as we talked about in findLongestWord exercise.

PR #444, opened 2018-06-12
MIA David Mudd

DragonINXS:


starter-code/tests/FunctionsAndArraysSpec.js
```diff
@@ -1,246 +1,418 @@
...
+
+// David Mudd & Ali Sattar
+
+// function maxOfTwoNumbers(num1, num2) {
+//   if (num1 > num2) {
+//     console.log(`${num1} is larger than ${num2}.`);
+//   } else {
+//     console.log(`${num2} is larger than ${num1}.`)
+//   }
+// }
```
MarcosPalaciosIII: how about if both numbers are equal to each other? you can add an else if to check for that as well.

starter-code/tests/FunctionsAndArraysSpec.js
```diff
@@ -1,246 +1,418 @@
...
+// ------------------------
 
+// var numbers = [2, 6, 9, 10, 7, 4, 1, 9];
+
+// // function calcAvg(numberArray) {
+// //   var sum = 0;
+// // }
+
+
+// var sum = 0;
```
MarcosPalaciosIII: if you put the sum variable inside of the sumBlah function then you can have sum reset to 0 each time you call the function. Otherwise having sum as a global variable it never really resets unless you make `sum = 0` inside of of your function.

starter-code/tests/FunctionsAndArraysSpec.js
```diff
@@ -1,246 +1,418 @@
...
+// function sumBlah(array){
+//   array.forEach(function(placeholder) {
+//     sum += placeholder;
+//   });
+//     return sum;
+// }
+
+// // sumBlah(numbers)
```
MarcosPalaciosIII: calling this first and then again within the function below having a global variable without a reset will give you an incorrect result.

starter-code/tests/FunctionsAndArraysSpec.js
```diff
@@ -1,246 +1,418 @@
...
+//     'fuel',
+//     'palace'
+// ];
+
+// words.forEach(function(i) {
+//   console.log(i.length);
+// });
+
+// var sum;
+
```
MarcosPalaciosIII: you already have a `var sum = 0`. in order to change that var afterwards you don't need to include var before the variable name.

MarcosPalaciosIII:
good work on the functions and arrays lab. The take away from this lesson is that you don't always want to have a global variable and if you do you have to remember to adjust that variable accordingly depending the situation that you use it for. 

Keep up the good work.

PR #460, opened 2018-06-13
MIA Rodrigo Flores

rflor012:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +15,101 @@ var words = [
   'crackpot'
 ];
 
+function longest(words) {
```
MarcosPalaciosIII: your function seems to be a little off. It should look something more like this 
`
function longest(words) {
var longest = "";
 for(i=0;i < words.length;i++) {
   if(words[i].length > longest.length) {
     longest = words[i];
     } else {
       continue;
     }
   }
 return longest;
}
`

MarcosPalaciosIII:
good work on the lab. Just remember to look over what was given for examples in class so yo can compare errors that you may have had.
Keep up the good work.

PR #488, opened 2018-08-07
PAR Abi & Pauline

paulettepaillette:
@abisummers


ta-web-paris:
Perfect, logic and you passed all the test! 
Great use of the forEach method
Be careful to your parameters name: naming them like a javascript data type like "array" is not advised as it can cause issues later, le latter "myArray" is better. Also, the function inside the forEach takes as a parameter one element of the array, so you can put it in singular (word instead of words) to better signal intent.
Good job!

PR #489, opened 2018-08-07
PAR John & Mohamed

SMoZam:
@johndemetroff

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -48,9 +48,48 @@ return result;
 
 // Calculate the Average
 
+function averageNumbers(arrayNumber) {
+
+var result = 0;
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,25 +1,95 @@
...
+
+
+
 // Calculate the Average
 
+function averageNumbers(arrayNumber) {
+
+var result = 0;
+for ( var i=0 ; i < arrayNumber.length ; i++) {
+
```

``` javascript
function averageNumbers(arrayNumber) {
   var result = sumArray(arrayNumber);
   if (arrayNumber.length > 0) {
   var average = result / arrayNumber.length; 
   }
   return average;
 }
```

ta-web-paris:
You passed the code and everything is well understood. I like that the name you give to your parameters are very specific, this is the good way of doing it!

PR #490, opened 2018-08-07
PAR Constance & Emmanuelle

emmanuellejaldorau:


emmanuellejaldorau:
@Cr27910

ta-web-paris:
You passed the tests and your code is clean and correct, great! 
Be careful to the name you give to your parameters, "myArray" is generally better than "array" as "array" has its own meaning in javascript. 
You can try to do a few more iterations, all of this you will use a lot later. Tell me if you have more things you want me to review :)

PR #491, opened 2018-08-07
PAR Luca & Manon

MSLN8:
@LucaMerzi


ta-web-paris:
Good job :1st_place_medal: :1st_place_medal: :1st_place_medal: !
Great logic.
Small tip: The name you give to your parameters matters, when your code will get longer and longer, you need to be able to understand it fast when scrolling on it. Parameters should be camelCase (not start with a cap letter) and having both "word" and "mot"as parameters doesn't make your intent too clear.

if you feel comfortable with that already, you can try to use the ternary operator instead of if statements when relevent, and the reduce and map methods :)

PR #492, opened 2018-08-08
PAR Yvens & Aurélien

LOOFFI:
@Yvens16

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,6 +1,10 @@
 // Find the maximum
+function maxOfTwoNumbers (a,b) {
+return Math.max(a,b);
+}
 
 // Finding Longest Word
+
 var words = [
   'mystery',
   'brother',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -59,6 +141,16 @@ var wordsFind = [
...
+function doesWordExist (myArray,word) {
+  var verif=false;
+  for (var i=0; i<myArray.length; i++) {
+    if (myArray[i] === word) {
+      verif=true;
+      break;
+    }
+  }
+  return verif;
+}
```
ta-web-paris: the code works but it is not ideal to use a break statement, 'break' should only be used in very specific cases and if you need to use it, it usually means that there is a better way to write the code, 

``` javascript
function doesWordExist (myArray,word) {
  
  for (var i=0; i<myArray.length; i++) {
    if (myArray[i] === word) {
      return true;
    }
  }
  return false;
}
```
whenever a function returns something, the rest of the code inside of the function is not executed anymore. If the condition myArray[i] === word is verified, you got what you were looking for and you can return. If the condition if never verified,  you can return false at the very end of the function, after the loop.

ta-web-paris:
Good work! keep up :+1:

PR #493, opened 2018-08-08
PAR Paul & Nérimène

PaulAbarnou37:
@nerimeneakari

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,12 @@
 // Find the maximum
+function maxOfTwoNumbers(nb1, nb2) {
+  if (nb1 < nb2) {
+    return nb2;
+  } else {
+    return nb1;
+  }
+ }
+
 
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +19,44 @@ var words = [
...
+  var sum = 0;
+  var average ;
+  if (numbersAvg.length < 1) {
+    return undefined ;}
+  else {for (i = 0; i < numbersAvg.length; i++) {
+    sum += numbersAvg[i];
+    }
+    average = sum / numbersAvg.length;
+    return average ;}
+  }
```
And to avoid having to return undefined, you can do the if statement the other way :
``` javascript
   var sumResult = sumArray(numbersAvg);
   if (numbersAvg.length > 0) {
   var average = sumResult / numbersAvg.length; 
   }
   return average;
 }
```

ta-web-paris:
Good work! The logic is good and the code is easy to read :+1:

PR #494, opened 2018-08-08
done

Youlie03:
@alice.dbrg

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,4 +1,12 @@
 // Find the maximum
+function maxOfTwoNumbers(x,y){
+  if (x > y){
+    return x;
+  }else{
+      return y;
+    }
+  }
+  maxOfTwoNumbers(1,3)
 
 // Finding Longest Word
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -10,16 +18,51 @@ var words = [
...
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+function averageNumbers (array){
+var sumAverage = 0;
+for (i=0; i<array.length; i++){
+sumAverage = sumAverage + array[i];
+var average = sumAverage/array.length
+}
+return average
+}
```
``` javascript
function averageNumbers(arrayNumber) {
   var sumRes = sumArray(arrayNumber);
   if (arrayNumber.length > 0) {
   var average = sumRes / arrayNumber.length; 
   }
   return average;
 }
```

ta-web-paris: the if statement here (  if (arrayNumber.length > 0) ) is to handle the case when we try to pass an empty array to the function, to please Jasmine :)

PR #497, opened 2018-08-14
MIA FTWD Nick Crispino

crispinonicky:


starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +142,24 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray(array){
```
MarcosPalaciosIII: this function will return 2 errors in the jasmine test. A simpler way to go about this function is to mainly compare arrays. You can do something like this 
```
function uniquifyArray(arrayOfWords){
if (arrayOfWords.length === 0){
  return undefined;
}
let resultArray= [];
 arrayOfWords.forEach((oneWord)=>{
  if (resultArray.indexOf(oneWord) === -1){
    resultArray.push(oneWord);
  }
});
return resultArray;
}
```

MarcosPalaciosIII:
great work on the functions and arrays lab. Just remember when working with a team to talk with your teammates when stuck on some code. Sometimes they may have a solution or portion of a solution which will allow for all to finish faster. Also many times a simple answer works best for most code.

keep up the great work.

PR #498, opened 2018-08-14
MIA Douglas DeJong FTWD

DougDeJong:


starter-code/src/functions-and-arrays.js
```diff
@@ -60,6 +143,30 @@ var wordsFind = [
...
+      return true 
+    }
+    else {
+      console.log('false');
+      return false
+    }
+  })
+
+}
+// The reason this loop doesn't work is because a forEach loop won't result in a return value.
```
MarcosPalaciosIII: good catch on the loop.

starter-code/src/functions-and-arrays.js
```diff
@@ -74,6 +181,17 @@ var wordsCount = [
   'disobedience',
   'matter'
 ];
+
+function howManyTimes(wordsCount, countWord){
+  if (wordsCount.length === 0){
+    return false;
+  }
+  for (var i = 0; i< wordsCount.length; i++){
+  let numTimes = (countWord.indexOf(wordsCount[i]))
```
MarcosPalaciosIII: best to set your variable within the function outside of loops or statements and then modify them within the loop or statement.

MarcosPalaciosIII:
Good job on the functions and arrays lab. Remember when working in the team to communicate with your partners to see if someone may have come up with a portion of the solution or the solution itself in order to complete the task. It Seems both you and your partner had a portion of the of the solution to the last part of the jasmine test. 

Keep up the good work.

PR #499, opened 2018-08-15
MIA FTWD - Functions + Arrays - Lauren Cahill

laurencahill:
functions-and-arrays assignment, unfinished with 6 errors to resolve

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,5 +1,13 @@
 // Find the maximum
 
+function maxOfTwoNumbers (firstNum, secondNum){
+  if (firstNum > secondNum){
+    return firstNum;
+  } else {
+    return secondNum;
+  }
+}
+
 // Finding Longest Word
 var words = [
   'mystery',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -60,6 +138,19 @@ var wordsFind = [
   'disobedience'
 ];
 
+function doesWordExist(wordsFind) {
```
MarcosPalaciosIII: You would probably need to pass 2 parameters into this function in order to have a word to check if it exists in the array you pass in

starter-code/src/functions-and-arrays.js
```diff
@@ -74,6 +165,23 @@ var wordsCount = [
   'disobedience',
   'matter'
 ];
+
+function howManyTimes(wordsCount) {
```
MarcosPalaciosIII: You would probably need to pass 2 parameters into this function in order to have a word to check how many times the word is in the array you pass in

MarcosPalaciosIII:
Good job on the functions and arrays lab. Remember when working in the team to communicate with your partners to see if someone may have come up with a portion of the solution or the solution itself in order to complete the task. It Seems both you and your partner had a portion of the of the solution to the last part of the jasmine test. 

Keep up the good work.

PR #502, opened 2018-08-15
Levijs and Sat assignment MIAMI-FTWD

levijskal:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,178 @@
...
 // Calculating a Sum
-
+​
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
-
+​
+function sumArray(nums) {
+ if (nums.length > 0) {
+ var reducer = (accumulator, currentValue) => accumulator + currentValue;
+ return nums.reduce(reducer);
```
sandrabosk: This works but make sure you understand how .reduce() works under the hood.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,178 @@
...
+ 'sharp',
+ 'playground',
+ 'poison',
+ 'communion',
+ 'simple',
+ 'bring'
 ];
-
+​
+function uniquifyArray(uniqueArray) {
```
sandrabosk: Try to find a simpler solution.

sandrabosk:
Good work. Thank you for submitting.

PR #507, opened 2018-08-20
Mia ftwd Alex Rojas

alexrojas:


MarcosPalaciosIII:
great job on the functions and arrays lab. The main takeaway is when doing test driven development is to help you realize things that you may have missed when trying to get your answer. You may have things working one way and the test requires something else that you may not have thought of which will in turn help you start thinking outside the box or about possible outcomes that you have not thought of before. 

Keep up the great work.

PR #518, opened 2018-08-21
BCN - Francesca Barbara

barkrrr:
@francescaropolo

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,52 +1,47 @@
 // Find the maximum
+function maxOfTwoNumbers (num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,52 +1,47 @@
...
+  for (var ix = 0; ix < numbers.length; ix++){
+    sum += numbers[ix];  
+  }
+  return sum;
+}
+    
 
 // Calculate the Average
+function averageNumbers (numbers) {
+  const numbers = [2, 6, 9, 10, 7, 4, 1, 9].reduce((a, b) => a + b, 0);
```
tawebbcn: Here you could have just called the sumArray function you have above and pass it the numbers. You over complicated a bit here. There's a log going on in that reduce

PR #524, opened 2018-08-21
BCN-Axel

axelgar:
**Problems with the .DS_Store**


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,118 @@
...
+}
+
+
+function averageNumbers(arrayOfNumbers){
+  if (arrayOfNumbers.length === 0){
+    return undefined;
+  }else {
+    var sum = 0;
+    for (var i = 0; i < arrayOfNumbers.length; i++){
+      sum += arrayOfNumbers[i];
```
tawebbcn: Why not use the fuction to sum numbers above?

PR #532, opened 2018-08-21
BCN - Caroline - functions and arrays exercises

Caroline-GMR:


starter-code/src/functions-and-arrays.js
```diff
@@ -13,13 +33,32 @@ var words = [
...
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arr){
+  var total = 0;
+  var average = 0;
+  arr.forEach(function(item){
+    total += item;
+  });
```
tawebbcn: Why not use the function to sum array above!

starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +98,16 @@ var wordsUnique = [
...
 ];
 
+function uniquifyArray(arr){
+  var unique = [];
+  arr.forEach(function(item){
+    if(unique.indexOf(item) < 0){
+      unique.push(item);
+    }
+  });
+  return unique.length > 0 ? unique : undefined;
```
tawebbcn: Nice touch! Terniary operator!

PR #536, opened 2018-08-22
BCN - Diana

Dbpautt:
I think didn't achieve the 10% 😞


starter-code/src/functions-and-arrays.js
```diff
@@ -1,79 +1,68 @@
...
-  'crocodile',
-  'pearl',
-  'orchard',
-  'crackpot'
-];
+
+function findLongestWord (words) {
+  var longestWord = '';
+  if (words.length === 0) { return; };
+  words.forEach(index) ;
```
tawebbcn: Careful here. forEach accepts a function

PR #540, opened 2018-08-22
BCN - MJ functions and arrays exercise

MJHRhacker:
:(

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,23 +1,74 @@
 // Find the maximum
+function maxOfTwoNumbers (numberOne, numberTwo) {
+  if (numberOne > numberTwo) {
+    return numberOne;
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,24 +1,75 @@
...
-  'crackpot'
-];
+var words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+function findLongestWord () {
+  var firstOccurence = 0;
+
+  for (var wordsIndex = 0; wordsIndex < words.length; wordsIndex++) {
+    return 0;
+    if (wordsIndex = words.indexOf(firstOccurence)) {
```
tawebbcn: Careful! One equals sign is an assignment, not a comparison.

PR #555, opened 2018-09-19
BCN -Iban Dasca

ibandasca:
Enhancement JS document


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,486 @@
 // Find the maximum
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
+  } else {
+    return (num1 = num2);
```
tawebbcn: ` return num1; `

PR #568, opened 2018-10-05
<PAR PT> Sandra

SanDDupt:
corrigé
@ta-webpt-par


starter-code/src/functions-and-arrays.js
```diff
@@ -10,14 +23,52 @@ var words = [
...
+
+// Calculate the Average
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+function averageNumbers(numbersAvg) {
+  var sum = 0;
+  var average = 0;
+  if (numbersAvg.length > 0) {
+  for (var i = 0; i < numbersAvg.length; i++) {
+    sum += numbersAvg[i];
+  }
```
ta-webpt-par: To compute the sum, you could have used the function `sumArray`.

ta-webpt-par:
Great work !

PR #570, opened 2018-10-09
Done with most of them. I'm not sure what I'm supposed to accomplish …

wildhamster26:
…with the bonus challenge...


wildhamster26:
Done with the bonus question as well.

PR #612, opened 2018-10-17
done

mehdilouraoui:
w/ @geoffroybag


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,12 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  }
+  else {
+    return num2;
+  }
+}
```
ta-web-paris: Perfect!

An even faster way to do it would have been : 
```
  return Math.max(num1, num2);
}
```

PR #613, opened 2018-10-17
Paris Cécile et Régis

jourisque:
@gally999

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,6 +1,20 @@
 // Find the maximum
 
-// Finding Longest Word
+
+function maxOfTwoNumbers (num1, num2) {
+  if (num1 > num2) {
+
+    return num1;
+  }
+
+  else {
+
+    return num2;
+  }
+}
+
+
+// -----------------Finding Longest Word
 var words = [
   'mystery',
   'brother',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +124,34 @@ var wordsUnique = [
...
+function uniquifyArray (array) {
+  var newArray = [];
+    if(array.length === 0) {
+        return undefined;
+    }
+
+    else {
+      newArray.push (array[0]);
+
+      for (var i=1; i< array.length; i++){
```
ta-web-paris: Using a for loop IN a for loop, is quite risky.. specially when you have a much more simple way to do it. 

You can use 2 methods:

indexOf : 
array.indexOf("Word") => will return the index of Word in the array if Word exists, or return -1 if the word doesn't exist. If it returns an index, you shouldnt push, if it returns -1 then you should push it.

Includes:
array.includes("Word") => will return TRUE if word exists, or FALSE if it doesnt exist. Push when its false. 

example with the indexOf :
```
      newArray.push(array[i]);
    }
  }
```

Hope it's clear :)

starter-code/src/functions-and-arrays.js
```diff
@@ -74,8 +192,43 @@ var wordsCount = [
...
+  var temp =0;
+
+    
+
+  return temp;
+}
+
 var matrix = [
   [8, 2, 22, 97, 38, 15, 0, 40, 0, 75, 4, 5, 7, 78, 52, 12, 50, 77, 91, 8],
   [49, 49, 99, 40, 17, 81, 18, 57, 60, 87, 17, 40, 98, 43, 69, 48, 4, 56, 62, 0],
```
ta-web-paris: Here is the correction for the Matrix: 

```
function greatestProduct(m) {
  var horizontal, vertical, diagonal;
  var greatProduct = 0;

  for (var i=0, nRows = m.length; i < nRows - 1; i++) {
    for (var j=0, nCols = m[0].length; j < nCols - 1; j++) {
      // 4 horizontal elements
      if (j < nCols-3) {
        horizontal = m[i][j] * m[i][j+1]   * m[i][j+2]   * m[i][j+3];
      }

      // 4 vertical elements
      if (i < nRows-3) {
        vertical   = m[i][j] * m[i+1][j]   * m[i+2][j]   * m[i+3][j];
      }

      // 4 Diagonal elements
      if (i < nRows-3 && j < nCols-3) {
        diagonal   = m[i][j] * m[i+1][j+1] * m[i+2][j+2] * m[i+3][j+3];
      }

      greatProduct = Math.max(greatProduct, horizontal, vertical, diagonal);
    }
  }

  return greatProduct;
}
```

ta-web-paris: It's globally well done! The code is clean and not too long. 💪

Keep up the good work!

PR #616, opened 2018-10-17
Antoine & Filippo

DoneDeal0:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,14 @@
 // Find the maximum
+function maxOfTwoNumbers (a, b) {
```
ta-web-paris: Perfect!

starter-code/src/functions-and-arrays.js
```diff
@@ -8,18 +18,45 @@ var words = [
   'crocodile',
   'pearl',
   'orchard',
-  'crackpot'
+  'crackpot',
+  'zizouisthegreatestofalltime'
 ];
+var bigWord = "";
```
ta-web-paris: Well done!

starter-code/src/functions-and-arrays.js
```diff
@@ -33,6 +70,15 @@ var wordsArr = [
...
 ];
 
+function averageWordLength (a){
+  var newSum = a.reduce(function (acc, curr){
+    return (acc + curr.length);
+  },0);
+  return newSum / a.length;
+  };
+
+console.log(averageWordLength (wordsArr));
```
ta-web-paris: That's good! Keep up the good work!

Make sure you use the right function names as the SpecRunner.html (that is supposed to help you debug your code) takes function names into account to run the debugger. 

Here is the correction for all the other exercices : https://github.com/ironhack-paris-webdev-august-2018/lab-javascript-functions-and-arrays/blob/master/solution-code/src/functions-and-arrays.js

PR #618, opened 2018-10-20
[PAR-PT] SARO

ta-webpt-par:
Just Starting
@ta-webpt-par


ta-webpt-par:
You're really doing good ! Keep on working on this exercise to have a more general idea about what's comming next. Really good !
- Jihad

PR #619, opened 2018-10-20
[PAR-PT] Elodie

ta-webpt-par:
19 errors again
@ta-webpt-par


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,14 @@
 // Find the maximum
+function maxOfTwoNumbers(numberOne, numberTwo) {
```
ta-webpt-par: You can either create three conditions (equal, greater or smaller) or only two with (greater or equal and smaller or the opposite)

PR #628, opened 2018-10-23
AMS Lyudmyla Baron

lxs07a:


starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +97,19 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray(array) {
```
IronhackAMS: Nice I like the way you solved this with localCompare(). If you'd like to write it shorter, however, have a look at .filter(), combined with indexOf();

starter-code/src/functions-and-arrays.js
```diff
@@ -98,3 +182,28 @@ var matrix = [
   [20, 73, 35, 29, 78, 31, 90, 1, 74, 31, 49, 71, 48, 86, 81, 16, 23, 57, 5, 54],
   [1, 70, 54, 71, 83, 51, 54, 69, 16, 92, 33, 48, 61, 43, 52, 1, 89, 19, 67, 48]
 ];
+
+function greatestProduct(array) {
```
IronhackAMS: Nice work!

PR #630, opened 2018-10-23
AMS Ana Kelly

anakellyc:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +21,51 @@ var words = [
...
+  return sum
+}
+console.log(sumArray(numbers))
+
+// Sum using reduce
+var sum = numbers.reduce(function(accumulator, currentValue){
+  return accumulator + currentValue
+},0)
+console.log(sum)
+
```
IronhackAMS: Very nice Ana that you implemented the reduce, your getting the hang of it.
Let's level up from now on and use ECMA 6 with let and const instead of var and the use of =>
Also one of the main features of higher functions like reduce is that it has less code.
If we combine the reduce method above with ECMA 6, we get even less, but more clean code:
`const sum = numbers.reduce((accumulator, currentValue) => {
return accumulator + currentValue;
}, 0);`

starter-code/src/functions-and-arrays.js
```diff
@@ -59,6 +138,16 @@ var wordsFind = [
...
+function doesWordExist (array, word) {
+  var wordExists = false
+  for (var i = 0; i<array.length; i++) {
+    if (word == array[i]) {
+      wordExists = true
+    } 
+  }
+  return wordExists
+}
+console.log(doesWordExist(wordsFind,"disobedienc"))
```
IronhackAMS: Nice job Ana! Maybe as a bonus try to solve this one also with a higher function like filter?
Kido

IronhackAMS:
Awesome coding Ana, keep it up!

PR #647, opened 2018-10-23
 MIA FTWD Isaiah Gonzalez

Isaiah361:


MarcosPalaciosIII:
Excellent work on the functions and arrays lab. The main take away from this lab is learning that there is more than one way to complete a challenge when coding. Some Methods are faster and some slower, some longer and some shorter. 

Keep up the excellent work.

PR #650, opened 2018-10-23
AMS Dea Kristina

Deakristina:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,16 @@
 // Find the maximum
 
+function maxOfTwoNumbers(numOne, numTwo){
```
IronhackAMS: Nice work. See if you can write less code by looking at a ternary statement.

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +22,59 @@ var words = [
...
 
+function averageNumbers(numArray){
+  let sum = 0;
+  for (let i = 0; i < numArray.length; i++){
+    sum += numArray[i];
+  }
+  return sum/numArray.length;
+}
+
+let calculateAvg = averageNumbers(numbersAvg);
```
IronhackAMS: Since you are using the return statement and the function is not anonymous, it's not necessary to store the function in an additional variable

starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +115,20 @@ var wordsUnique = [
   'bring'
 ];
 
+
```
IronhackAMS: Very nice!

PR #652, opened 2018-10-23
MIA FTWD Javier Arrinda

javierarrinda:


MarcosPalaciosIII:
Good job on the functions and arrays lab. The main takeaway is that there are multiple ways to resolve challenges. Working with a partner is a good way to get to see problem solving from a different prospective. 

Keep up the good work.

PR #653, opened 2018-10-23
MIA- AngelWu

angelwu007:


MarcosPalaciosIII:
Good job on the functions and arrays lab. As we progress you will start to use these functions a bit more and will then make it easier to use and understand. 

Keep up the good work.

PR #654, opened 2018-10-23
AMS David Z - Fuctions and Array's

zlobic:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,30 +27,89 @@ var words = [
   'crackpot'
 ];
 
-// Calculating a Sum
```
IronhackAMS: Strictly speaking you could , instead of declaring the variable lengthOfWord use array[i].length in the conditional statement. This will make your code slightly shorter and more readable. Nevertheless, good work!

starter-code/src/functions-and-arrays.js
```diff
@@ -11,30 +27,89 @@ var words = [
...
+   'smell',
+   'smart',
+   'chaos',
+   'fuel',
+   'palace'
+ ];
+
 
-// Unique Arrays
-var wordsUnique = [
```
IronhackAMS: numOfElements could be written at array.length right?

zlobic: Yes, I need to watch this, thanks for the feedback:)

PR #655, opened 2018-10-23
MIA FTWD Jaime Jacobo

JaimeJacobo:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,16 +22,67 @@ var words = [
...
+let longestNumber = 0;
+let longestWord = '';
+
+if (array.length === 0) {
+  return undefined;
+}
+
+for(let i = 0; i < array.length; i++) {
+  if (array[i].length > longestNumber) {
+    longestNumber = array[i].length;
```
MarcosPalaciosIII: this could be done by comparing ```array[i].length and longestWord.length```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,16 +22,67 @@ var words = [
...
+
+
+function sumArray(x){
+  let total = 0;
+  if (numbers.length === 0){
+    return 0
+  }
+
+ x.forEach((numbers)=>{
+    total = total + numbers;
```
MarcosPalaciosIII: here you can do  ``` total += number;```

starter-code/src/functions-and-arrays.js
```diff
@@ -60,8 +155,24 @@ var wordsFind = [
...
+
+let newWords;
+if(array.length === 0) {
+  return false;
+}
+for (let i = 0; i < array.length; i++) {
+  if (word === array[i]) {
+ return newWords = true;
+  } else {
+     newWords = false;
```
MarcosPalaciosIII: for this function you can just return true or false in your if else statement. There is no need to have a variable hold the boolean that you will return;

MarcosPalaciosIII:
Great job on the functions and arrays lab. I few things you can do to refactor your code to be a bit more streamline, but with time and practice you will start naturally doing it. 

Keep up the great work.

PR #671, opened 2018-10-28
AMS Amel Ramirez

amelrc:


IronhackAMS:
Well done! Please note the formatting of your code. Keep good practice indenting for example. And if you're not using a variable you don't need to declare it. So look at moments when your code is redundant. In some cases you also use parameters that are not very descriptive. If you make your code more descriptive it will be easier to read, debug and understand.

PR #680, opened 2018-10-30
[PTMIA1018] Chris - Nicolas

cvazquez08:
Lab solutions

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -90,7 +90,7 @@ function averageWordLength(words){
 averageWordLength(lengthWords);
 
 // Unique Arrays
-var wordsUnique = [
+let uniqueWords = [
   'crab',
   'poison',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +22,45 @@ var words = [
...
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+let counter = 0;
+
+function sumArray(nums){
+  for(let i = 0; i < nums.length; i++)
+    counter += nums[i];
+    console.log(counter);
+}
```
ta-web-mia: I see that you didn't return something with some other functions as well. I'd recommend to return something with most functions. Though not all functions necessarily should return something, it's good practice to do for the most part, especially when you have a function so specific as sumArray() where you'd return the sum

PR #684, opened 2018-10-30
[PTMIA1018] Chris - Gustavo

Jackslaps:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +22,58 @@ var words = [
...
+  if(words.length == 0) {
+    return undefined;
+  }
+
+  let length = 0;
+  let longest = [];
+
+  for(let i=0; i < words.length; i++){
+    if(words[i].length > length){
+        length = words[i].length;
```
ta-web-mia: good logic! Though where are you using your variable length? Remember if you're not going to use it, you don't need to have a variable there

ta-web-mia:
Good code so far! Make sure to complete this lab that way I can complete your pull request

PR #757, opened 2019-01-09
PARIS - ESTELLE & JULES F. - FUNCTIONS AND ARRAYS 

F-Jules:
@wnski

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,100 +1,433 @@
 // Find the maximum
 
+function maxOfTwoNumbers(x, y) {
+  if (x > y) {
+    return x;
+  } else if (y > x) {
+    return y;
+  } else {
+    return x;
+  }
+}
+
 // Finding Longest Word
 var words = [
-  'mystery',
-  'brother',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,433 @@
...
+  }
+  var wordsString = words.toString();
+  var regex = /,/gi;
+  var wordsReplace = wordsString.replace(regex, "");
+  var avgWords = wordsReplace.length / words.length;
+  return avgWords;
+}
+
+console.log(averageWordLength(wordsArr));
+
```
ta-web-paris: Ok, this is working. Here is another way to do it that I find easier to understand:
```
  if (array.length === 0) {
    return undefined;
  }
  var sumLetters = 0;
    sumLetters += array[i].length;
  }
  return sumLetters / array.length;
}
```

ta-web-paris:
That's good! You understood the concepts and your code is clear and easy to understand. Continue this way!

ta-web-paris:
That's good! You understood the concepts and your code is clear and easy to understand. Continue this way!

PR #760, opened 2019-01-09
Paris - Raja & Delphine

Deie:
@RaproCode

Exercise not done but we'll try to finish and most importantly understand it by tonight : )

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,5 +1,9 @@
 // Find the maximum
+function maxOfTwoNumbers(a, b) {
+  return Math.max(a, b);
+}
 
+maxOfTwoNumbers(4, 6);
 // Finding Longest Word
 var words = [
   'mystery',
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +15,69 @@ var words = [
...
+  }
+  for (var i = 0; i < avgNumber.length; i += 1) {
+    //  console.log(avgNumber[i]);
+    total = total + avgNumber[i];
+    if ()
+  }
+  return total;
+}
+averageNumbers(avgNumber);
+
```
ta-web-paris: The method began in a good way. You probably missed a bit of time to finish it. Here is the solution for this part. 
```
  if (array.length === 0) {
    return undefined;
  }
  console.log(sumArray(array), array.length);
}

var average = averageNumbers(numbersAvg);
console.log(average);
```

ta-web-paris:
Keep on going your good work! And let us know if you continue working on the exercise because we won't receive any notification. Like that we can correct it. :)

PR #761, opened 2019-01-09
Paris - Christina & Thomas

tomdreamer:
@CIBailey

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,100 +1,438 @@
 // Find the maximum
 
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  } else {
+    return b;
+  }
+}
+m;
+console.log(maxOfTwoNumbers(4, 2));
 // Finding Longest Word
+
 var words = [
-  'mystery',
-  'brother',
-  'aviator',
-  'crocodile',
-  'pearl',
-  'orchard',
-  'crackpot'
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
 ];
+var lgth = 0;
+
+for (var index = 0; index < words.length; index++) {
+  if (words[index].length > lgth) {
+    var lgth = words[index].length;
+    answer = words[index];
+  }
+}
+
+console.log(answer);
+
+function findLongestWord(words) {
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,438 @@
...
+
+function findLongestWord(words) {
+  words.forEach(function(wordsIndex) {
+    if (wordsIndex.length > lght) {
+      var lght = wordsIndex.length;
+      answer = wordsIndex;
+    }
+  });
+}
+console.log(answer);
```
ta-web-paris: You should create your function first and and initialise the var inside the function. The variable is there to stock the longest word.
Then you define if the array is empty by an `if statement`. You could use a `forEach` but I think it's simpler with a `for loop`. 

Here is the way to do it: 
```
function findLongestWord(array) {
  var longest = wordsArr[0];
if (array.length === 0) {
  return undefined;
  }

for (var = i; i < array.length; i++) {
  if (array[i].length > longestWord.length){
    longestWord = array[i];
    return longestWord;
    }
  }
};
```

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,438 @@
...
+    sum1 = sum1 + numbers[index];
+  }
+  return sum1;
+}
+
+function averageNumbers(fff) {
+  var average = sumArray(fff) / fff.length;
+  return average;
+}
+averageNumbers(numbersAvg);
```
ta-web-paris: If you put twice the same function, Jasmine is not gonna like it. :)

The first part is really confused but you did well in the second part. Just the variable is still declared outside the function and should be inside and the condition that if the array is empty is not written but it's right.

Here is how you could have it completely right: 
```
function sumArray(numbers) {
  var sum1 = 0;
  for (var index = 0; index < numbers.length; index++) {
    sum1 = sum1 + numbers[index];
  }
  return sum1;
}

function averageNumbers(fff) {
  if (fff.length === 0) {
    return undefined;
  }
  var average = sumArray(fff) / fff.length;
  return average;
}
averageNumbers(numbersAvg);
```

ta-web-paris:
Some parts seems really confused but some others are working. Don't hesitate to ask us (TAs) to have more explanations during the labs and even now. We are here to help. :) Don't worry, after coding and coding again, it's going to be just a way of thinking. (Chloé)

PR #763, opened 2019-01-09
Paris - Francesca && Kevin M

wowfranzik:
We need extra time to finish!!

@KevGhub

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,100 +1,413 @@
 // Find the maximum
-
+function maxOfTwoNumbers(numberOne, numberTwo) {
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,413 @@
...
+    return 0;
+  }
+  var sum = numbers.reduce(function(total, currentValue) {
+    return total + currentValue;
+  });
+  return sum;
+}
+
+console.log(sumArray(numbers));
+
```
ta-web-paris: nice use of `reduce`

ta-web-paris:
Well done! The part you did is perfect. Sometimes it's hard to finish the labs. If you have time to finish them later that's good! Keep going like this!

PR #765, opened 2019-01-09
Paris Aurelie & Alexandre

AurelieBeno:
We struggle on the level 2 Array of string :(

@alexjoly13

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,100 +1,430 @@
 // Find the maximum
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  } else {
+    return b;
+  }
+}
 
 // Finding Longest Word
+
 var words = [
-  'mystery',
-  'brother',
-  'aviator',
-  'crocodile',
-  'pearl',
-  'orchard',
-  'crackpot'
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
 ];
+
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,430 @@
...
+  //  var numbers = [2, 6, 9, 10, 7, 4, 1, 9];
+  var reducer = (accumulator, currentValue) => accumulator + currentValue;
+  if (numbers.length < 1) {
+    return undefined;
+  } else {
+    var average = numbers.reduce(reducer) / numbers.length;
+    return average;
+  }
+}
+averageNumbers(numbers);
```
ta-web-paris: Again good use of `reduce`

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,430 @@
...
+  var num = wordsArr.length;
+  console.log(num);
+  console.log(sum);
+  for (var i = 0; i < num; i++) {
+    sum += num;
+
+    return sum / num;
+  }
+}
+
```
ta-web-paris: Your function can't work because you're not passing the variable in into the `for loop`.
You need to pass all the elements of the array and add their length to the sum. 

You could have done it like this: 
```
  if (array.length === 0) {
    return undefined;
  }
  var sumLetters = 0;
    sumLetters += array[i].length;
  }
  return sumLetters / array.length;
}
```

ta-web-paris:
Well done! You understood the notions of the lab. Your code is clear and easy to read which is good! Keep doing this way!

PR #766, opened 2019-01-09
Paris - Nadjie & Jules B

NadjieVal:
we had some problem to increment everything needed in the function. Sometimes we don't know what should be in or out.

@julesbadin


ta-web-paris:
Well done! You seems to have understood the notions but don't hesitate to ask us (TAs) questions if something is not clear.

PR #767, opened 2019-01-09
Paris - Céline

frietpoune:
Got stuck for Finding Longest Word

For Finding Elements, could not find a way to stop the code once they found the word. My code compares the word given with each element of the array.

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,100 +1,465 @@
 // Find the maximum
+var nb1 = 2;
+var nb2 = 3;
+
+function maxOfTwoNumbers(nb1, nb2) {
+  return Math.max(nb1, nb2);
+}
 
 // Finding Longest Word
 var words = [
-  'mystery',
-  'brother',
-  'aviator',
-  'crocodile',
-  'pearl',
-  'orchard',
-  'crackpot'
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
 ];
+
+function findLongestWord(words) {
+  for (var i = 0; i < words.length; i++) {
+    var long = "";
+    if (words[i].length > long.length) {
+      long = words[i];
+    } else {
+      long = long;
+    }
+  }
+  return long1;
+}
+
+findLongestWord(words);
 
 // Calculating a Sum
 
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(numbers) {
+  var sum = 0;
+  for (var i = 0; i < numbers.length; i++) {
+    sum = sum + numbers[i];
+  }
+  return sum;
+}
+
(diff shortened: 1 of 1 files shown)


frietpoune:
Binôme avec @kevbarns

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,465 @@
 // Find the maximum
+var nb1 = 2;
+var nb2 = 3;
+
+function maxOfTwoNumbers(nb1, nb2) {
+  return Math.max(nb1, nb2);
+}
```
ta-web-paris: Here you don't need to create variables with real numbers at the beginning. You just need to write the function you did after which is perfectly working.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,465 @@
...
+      long = words[i];
+    } else {
+      long = long;
+    }
+  }
+  return long1;
+}
+
+findLongestWord(words);
+
```
ta-web-paris: You need to declare your variable inside the function but outside the `for loop` to store the value of the longest word.
The else is not necessary.
You are returning something different `long1` than the name you gave at the beginning of your variable `long` so it can't work.

Here is how you could have done it (but you were pretty close):
```
  var long = "";
  if (words.length === 0) {
    return undefined;
  }
    if (words[i].length > long.length) {
      long = words[i];
    } 
  }
  return long;
}

findLongestWord(words);
```

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,465 @@
...
+      if (indices.length >= 2) {
+        var indicebis = indices[1];
+        wordsUnique.splice(indicebis, 1);
+      }
+    }
+  }
+  return wordsUnique;
+}
+
```
ta-web-paris: It's a pretty tricky way to do it but it's good! 

You could have done it, in a simpler way like this:
```
  var newArray = [];
  if (array.length === 0) {
    return undefined;
  }
      newArray.push(array[i]);
    }
  }
  return newArray;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,465 @@
...
+        console.log(answer);
+      } else {
+        answer = "false";
+        console.log("false");
+      }
+    }
+  }
+  return answer;
+}
+
```
ta-web-paris: When you are using true and false, you don't need to use "". They are not strings but booleans. It would have solved some problems. :)

If you initialise your variable by a false status, you don't need the else part. Here is a solution: 
```
  var exist = false;
  for (var i = 0; i < wordsArray.length; i++) {
    if (wordsArray[i] === word) {
      exist = true;
    }
  }
  return exist;
}
```

ta-web-paris:
Very good work! You went very far in the lab and you seem to have understood the notions very well! Your code is well organise and easy to read. Continue this way, that's great!

PR #768, opened 2019-01-09
Paris - Kevin B. (Standalone) done with Céline

kevbarns:
Binôme with Céline

@frietpoune


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,10 @@
 // Find the maximum
+var nb1 = 2;
+var nb2 = 3;
+
+function maxOfTwoNumbers(nb1, nb2){
+  return Math.max(nb1, nb2);
+}
```
ta-web-paris: Here you don't need to create variables with real numbers at the beginning. You just need to write the function you did after which is perfectly working.

starter-code/src/functions-and-arrays.js
```diff
@@ -11,15 +17,44 @@ var words = [
...
+  var longest=[];
+  for (i = 0; i < words.length; i++) {
+    if (longest.length < words[i].length) {
+      return longest = words[i];
+    } else if (longest.length == words[i].length){
+      return longest = words[i];
+    }
+  }
+}
+
```
ta-web-paris: Here the else if is unnecessary. You shouldn't return `longest = words[i];` because the return is going to stop the loop. Just delete the return and it's good. 
Then you will have to return you variable `longest` outside the `for loop`.
And to pass all the test, you should add, at the beginning an `if statement` in case the array is empty.

Here is the solution: 
```
  var longest="";
  if (words.length === 0) {
    return undefined;
  }
    if (longest.length < words[i].length) {
      longest = words[i];
    } 
  }
  return longest;
}
```

starter-code/src/functions-and-arrays.js
```diff
@@ -11,15 +17,44 @@ var words = [
...
 // Calculating a Sum
 
+function sumArray(numbers){
+  return numbers.reduce((a,b) => a+b,0);
+}
+
```
ta-web-paris: good use of reduce!

starter-code/src/functions-and-arrays.js
```diff
@@ -34,6 +69,14 @@ var wordsArr = [
 ];
 
 // Unique Arrays
+function uniquifyArray(wordsUnique) {
+  if(wordsUnique.length === 0){
```
ta-web-paris: add a space between your closing parenthese and your curly brace

starter-code/src/functions-and-arrays.js
```diff
@@ -34,6 +69,14 @@ var wordsArr = [
...
+    return undefined
+  } else{
```
ta-web-paris: same here, add a space after the `else`

starter-code/src/functions-and-arrays.js
```diff
@@ -34,6 +69,14 @@ var wordsArr = [
...
+    return unique = [...new Set(wordsUnique)];
+  }
```
ta-web-paris: very good way to do it! You don't need the `unique =` part tough.

starter-code/src/functions-and-arrays.js
```diff
@@ -49,6 +92,15 @@ var wordsUnique = [
 ];
 
 // Finding Elements
+
+function doesWordExist(wordsFind, searchWord) {
+  if (wordsFind.indexOf(searchWord) != -1){
```
ta-web-paris: Here a double `==` would be better, just in case

starter-code/src/functions-and-arrays.js
```diff
@@ -61,6 +113,17 @@ var wordsFind = [
 ];
 
 // Counting Repetion
+
+function howManyTimes(wordsCount, searchWord){
+  var res = [];
+  var idx = wordsCount.indexOf(searchWord);
+  while (idx != -1){
```
ta-web-paris: same here for the `==`!

ta-web-paris:
Very well done! Continue this way!

PR #801, opened 2019-01-28
[AMS] - Isabela and Camila

isabeladowsley:
Not including the bonus assignment.


TA-Ironhack-Matt:
Hi Isabela and Camila, 

I just reviewed your exercise and your almost done:) I found a few mistakes on exercise 3 & 4. Please read my comments and try to correct the exercises: 

- [ex.3] function sumArray is defined but not called, so there is no output > http://prntscr.com/mdalfu
- [ex.3] the sumArray function can only sum two given numbers, but not an array of multiple numbers > http://prntscr.com/mdamt9
- [ex.4] syntax error: the return statement can not be used outside of a function and {} > http://prntscr.com/mdapw6
- [ex.4] try to first solve [ex.3] and then reuse the sumArray function in [ex.4] to divide its outcome with the length of the array that was inserted. 

Let me know if you need any help. 

Regards, 
Matt

PR #853, opened 2019-03-23
Paris, Miguel Bento

MiguelIronHack:
No need to scroll past the grid as the bonus part isn't done.


ta-web-paris:
Your code is really good! Everything is clear and efficient 🙌 

If you want to keep the matrix as a grid and ignore the default settings by prettier, you can add this above the grid `// prettier-ignore`. This rule applies until the next empty line (I think- I might be wrong about that part)

PR #854, opened 2019-03-24
PARIS Finished work-Function&Array-DucNGUYEN

ducky92fr:


starter-code/src/functions-and-arrays.js
```diff
@@ -10,16 +14,39 @@ var words = [
   'orchard',
   'crackpot'
 ];
+function findLongestWord(myArray){
+  if(myArray.length == 0) return undefined;
```
ta-web-paris: you're doing a lot of unneeded check here- your code works without these two lines. You just need to change `>`to be `>=` 

```
function findLongestWord(myArray) {
  if (myArray.length == 0) return undefined;

  return myArray.reduce((acc, currVal) => {
    return (acc.length >= currVal.length) ? acc : currVal
  }, 0
  )
} 
```

starter-code/src/functions-and-arrays.js
```diff
@@ -10,16 +14,39 @@ var words = [
...
+  if(myArray.length == 1) return myArray[0];
+  if(myArray.every(el=>el===0)) return 0;
+  return myArray.reduce((acc,currVal) => acc + currVal,0);
+}
 var numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
 // Calculate the Average
-
+function averageNumbers(myArray) {
+  if(myArray.length == 0) return undefined;
```
ta-web-paris: again here, you're adding unneeded code: 

```
function sumArray(myArray) {
  if (myArray.length === 0) return 0;

  return myArray.reduce((acc, currVal) => acc + currVal, 0);
} 
```

While not needed, it's better to use `===` rather than `==`

starter-code/src/functions-and-arrays.js
```diff
@@ -47,7 +74,10 @@ var wordsUnique = [
   'simple',
   'bring'
 ];
-
+function uniquifyArray(myArray){
+  if(myArray.length == 0) return undefined;
+  return [...new Set(myArray)]
```
ta-web-paris: this is super cool 👍

starter-code/src/functions-and-arrays.js
```diff
@@ -98,3 +140,19 @@ var matrix = [
   [20, 73, 35, 29, 78, 31, 90, 1, 74, 31, 49, 71, 48, 86, 81, 16, 23, 57, 5, 54],
   [1, 70, 54, 71, 83, 51, 54, 69, 16, 92, 33, 48, 61, 43, 52, 1, 89, 19, 67, 48]
 ];
+
+function greatestProduct(myArray){
```
ta-web-paris: Your bonus doesn't quiet work- this part is hard and the instructions are not super clear. You need to loop of the matrix (a for loop inside a for loop is the easiest way to do this) 

by returning 16 directly is cheating the tests meaning it won't work if Jasmine changes the input.

I'm happy to review this part with you, so you can speak to me (Abi) during a break or lunch, or I can send you a copy of the solution code. But that depends on if you want to try again on your own first

ta-web-paris:
Your code is good 🙌

PR #855, opened 2019-03-24
PARIS Daniel Diaz

DaniTOLETUM:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,100 +1,180 @@
...
-  'palace'
-];
+var wordsArr = [ 'seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace' ];
+
+function averageWordLength(arr) {
+	var sumator = 0;
+	if (arr.length === 0) {
+		return undefined;
+	} else {
+		for (let i = 0; i < arr.length; i++) {
```
ta-web-paris: while this works, using a reduce would be better. A reduce is a really powerful array method, but it also signified intent to other people who are reading the code. If they see a reduce, they know what you're trying to do without reading through the loop

ta-web-paris:
it looks good and everything works 🙌 

You should try to use different array methods (forEach, reduce etc) when possible - once you've understood them, your code will become a lot clearer and you're less likely to forget parts of the code

PR #856, opened 2019-03-25
PAR alex

alexandfox:
please review, thank you!!


starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +83,24 @@ var wordsUnique = [
   'bring'
 ];
 
+function uniquifyArray( wordsArr ) {
+  var noDuplicates = []
+  if (!wordsArr.length) return;
+
+  wordsArr.forEach( (word, i) => {
```
ta-web-paris: You didn't use the `i` here so you could remove it

ta-web-paris:
everything works and it's easy to follow 🙌 

Changing your functions into arrow functions could help keep your code shorter because then you need to use less return statements etc

PR #873, opened 2019-04-03
MIA FTWD - Nour Cherif Essoussi - April 2019

Nourcs:


ta-web-mia:
Good job with the bonus iteration! You definitely have a good grasp on JS array methods. Keep up the good work!

PR #874, opened 2019-04-03
MIA-FTWB-April Parker Cooper

parkcoop:


ta-web-mia:
Great job! Glad to see you're becoming more comfortable with the forEach method in addition to the other array methods. Keep it up!

PR #875, opened 2019-04-03
MIA FTWD - Angel Biondi

AngelBiondi:


ta-web-mia:
Well done! Although there was a typo repeated multiple times that prevented the spec runner from checking you off. Your 5 word arrays were all missing commas in the exact same spots. Please pay closer attention to the red underlines and the error counter in the bottom left from now on. Other than that, all of your functions ran as intended. Keep up the good work!

PR #876, opened 2019-04-03
MIA-FTWD-APR- Justin Velez

jvelez523:
Done


ta-web-mia:
Well done! Very happy with the ES6 syntax you used for the uniquify array function. Becoming more comfortable with shorthand like that will end up saving you time in the future. Keep up the good work!

PR #877, opened 2019-04-03
MIA - FTWD - APR - Nathaniel Carvajal

nathancarv:


ta-web-mia:
You're getting there! The uniquify array function is definitely the hardest. In order to solve that one, you would need to utilize the indexOf() array method, and declare a separate variable for another array. If you ever need more practice on these types of algorithms, CodeWars has no shortage of them. The effort you're putting into all of this isn't going unnoticed. Keep up the hard work!

PR #879, opened 2019-04-03
MIA FTWD - Christian Khawam

CodeByCK:


ta-web-mia:
Good job! The only thing that can really be improved on would be to use newer ES6 (and beyond) syntax. But I think that goes for just about everyone, myself included. Glad to see you have a good handle on algorithms. Keep up the good work!

PR #881, opened 2019-04-03
MIA FTWD APR Daniela Olano

dolano13:


ta-web-mia:
Good job! Really happy to see that you're using console.log for debugging purposes. With more difficult algorithms, it's basically impossible to get the whole thing solved without checking your output; so this is definitely a good habit to get into early on. Keep up the good work 👍

PR #882, opened 2019-04-03
MIA-FTWD-APR-Yulia

salapjaka:
done without the bonus


ta-web-mia:
Good job! Just a reminder that closing curly braces should usually be placed on their own line so that the code doesn't get too messy. For example, instead of this:

image

Format it like this instead:

image

It's mostly a nitpick, but improved readability will help out a lot once your code gets much more massive. Keep up the good work!

PR #883, opened 2019-04-04
MIA FTWD APR- Katy Brown

katybrown91:
initial submission


ta-web-mia:
All of your functions seem correct, but the spec runner just isn't passing any of the tests for some reason. I'm not sure if you changed any of the files, but just a reminder that with activities like this you should only alter the JS file it's looking over. It's very possible that Jasmine is just being difficult for no reason as well. Either way, the code you have seems pretty good. Keep up the good work!

PR #884, opened 2019-04-04
MIA FTWD ANGHY POMAR

apomarn:


ta-web-mia:
Nice job! Really impressed that you managed to get the bonus as well. Don't really have much to critique here, but I'm happy that you're getting more comfortable with JavaScript algorithms. Keep it up!

PR #885, opened 2019-04-04
MIAFTWD - April - Chris C

ccardelle:


ta-web-mia:
Good job! The uniquify array function is pretty tricky, so don't get too hung up on not quite solving it. The key to solving it is using an empty array and the indexOf() method. You can use indexOf to check if an array item exists in the empty array, pushing it if it doesn't, and then the originally empty array will contain each item only once. Here's an example of what it would look like:

function uniquifyArray(wordsUnique) {
  if (wordsUnique.length > 0) {
    var result = [];
    for (i = 0; i < wordsUnique.length; i++) {
      console.log(result.indexOf(wordsUnique[i]));
      if (result.indexOf(wordsUnique[i]) === -1) {
        result.push(wordsUnique[i]);
      }
    }
    return result;
  }
}

Everything else I saw in here looked really good though. Keep up the good work!

PR #886, opened 2019-04-04
[MIA FTWD] (Jon R.)

jonriquer:


ta-web-mia:
Good job! Ran into one tiny error though; on your longest word function, you declared the longest word variable within the loop like so:

`function findLongestWord(words) {

  
  for (i=0; i<words.length; i++) {
    var longestWord = words[0];
      longestWord = words[i]; 
    }
  }
  return longestWord;
}`

doing so will reset the value of the longest word each time the loop repeats, meaning it will always confuse the word at the first index as the longest word. To fix this, simply move the declaration outside of the loop:

`function findLongestWord(words) {

  var longestWord = words[0];
  
  for (i=0; i<words.length; i++) {
      longestWord = words[i]; 
    }
  }
  return longestWord;
}`

That's the only error I could find, since everything else looked really good. Keep up the good work!

PR #888, opened 2019-04-10
MIA-FTWD-Allison

allisonkopp:


ta-web-mia:
Good job! Don't worry about the bonus too much, I would say the array methods are much more important. If you feel good about that, then you got a lot out of the exercise. Keep up the good work!

PR #914, opened 2019-05-29
[PAR-FTWD-Elin]

elinlarsen:
all without the bonus quest

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,5 +1,22 @@
 // Find the maximum
 
+function maxOfTwoNumbers(num1, num2){
+  var largest=0;
+  if (num1>num2){
+    console.log("The largest number is: "+ num1);
+    largest=num1;
+  }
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -48,6 +130,21 @@ var wordsUnique = [
   'bring'
 ];
 
+
+function uniquifyArray(w){
```
ta-web-paris: Pour passer Jasmin a coup sur:

`function uniquifyArray(w) {
  var uniqueArray = [...new Set(w)];
  return uniqueArray.length === 0 ? undefined : uniqueArray;
}`

https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Set
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

ta-web-paris:
Good Job. RAS

PR #916, opened 2019-05-29
lab-javascript-functions-and-arrays - camille c & clara

clarabesset:
Clara and Camille C - works but Jasmine almost there!


clarabesset:
final -

ta-web-paris:
Nice work, Jasmine was not ok with two stuff : 

//  averageNumbers(numbersAvg) you have to initialize the variable sumAv inside the function otherwise, each time the function executed the variable will contain the result of the previous execution


//function averageWordLength  the function was not returning anything so jasmin was not aware of the result

PR #947, opened 2019-06-06
[MAD DE] Rubén Vallejo

rubander:
The worst thing about this exercise was to understand how some of the functions organize in order to search inside the arrays. 

For me, the uniquifyArray was the hardest, and if it wasnt for a strange in Internet that shared his example in StackOverflow, i would have never passed that test. I could not really understand how to get the unique words of an array and "delete" the ones that were repeated. I know that it involved the use of push() but I didnt know how to apply it.


ta-web-mad:
Well done!!

Sito🤪

PR #964, opened 2019-06-09
MIA-PTWD-AlejandroSalas

alejotsv:
I completed the bonus, but I am still getting 2 failures on the Jasmine test, so I have a couple of questions:

1) Was the function supposed to run for one direction at a time? I created the function expecting four parameters: matrix, row, column, and direction

2) The failures I get are:
Return 1 when all the numbers of the arrays are 1 - I tested the function with a set of four 1's and it returned 1.

Return 16 when all the numbers of the arrays are 2 - why would we want it to return 16 instead of 2?

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,6 +1,27 @@
 // Find the maximum
+function maxOfTwoNumbers(num1,num2){
+  if (num1 > num2){
+    return num1;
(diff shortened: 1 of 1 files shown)


ta-web-mia:
Looks good, nice work! I agree with you that I think the "return 16" isn't correct. I'll look into the source code for the tests and see what I can find out.

PR #965, opened 2019-06-09
Fabricio Policarpo mia-ptwd

fpolica91:
Hello I would really appreciate if you guys could go over the bonus during next class, it has more than my curiosity.


ta-web-mia:
Looks pretty good. You're missing some functionality, but a majority of the tests pass.

PR #967, opened 2019-06-10
MIA PTWD Jose De la Cruz

jdelacruzjose:


ta-web-mia:
Make sure to account for edge cases like handling empty arrays in your code, but other than that it looks good.

PR #969, opened 2019-06-11
PTWD Arrrays and Functions  - Henry Hoyos

HDOSURREGO:


ta-web-mia:
You're only failing the last 3 cases (not including the bonus section). Good work! Try to finish it completely at some point though.

PR #975, opened 2019-06-11
MIA-FTWD- Candido Concepcion with Taluama velludo

candidocon:


starter-code/src/functions-and-arrays.js
```diff
@@ -34,6 +75,20 @@ var wordsArr = [
 ];
 
 // Unique Arrays
+function uniquifyArray(arrayWithDuplicates){
```
ta-web-mia: Nice work, this one can be a bit tricky. Just so you know, this can be simplified to a single conditional nested within a for loop. It would look something like this: 

let unique = []

for (i = 0; i < arrayWithDuplicates.length; i++){
  if (unique.indexOf(arrayWithDuplicates[i]) === -1) {
    unique.push(arrayWithDuplicates[i])
  }
}

As a general rule of thumb, it's good to maintain the original set of data without splicing or removing any elements from it as well. Just a little something to keep in mind.

starter-code/src/functions-and-arrays.js
```diff
@@ -75,7 +151,28 @@ var wordsCount = [
   'matter'
 ];
 // Bonus Quest
+function greatestProduct(matrix){
```
ta-web-mia: Well done!

ta-web-mia:
Nice job, I'm glad you were able to get through all of these. Keep up the good work, and let me know if you have any questions 👍 
- Nick

PR #977, opened 2019-06-11
Serena + Matt FTWD functions and arrays 

SDominguez21:


ta-web-mia:
So far so good, but the SpecRunner is unfortunately being difficult. It seems like most of your code is correct and there are no errors popping up, but something is causing the tests to return as failures. If you fix this and make any additions to the program, let me know and I can take a quick look. Keep it up 👍 
- Nick

PR #980, opened 2019-06-11
MIA-FTWD William Bennett

william1benn:


MarcosPalaciosIII:
Great job on the function and arrays lab. When working with test driven development, you sometimes don't get a clear response as to what is expected. But it is a good way to assure that specific requirements are met when writing code for an app. Primarily used when attempting to add new features to an already existing app that something broken can effect millions of users.

Remember to give credit to people that worked on projects with you, either in the title, comments, notes or the readme of the app. 

Keep up the great work.

PR #981, opened 2019-06-11
MIA-FTWD Anabel/Artem

anabelpdv:


MarcosPalaciosIII:
Great job you two. When working with test driven development, you sometimes don't get a clear response as to what is expected. But it is a good way to assure that specific requirements are met when writing code for an app. Primarily used when attempting to add new features to an already existing app that something broken can effect millions of users.

Keep up the great work.

PR #982, opened 2019-06-11
mia-ftwd Manuel-William

manuelga22:


starter-code/src/functions-and-arrays.js
```diff
@@ -10,15 +14,41 @@ var words = [
...
+  for(var a =0;a<numbers.length;a++){
+    sum += numbers[a];
+  }
+  return sum;
+}
 // Calculate the Average
 
 var numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(numbersAv){
```
ta-web-mia: In this function sometimes you call numbersAvg 'numbersAv', and it crashes the application. The function works, otherwise.

starter-code/src/functions-and-arrays.js
```diff
@@ -47,6 +85,18 @@ var wordsUnique = [
   'simple',
   'bring'
 ];
+function uniquifyArray(array){
```
ta-web-mia: Your function is not bringing back 'communion' for its list of unique words.

ta-web-mia:
It's good except for the uniquifyWords array. It isn't bringing back the final unique word. Other than that and the typo, good job!

PR #985, opened 2019-06-12
MIA-FTWD-Nicolle/Billy

nicollegonzalez:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,4 +1,7 @@
 // Find the maximum
+function maxOfTwoNumbers (arg1, arg2) {
+  return Math.max(arg1, arg2);
```
ta-web-mia: Clever! This can also be done using a conditional to compare the two numbers directly, but using the built in Math.max is more efficient.

starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +14,73 @@ var words = [
...
+  'palace'
+];
+
+function averageWordLength(arr) {
+  if( arr == false) {
+    return undefined;
+  }
+  let arrayWordLength = [];
+  for (i=0; i<arr.length; i++) {
+    arrayWordLength.unshift(arr[i].length);
```
ta-web-mia: Nothing wrong with using unshift to add elements to the array, but it's worth nothing that push is more commonly used, and often considered best practice.

starter-code/src/functions-and-arrays.js
```diff
@@ -60,6 +137,26 @@ var wordsFind = [
...
 
+function doesWordExist( arr, str){
+  //console.log(typeof(arr));
+  if( arr === false) {
+    return false;
+  }
+  let count = 0;
+  for(let i = 0; i < arr.length; i++) {
+    if(str === arr[i]){
+      count ++;
```
ta-web-mia: Since the function is asking you to return true if the word is in the array, you can actually just return true here and exit the function. No need to iterate, as anything greater than 0 will return true anyway.

ta-web-mia:
Well done! Take a look at my feedback when you two have the chance. Keep up the good work!
- Nick

PR #1005, opened 2019-06-15
PTWD - MIA - Nelly Gonzalez-Vinas

nellygvinas:
Had difficulty with the spec tests. Most of my original codes worked, but once the spec tests were run, there were various issues. Had to rewrite code, and used some of the examples in class to finish assignment.


ta-web-mia:
Excellent work!

PR #1014, opened 2019-06-20
MIA PTWD  Rachel Feldman

RFeldman84:


ta-web-mia:
So far so good! This should give you a better idea of how to declare functions and work with arrays. The average word length function would work very similarly to the average value function. As a reminder, the .length property can be used on strings as well as arrays. Keep up the good work 👍

PR #1035, opened 2019-06-30
MIA PTWD Alina

AlinaZakirod:
functions


ta-web-mia:
Nice job! The bonus iteration is really tricky, and I wouldn't worry too much about it right now. In case you're curious, here's a code snippet that would pass the tests for the bonus problem: 

`function greatestProduct(matrix){
  let max = 0;
  for(let i=0;i<matrix.length-4;i++){
    for(let j=0;j<matrix.length-4;j++){
      let product1 = matrix[i][j]*matrix[i][j+1]*matrix[i][j+2]*matrix[i][j+3];
      let product2=matrix[i][j]*matrix[i+1][j]*matrix[i+2][j]*matrix[i+3][j];

       if(product1 > max){
        max = product1;
      }
      if(product2 > max){
        max = product2;
      }
    }
  }
  return max;
}`

Nested for loops are very important when looping through 2d arrays like this. It's nothing you'll run into too often, but it's certainly good practice. Keep up the good work 👍

PR #1072, opened 2019-08-13
updated tests w/ correct return values for empty arrays

mjarraya:


sandrabosk:
@mjarraya Thank you for making the update!

PR #1086, opened 2019-08-14
FTWD PAR - Aimée

aimeestone:


ta-web-paris:
  var wordCount = 0;
  if (array.length === 0) {
    return false;
  }
    if (array[i] === word) {
      wordcount += 1;
    }
  }
  return wordCount;
}

------------------------------------------------------------------------------
  wordcount += 1;   Should be wordCount : )
-------------------------------------------------------------------------------

PR #1092, opened 2019-08-15
add first draft lab

PaulaMarcella:
I did not finish everything in time but will keep on working on it.
i did not get how to do this: "returns the longest occurrence when it has multiple words"


GTCarmona:
Hey Paula!

You already have all the logic and code ready, but your "return" statement was inside the FOR loop, and should be outside, like this


```
function findLongestWord(array) {
  var longest = "";
  if (array.length === 0) {
    return null;
  } else {
    for (i of array) {
      if (i.length > longest.length) {
        longest = i;
      }

    }
      return longest;
  }
}
```


Functions can get really messy with all the brackets, parenthesis and etc. I suggest you always keep some clear space between lines, and using the "Prettier" extension on Visual Code to help you keep the code clean

Let me know if you go back to finish this exercise!

PR #1096, opened 2019-08-17
LIS-FTWD-[DANIEL VALERO] #156

damvalero:
added function averageWordLength

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -93,11 +93,11 @@ var wordsUnique = [
   'bring'
 ];
 
-const uniqueArray= words => {
+const uniqueArray= wordsUnique => {
   let tempArray=[];
   for (let i = 0; i< words.length; i++){
-    if (tempArray.indexOf(words[i]) < 0){
-      tempArray.push(words[i]);
+    if (tempArray.indexOf(wordsUnique[i]) < 0){
+      tempArray.push(wordsUnique[i]);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


GTCarmona:
Alright, let's go step by step

## averageNumbers function

You got everything right, just a tiny typo error 

```
const averageNumbers =numbers =>{
  if(!numbers.length) return null;
  return sumArray(numbers)/numbers.length;
}
```

you had it "sumaArray" instead of "sumArray", so just a typo

## Uniquify array

You were so clooooose!!! you just had 1 typo, 1 syntax error and you were missing the return step of the function

so this is what you had


```
const uniqueArray= wordsUnique => {
  let tempArray=[];
  for (let i = 0; i< words.length; i++){
    if (tempArray.indexOf(wordsUnique[i]) < 0){
      tempArray.push(wordsUnique[i]);
    }
  }
}

```
And this is the correct version. I changed the name of the function, that needs to be the same 
```
const uniquifyArray= wordsUnique => {
  let tempArray=[];
  for (let i = 0; i< wordsUnique.length; i++){
    if (tempArray.indexOf(wordsUnique[i]) < 0){
      tempArray.push(wordsUnique[i]);
    }
  }
  return tempArray
}
```

## doesWordExist function 

again, your logic was really good! just some tiny syntax stuff. I made a few changes and now it's passing the tests, take a look

```
function doesWordExist (wordsFind, word){
  for ( let item of wordsFind){  
    if ( item === word) {
      return true
      }
    }
      return false;
    }
  
```

And this is what you had before

```
function doesWordExist (wordsFind, word){
  for ( let item of wordsFind){  
  //console.log(wordsFind[i])
    if ( item === word){
      //console.log(item);
      return true;
      
    }else {
      return false;
    }
  }
};
```
I think you just had some brackets "{}" in the wrong place.

So, to be honest you did great, just some stuff that wasn't 100% and tests didn't pass, but in practice you got the logic right and that is the most important part, the syntax stuff comes with practice

keep it up!

PR #1098, opened 2019-08-18
LIS-FTWD-Andre Almeida-#156

andreH2M:


GTCarmona:
Hey Andre,

Good job!! for there was some random leftover that was making everything break, for example a "return" outside a function.

I saw that you used for loops for your solutions, which is great! I just encourage you to take a look again at "map" and "filter" as they are the best alternatives. 

Great job, keep it up!!

PR #1184, opened 2019-10-15
[BER] [Daniel] Almost satisfied Jasmin

daneklustig:


Svemakawe:
Nice Job on that lab! Maybe try to start using methods like forEach instead of the for-loop to get used to it.  Sometimes you can shorten your code a little, f.eg. in the doesWordExist function you can just return stringArray.includes(word). Its already giving back a boolean and you don’t need the if/else.

PR #1185, opened 2019-10-15
[Berlin-FTWD-William]

whbrown:
all challenges + bonus


Svemakawe:
Good Job (also solving the bonus)! Nice that you use forEAch, although I think for „maxOfTwoNumbers“ it makes your code unnecessarily long. With ternary this could be a one-liner: return num1 > num2 ? Num1 : num2 or Math.max(num1,num2).

PR #1192, opened 2019-10-15
BER_FTWD_Pascal

PascalH91:


starter-code/src/functions-and-arrays.js
```diff
@@ -11,14 +18,47 @@ const words = [
...
+    }
+  };
+  for (let i = 0; i < wordList.length; i++) {
+    if (wordList[i].length === longestLength) {
+      return wordList[i];
+    }
+  };
+};
+
+
```
pierreportal: `...
if (wordList[i].length >= longestLength) {

      longestLength = wordList[i].length;

    }
...`

PR #1196, opened 2019-10-15
[BER-FTWB-Maria P]

mariakristiina:


DalinaLeila:
Good Job 👍 ! You can take out one of the if statements in the "averageNumbers" function and in the" findLongestWord" function, because if the array only contains one value, the output will automatically be this value.

PR #1197, opened 2019-10-15
[BER-FTWD-Davide]

davideberno:


DalinaLeila:
Nice work 👍 !! I like that you used a forEach loop and the reduce function. 💯 

Always remember that you want to give your variables the most recognizable name possible, so when you look back at your code you will be able to understand what you are storing. In the future when you have bigger projects and need to pass data into different js files it can become very tricky if you don't remember what you stored.

e.g. Instead of calling it "cntrlWord" maybe call the variable "longestWord".

PR #1198, opened 2019-10-15
[BER-WDFT-102019-Michael]

MikeVanYarnrider:
all except the bonus, yet

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,101 +1,487 @@
 // Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 < num2) {
+    return num2;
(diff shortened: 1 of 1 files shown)


DalinaLeila:
Nice work! 👍 💯 

 // Find the maximum
--> you can remove the else if statement because in this case, it will automatically return the latter number. 

NEW VERSION:
  if (num1 < num2) {
    return num2;
  }
  return num1;
}

//find the longest word
There is no need to push and store the length of each word in an array. You can directly compare the length of one word to the length of the word stored in the variable.

MikeVanYarnrider:
Thank you! I already figured out that I can get rid of a lot of code…tried to refactor a bit this morning, it was clearer to me after a couple of hours of sleep.

Thx for checking.

PR #1199, opened 2019-10-15
[BER-FLWD-2019] Nino

NinoGlonti:


DalinaLeila:
Nice work 👍 ! 

For the function "findLongestWord", in order to get rid of the error shown in Jasmine you have to return your variable outside of the loop. 

NEW VERSION:
`
  let longestWord = "";
  if (words.length === 0) {
    return null;
  }
    if (longestWord.length < words[i].length) {
      longestWord = words[i];
    }
  }
  return longestWord;
}`

PR #1217, opened 2019-10-18
Final version - Andrew and Parina

ParinaKh:
Hey Guys, 

We've completed the whole exercice except of the bonus one ;)

Thanks


ta-web-paris:
Great job - you just need to add condition if the array is empty for findLongestWord, averageNumbers & averageWordLength :)

PR #1218, opened 2019-10-20
[PAR_FTWD] Nicolas

Empty-n:
No bonus. But i don't give up !


ta-web-paris:
Good job! I'll mark you as done but leave this open in case you want to do the bonus in the next few days :)

PR #1219, opened 2019-10-20
[PAR_FTWD] Pierre-Henri

Jepassaispar:
Couldn't get the bonus, still 1 failure on jasmine


ta-web-paris:
Basically it's all good - but for uniquifyArray, try to use a for loop (Jasmine wants the new array to have the same order, but without duplicates). You're almost there :)

PR #1223, opened 2019-10-21
MEX-PTWD-EDGAR ZEA - LIZBETH MAGAÑA

cesszea:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,5 +1,11 @@
 // Find the maximum
-
+function maxOfTwoNumbers (number1, number2) {
```
ta-web-mex: this function can be simplified
```
if (number>=number2) {
return number1;
}

return number2
```

ta-web-mex:
Cool, just missing the bonus quest but its fine, congrats :)

PR #1228, opened 2019-10-22
[MX-PTWD-102019] - Tonalli R. & Andres C.

t3rodrig:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,6 +1,32 @@
 // Find the maximum
 
+function maxOfTwoNumbers(number1, number2){
```
ta-web-mex: good, but I think that you can simplify this function

starter-code/src/functions-and-arrays.js
```diff
@@ -13,13 +39,39 @@ const words = [
...
 
 // Calculate the Average
 
+function averageNumbers(numbersArray) {
+  if (numbersArray.length === 0){
+    return null;
+  }
+  let sumTotal = sumArray(numbersArray);
+  let average = sumTotal / numbersArray.length;
+  return average;
```
ta-web-mex: also we can return like this `return (sumTotal / numbersArray.length);`

starter-code/src/functions-and-arrays.js
```diff
@@ -34,6 +86,16 @@ const wordsArr = [
 ];
 
 // Unique Arrays
+
+function uniquifyArray(wordsArray){
+  if (wordsArray.length === 0) {
+    return [];
+  }
+  let newArray;
+  newArray = wordsArray.filter((item, index)=> wordsArray.indexOf(item) === index);
```
ta-web-mex: nice

PR #1229, opened 2019-10-22
MIA FTWD Masaki Endo

mendosaurus:


starter-code/src/functions-and-arrays.js
```diff
@@ -33,7 +96,20 @@ const wordsArr = [
   'palace'
 ];
 
+function averageWordLength(wordsArr) {
+  if (wordsArr.length !== 0) {
+    let sum = wordsArr.reduce((accum,currentValue) => accum + currentValue).length
```
ta-web-mia: Nice use of reduce!

ta-web-mia:
Nice job! I'm glad to see you're already comfortable with so many different array methods. Keep up the good work :)
- Nick

PR #1230, opened 2019-10-22
mia-ftwd oct2019 Gil

CodesGM:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,101 +1,469 @@
...
+    19,
+    67,
+    48,
+  ],
 ];
+
+function greatestProduct(arr) {
+  let multi = 0;
+  for (let i = 0; i < arr.length; i += 1) {
+    for (let j = 0; j < arr[i].length; j += 1) {
```
ta-web-mia: You're on the right track! You'll have to make similar nested for loops to check for vertical and diagonal combinations as well.

ta-web-mia:
Nice work! Keep it up :)
- Nick

PR #1240, opened 2019-10-22
MIA FTWD Oct Nico Bared

nbared:


starter-code/src/functions-and-arrays.js
```diff
@@ -75,8 +156,36 @@ const wordsCount = [
...
+      if (arr[i] === word) {
+        counter++
+      } 
+    } return counter
+  }
+  howManyTimes(words, 'matter')
+
 // Bonus
 
+function greatestProduct(){
```
ta-web-mia: You're definitely on the right track! If you have a chance to go back to this function, you'll end up having to make similar loops to look for vertical and diagonal combinations. It's very redundant, so don't worry too much about it!

ta-web-mia:
Nice job! If you ever have any questions, please let me know. Keep it up :)
- Nick

PR #1274, opened 2019-10-28
[PTWDMIA102019] Ashraf

alforsii:
Done 1->7.


MarcosPalaciosIII:

Keep up the great work.

alforsii:
agree:) day by day I'm falling in love with JS coding since I start understanding more

PR #1317, opened 2019-11-16
[ING-FT-112019] - Tomasz

twalasek:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,40 @@
...
+  return Math.max(a, b);
+}
+
 // Iteration #2: Find longest word
+
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+//console.log ("The longest: " + findLongestWord(words));
+function findLongestWord (words) {
+  let longest = "";
+  if (typeof words === "undefined" || !Array.isArray(words) || words.length === 0)
```
MarcoSantonastasi: first condition may be unnecessary

twalasek: Agreed. Thank you.

PR #1319, opened 2019-11-19
ING-juliane-dirk

dirk-biermann:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,58 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers( firstNumber, secondNumber ) {
```
MarcoSantonastasi: how would you do it using built-in Math library?

starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,58 @@
...
+function maxOfTwoNumbers2( firstNumber, secondNumber ) {
+  return (firstNumber > secondNumber) ? firstNumber : secondNumber;
+}
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+function findLongestWord( array ){
+  if ( array.length != 0 ) {
+    array.sort(
```
MarcoSantonastasi: sehr gut!

starter-code/src/functions-and-arrays.js
```diff
@@ -40,6 +94,19 @@ const wordsUnique = [
...
 
+function uniquifyArray ( array ) {
+  let destArray = [];
+  if(array.length!=0) {
+    for( let i=0; i<array.length; i++) {
+      if( destArray.indexOf( array[i] ) < 0 ) {
+        destArray.push( array[i] );
+      }
+    }
+  }
```
MarcoSantonastasi: looks like there is a missing "else"...

PR #1320, opened 2019-11-19
ING-Anastasia-Jürgen-Alina

kilianal:
Iteration 1 complete

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,5 +1,17 @@
 // Iteration #1: Find the maximum
 
+
+function maxOfTwoNumbers(number1, number2){
+  if(number1>number2){
+    return number1;
+  }
+  else{
+    return number2;
+  }
+}
+
+maxOfTwoNumbers(5,6);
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,80 @@
 // Iteration #1: Find the maximum
 
+
+function maxOfTwoNumbers(number1, number2){
```
MarcoSantonastasi: how would you do it with a built in Math library?

starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,80 @@
...
+findLongestWord(words);
+
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(numberArray){
+  if(numberArray.length===0) return 0;
+   
+  let sum= numberArray[0];
```
MarcoSantonastasi: very good!

PR #1323, opened 2019-11-19
ING_Malte_Sergej

S-Sergej:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,52 @@
...
+  return finalSum;
+}
+console.log(sumArray(numbers));
+
+
+// Iteration #4: Calculate the average 
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(numArray){
```
MarcoSantonastasi: how could you do it by reusing the sumArray() function?

S-Sergej: thx for hint @MarcoSantonastasi :) got it

PR #1350, opened 2020-01-07
[20-01-07-BER - FT] - Jonathan

JonathanSaudhof:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,69 @@
...
+
+function findLongestWord(wordsList) {
+  let temp = '';
+
+  wordsList.forEach((word) => {
+    word.length > temp.length ? (temp = word) : false;
+  });
+
+  return temp ? temp : null;
+}
```
DalinaLeila: nice that you use the ternary operator here  👍!

starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,69 @@
...
+function sumArray(numbersArray) {
+  let sum = 0;
+
+  numbersArray.forEach((element) => {
+    sum += element;
+  });
+
+  return sum;
+}
+
```
DalinaLeila: 👍

starter-code/src/functions-and-arrays.js
```diff
@@ -37,9 +102,19 @@ const wordsUnique = [
   'poison',
   'communion',
   'simple',
-  'bring'
+  'bring',
 ];
 
+function uniquifyArray(wordsList) {
```
DalinaLeila: have a look at `... new Set`, this might be useful here:)

PR #1351, opened 2020-01-07
WebDevFT-0120 Vaso

vasilikibalidou:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,460 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
```
DalinaLeila: have a look at `Math.max()` this might be helpful here :)

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,460 @@
...
 
+  return longest;
+}
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
-
+function sumArray(arr) {
+  let sum = 0;
+  if (arr.length === 0) {
```
DalinaLeila: you can take this away, and would still get 0 if the length of the arr is 0 because the sum that you declared on line 38 is = 0

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,460 @@
...
+  "fuel",
+  "palace"
 ];
+function averageWordLength(arr) {
+  let letterSum = 0;
+  let wordSum = 0;
+  if (arr.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < arr.length; i++) {
```
DalinaLeila: here you are just checking the number of words in the arr, which you do not need to use a loop. You can simply check this by writing `arr.length`

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,460 @@
...
+  "sharp",
+  "playground",
+  "poison",
+  "communion",
+  "simple",
+  "bring"
 ];
+function uniquifyArray(arr) {
+  if (arr.length === 0) {
+    return [];
```
DalinaLeila: return null ?

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,460 @@
...
+  "subset",
+  "trouble",
+  "starting",
+  "matter",
+  "eating",
+  "truth",
+  "disobedience"
 ];
+function doesWordExist(arr, word) {
+  return arr.includes(word);
```
DalinaLeila: 👍

PR #1352, opened 2020-01-07
[BER-FT-012020] - Jacqueline C.

chenjacqueline:
Submitting without the bonus; still working on that one!

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -4,10 +4,8 @@
 function maxOfTwoNumbers(num1, num2) {
   if (num1 > num2) {
     return num1;
-  } else if (num2 > num1) {
-    return num2;
   } else {
-    return 4; // Jasmine expected this to return 4. But why 4?
+    return num2;
   }
 }
 
@@ -24,17 +22,15 @@ const words = [
 ];
 
 function findLongestWord(arr) {
+  // findLongestWord returns null when called with an empty array:
   if (arr.length === 0) {
     return null;
-  } else if (arr.length === 1) {
-    return arr[0];
   }
-  let longestWordLength = 0;
+
   let longestWord = "";
   arr.forEach(function(element) {
-    if (element.length > longestWordLength) {
+    if (element.length > longestWord.length) {
       longestWord = element;
-      longestWordLength = element.length;
     }
   });
   return longestWord;
@@ -63,6 +59,11 @@ sumArray(numbers);
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
 function averageNumbers(arr) {
+  // Function should return null if called with an empty array
+  if (!arr.length) {
+    return null;
+  }
+
   let sum = 0;
   arr.forEach(function(num) {
     sum += num;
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,479 @@
 // Iteration #1: Find the maximum
+// Define a function maxOfTwoNumbers that takes two numbers as arguments and returns the largest.
+
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
+  } else {
+    return 4; // Jasmine expected this to return 4. But why 4?
```
DalinaLeila: jasmine wanted 4 because of the numbers they test your code with. you do not need this here though. 

Simply writing this 
`if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }`

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,479 @@
...
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  if (arr.length === 0) {
+    return null;
+  } else if (arr.length === 1) {
+    return arr[0];
+  }
+  let longestWordLength = 0;
```
DalinaLeila: you do not need 2 variables here

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,479 @@
...
+  let longestWord = "";
+  arr.forEach(function(element) {
+    if (element.length > longestWordLength) {
```
DalinaLeila: instead of longestWordLength you can write longestWord.length

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,479 @@
...
+      longestWord = element;
+      longestWordLength = element.length;
```
DalinaLeila: you can also remove this ` longestWordLength = element.length;`

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,479 @@
...
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  if (arr.length === 0) {
+    return null;
+  } else if (arr.length === 1) {
```
DalinaLeila: the `else if` including that code can also be removed. It will automatically return the first word if there is only one in the arr, because of your forEach loop.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,479 @@
...
+sumArray(numbers);
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
+// Declare a function named averageNumbers that expects an array of numbers and returns the average of the numbers:
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arr) {
+  let sum = 0;
+  arr.forEach(function(num) {
```
DalinaLeila: here you can just use the function you created above `sumArray(numbersAvg)` instead of writing the same code (forEach loop) as above

PR #1354, opened 2020-01-07
[BER-FT]- Ozhan

ozhann:
Done.

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,93 +1,454 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(x, y) {
+  return Math.max(x, y);
+}
+
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  let longest = "";
+  if (arr.length === 0) {
+    return null;
+  } else {
+    for (i = 0; i < arr.length; i++) {
+      if (arr[i].length > longest.length) {
+        longest = arr[i];
+      }
+    }
+  }
+  return longest;
+}
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(arr1) {
+  let sum = 0;
+  for (let i = 0; i < arr1.length; i++) {
+    sum += arr1[i];
+  }
+  return sum;
+}
+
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,454 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(x, y) {
+  return Math.max(x, y);
```
DalinaLeila: 💯

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,454 @@
...
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  let longest = "";
+  if (arr.length === 0) {
+    return null;
+  } else {
```
DalinaLeila: no need to have the else here because you have a return in the if. 

the rest will only get evaluated if the if is not evaluating to true

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,454 @@
...
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(numArr) {
+  let sum = 0;
+  if (numArr.length === 0) {
+    return null;
+  } else
```
DalinaLeila: also here no need to have the else because of the return on line 51. 

Also you instead of looping to get the sum you can use the function from before `sumArray(numbersAvg)`

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,454 @@
...
+  "subset",
+  "trouble",
+  "starting",
+  "matter",
+  "eating",
+  "truth",
+  "disobedience"
 ];
 
+function doesWordExist(arr3, str2) {
```
DalinaLeila: why arr3 and str2? Maybe it's better to simply name it array, string or just arr, str

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,454 @@
...
+  if (arr3.length === 0) {
+    return false;
+  }
+
+  if (arr3.includes(str2)) {
```
DalinaLeila: includes already returns a boolean by default so you do not need the if/else conditions. 

simply to `return arr3.includes(str2)`

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,454 @@
...
+  "starting",
+  "matter",
+  "eating",
+  "matter",
+  "truth",
+  "disobedience",
+  "matter"
 ];
 
+function howManyTimes(arr4, str3) {
```
DalinaLeila: see above for naming variables. the parameters are just placeholders so you can give them the same name. They do not need to be different.

PR #1356, opened 2020-01-07
[Ber FT] Alexis

Perdikal:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,435 @@
 // Iteration #1: Find the maximum
-
+let maxOfTwoNumbers = (a, b) => {
+  return a > b ? a : b;
```
DalinaLeila: 💯

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,435 @@
...
+  return longest;
+};
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+let sumArray = arr => {
+  if (arr.length === 0) {
+    return 0;
+  }
+  let sum = arr[0];
```
DalinaLeila: if your sum would start at 0, you can delete the if statement on line 31

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,435 @@
...
+  "subset",
+  "trouble",
+  "starting",
+  "matter",
+  "eating",
+  "truth",
+  "disobedience"
 ];
+let doesWordExist = (arr, word) => {
+  return arr.includes(word);
```
DalinaLeila: 💯

PR #1357, opened 2020-01-07
[BER - FT] Nora Nicklis

Norizinus:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,610 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2) {
+    return number1;
+  } else if (number2 > number1) {
+    return number2;
+  }
+  return number1;
```
itstheandre: Hey Nora. Here you could have used the Math.max(number1, number2) -> it automatically returns the highest number between them.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,610 @@
...
+      longestWord.length > words[i].length ||
+      longestWord.length === words[i].length
+    ) {
+      continue;
+    } else {
+      longestWord = words[i];
+    }
+  }
+  return longestWord;
+}
```
itstheandre: This option makes a lot of sense but you didn't have to the if/ else statement.
The for loop is going to run until the end and if you dont give it nothing to do it will just ignore.
Which means that you could have checked just one thing.

```
    if (longestWord.length < words[i].length) {
        longestWord = words[i]
    }
}
```
You didnt have to do the continue

You s

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,610 @@
...
+  let sum = 0;
+  let average = null;
+  for (let i = 0; i < numbers.length; i++) {
+    sum += numbers[i];
+  }
+  if (sum !== 0) {
+    average = sum / numbers.length;
+  }
+  return average;
+}
```
itstheandre: In this function you are basically repeating the previous function you could have replaces the for loop with a:
```
sum = sumArray(numbers)
```

Besides that all good

PR #1358, opened 2020-01-07
[BER - FT] Laurel Wright 

lwrocks:
Done...thanks for all of your feedback!


starter-code/src/functions-and-arrays.js
```diff
@@ -1,15 +1,57 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b) {
+  if( a > b){
+    return a; 
+  } else {
+    return b;
+  }
```
itstheandre: Here you could have also used Math.max to get to the biggest number

PR #1361, opened 2020-01-07
[BER-FT] Jason

jfrank325:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,30 +1,56 @@
 // Iteration #1: Find the maximum
+const maxOfTwoNumbers = (firstNum, secondNum) => (firstNum > secondNum ? firstNum : secondNum);
```
itstheandre: interesting use of the reduce and of the ternary. You could have used Math.max(firstNum, secondNum) and that would have given you the response right away

starter-code/src/functions-and-arrays.js
```diff
@@ -64,9 +87,21 @@ const wordsCount = [
...
+  'matter',
 ];
 
+function howManyTimes(words, word) {
+  let newArr = [];
+  let count = 0;
+  for (let i = 0; i < words.length; i++) {
+    newArr.push(words[i].includes(word));
+    if (newArr[i] == true) {
+      count++;
```
itstheandre: Interesting approach. You could have directly checked if its true to then add it to the newArr. And then at the end just return newArr.leength

PR #1363, opened 2020-01-07
[BER-FT] Karolina 

karo-cz:
without the bonus section, work in progress

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,93 +1,488 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
+
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arrOfWords) {
+  let longestWord = [];
+  let counter = 0;
+
+  if (arrOfWords.length === 0) {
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,488 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
```
itstheandre: Here you could have also tried using the return Math.max(num1, num2).
This way you didnt need to check if one is bigger than the other and etc

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,488 @@
...
+  let longestWord = [];
+  let counter = 0;
+
+  if (arrOfWords.length === 0) {
+    return null;
+  } else {
+    for (let i = 0; i < arrOfWords.length; i++) {
+      if (arrOfWords[i].length > counter) {
+        counter = arrOfWords[i].length;
+        longestWord.unshift(arrOfWords[i]);
```
itstheandre: I believe that your solution is a little bit too complicated for this exercise actually. You could always be checking directly the size of the word and, instead of doing an array and unshifting, you could just reassign it.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,488 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arrOfNumbers) {
+  if (arrOfNumbers.length === 0) {
+    return null;
+  } else {
+    let average = sumArray(arrOfNumbers) / arrOfNumbers.length;
+    return average;
```
itstheandre: Perfect. you could also do it in one line -> return sumArray(arrOfNumbers)/arrOfNumbers

PR #1365, opened 2020-01-07
[BER-FT-012020]- Victoria B.

VictoriaBlat:
There is some code commented for the tasks 5&7, otherwise jasmine wouldn't be running correctly. I was wondering if I was at least on the right track ...
Thanks!

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,93 +1,459 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 >= number2) {
+    return number1;
+  } else {
+    return number2;
+  }
+}
+
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(words) {
+  var longestWord = "";
+  if (words.length === 0) {
+    return null;
+  }
+  for (var i = 0; i < words.length; i++) {
+    if (words[i].length > longestWord.length) {
+      longestWord = words[i];
+    }
+  }
+  return longestWord;
+}
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(numbers) {
+  var summed = 0;
+  for (var i = 0; i < numbers.length; i++) {
+    summed += numbers[i];
+  }
+  return summed;
+}
 // Iteration #4: Calculate the average
+
 // Level 1: Array of numbers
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,459 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 >= number2) {
+    return number1;
+  } else {
+    return number2;
+  }
+}
+
```
fonsogms: This correctly done!
Another way to do it which will be cleaner is by  
returning Math.max(a,b) which returns the largest number

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,459 @@
...
+  if (words.length === 0) {
+    return null;
+  }
+  for (var i = 0; i < words.length; i++) {
+    if (words[i].length > longestWord.length) {
+      longestWord = words[i];
+    }
+  }
+  return longestWord;
+}
```
fonsogms: Nicely done!
A small comment in case you wanna now you can also do another condition for empty arrays which would be ```if(!words.length){
      return null
}``` 
Is basically the same as what you wrote but shorter :)

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,459 @@
...
+    return null;
+  }
+  for (var i = 0; i < numbersAvg.length; i++) {
+    summed += numbersAvg[i];
+  }
+  var av = summed / numbersAvg.length;
+  return av;
+}
+
 // Level 2: Array of strings
```
fonsogms: Well done!
Small tip.
I would avoid using var as is getting obsolete and it can bring to many problems in the future when coding in a large scale.
Use` let` for changing variables and `const` for fixed variables instead.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,459 @@
...
+  var summ = 0;
+  if (wordsArr.length === 0) {
+    return null;
+  }
+  for (var i = 0; i < wordsArr.length; i++) {
+    summ += wordsArr[i].length;
+  }
+  var averageString = summ / wordsArr.length;
+  return averageString;
+}
```
fonsogms: Also good!
If you also want to loop through the elements of an array you can also use `for.Each()` in this case for example.
In this case will make your code cleaner and it will be shorter.
As an example `worsdArr.forEach(elem=>{
summ+=elem.length
})`

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,459 @@
...
+  "communion",
+  "simple",
+  "bring"
 ];
 
+// function uniquifyArray(wordsUnique) {
+//   for (var i = 0; i < wordsUnique.length; i++) {
+//     let check = [];
+//     if (wordsUnique.indexOf(wordsUnique[i]) > -1) {
+//       check = uniquifyArray.splice(wordsUnique[i], 1);
```
fonsogms: You were trying to splice and empty array which is uniquifyArray. Maybe you wanted to splice wordsUnique?

PR #1367, opened 2020-01-07
FTWD Berlin Sanka Update functions-and-arrays.js

san4eto:
FTWD Berlin Sanka

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,93 +1,490 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  }
+  return b;
+}
+
+//let number = "";
+//for (let i = 0; i < arr.length; i++) {
+//  if ($number < $arr[i]) {
+///  maximumNumber = arr[i];
+///}
+//}
+//return maximumNumber;
 
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  let word = "";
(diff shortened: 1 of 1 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,490 @@
...
+//function howManyTimes (parameter,toSearch){
+//  callback
+//}
+//orEach()
+
+let howManyTimes = 0;
+
+array.forEach(function(word) {
+  console.log(word);
+  howManyTimes++;
```
fonsogms: small syntax typo is for.Each() not forEach :)

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,490 @@
...
+  }
+  for (let i = 0; i === placeholder.length; i++) {
+    //const index;
+    let index = placeholder.indexOf(i);
+    if (placeholder.indexOf(i) > -1) {
+      placeholder.splice(index, 1);
+      uniquifyArray(placeholder); //should return  uniquified array but redundand
+    }
+
+    uniquifyArray(console.log(placeholder)); //should return  uniquified array
```
fonsogms: Another small comment here, you are passing a console.log as a parameter in your fuction uniquify, when what you actually want is to pass is a the array you are about to uniquify.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,490 @@
...
+  let uniqueWord = 0;
+  if (placeholder.length === 0) {
+    return [];
+  }
+  for (let i = 0; i === placeholder.length; i++) {
+    //const index;
+    let index = placeholder.indexOf(i);
+    if (placeholder.indexOf(i) > -1) {
+      placeholder.splice(index, 1);
+      uniquifyArray(placeholder); //should return  uniquified array but redundand
```
fonsogms: here you are calling the function you created, which is cool to use for recursive problems, but here what is actually happenning is that your code is starting the function again. In this if statement you are not actually returning anything if you wanted to return s
omething here, you should `return placeholder.splice(index,1)`, if that is what you wanted to do here :)

PR #1377, opened 2020-01-08
FT-WD Berlin - Jan/2020 - Ciwan

ciwan-aram:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,432 @@
...
+  "trouble",
+  "starting",
+  "matter",
+  "eating",
+  "truth",
+  "disobedience"
 ];
-
+function doesWordExist([wordsFind, wordFinder]) {
+  let wordFinder = "";
```
fonsogms: try not to create a let variable called the same as the parameter inside the function, if you do it, it will give you an error.

fonsogms: what I mean by this, if you have `function (banana){
let banana}`
you will have an error.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,432 @@
...
+  "subset",
+  "trouble",
+  "starting",
+  "matter",
+  "eating",
+  "matter",
+  "truth",
+  "disobedience",
+  "matter"
 ];
```
fonsogms: All good in general. try and keep up the good work

PR #1383, opened 2020-01-08
[BER-FT] Sebastian Mosneagu

MngSebastian:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,465 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  } else {
+    return b;
+  }
+}
```
fonsogms: all good in the hood

PR #1389, opened 2020-01-09
[PAR-WDFT]-Baptiste WARGNY - ongoing

baptistewar:


Jepassaispar:
Good job, jasmine is all green :) just few things about some of your functions : 

```
  if (words.length === 0) {
    return null;
  }
  let word = "";
    let letters = 0;
    for (let j = 0; j < words[i].length; j++) {
      letters += 1;
      if (letters > word.length) {
        word = words[i];
      }
    }
  }
  return word;
}
```

Usually for loop inside of other for loop ares used when you have a grid like the one in the bonus, except that, you would prefer to do that differently, it works either way but in very big projects with 100.000 length array it could take more processing power. Here how I would have done it : 

```
  if (words.length === 0) {
    return null;
  }
  let word = "";
    if (word.length < words[i].length) {
      word = words[i];
    }
  }
  return word;
}
```

Also for you averageWordLength you don't need a counter, you can just use the words.length : 
```
function averageWordLength(words) {
  let average = 0;
  if (words.length === 0) {
    return null;
  }
    average += words[i].length;
  }
  return average / words.length;
}
```

Anyway everything works so all good :)

PR #1391, opened 2020-01-09
[PAR-WDFT]-Franck TREBOIT - functions and arrays - ongoing

franck-treboit:
Terminé mais a la recherchje de meilleures solutions


clarabesset:
Very good job on this lab!!

One small thing, writing multiple "if" statements one after the other works most of the time (because your conditions here are mutually exclusives), but "else if" is preferable: using it prevents the program from entering into the next condition if the one before is fulfilled. 

And just so you know (your code works great, this is just simply a suggestion) - (arr.length === 0) is the same as writing (!arr.length) because 0 (length of this array) is a falsy value. Therefore you could write a shorter boolean condition :)

PR #1397, opened 2020-01-09
[PAR-WDFT] - Anthony B. - ongoing

abrebion:
Trying with array methods as much as possible. Will try bonus later.


abrebion:
The second commit includes the bonus and also use standard for loop for all exercice.
Short versions using array methods are commented out.

Jepassaispar:
Great job, very nicely done both with for loop and higher order functions, nothing to add :)

PR #1398, opened 2020-01-09
[PAR-FT-WD] - Coline VELARD ongoing

colinevel:
done


clarabesset:
Great job! Just so you know (your code works great, this is just simply a suggestion) - (arr.length === 0) is the same as writing (!arr.length) because 0 (length of this array) is a falsy value. Therefore you could write a shorter boolean condition :)

PR #1399, opened 2020-01-09
[PAR-WDFT-0120] - Yannick Bourenane ongoing 

yannick-bourenane:
Didn't try the diagonal check in the bonus yet.


clarabesset:
Very nice job!! I like how you wrote doesWordExist() setting is to false by default with only one condition if the word is included - it makes it much shorter !

PR #1400, opened 2020-01-09
[PAR-WDFT]-MORSLI ongoing

PaulineLalou:
Je n'ai pas réussi à valider tous mes tests. Je n'ai pas trouvé de méthode plus élégante pour le dernier exercice mais je sais que c'est possible. J'attends votre retour merci à vous ! :)

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,93 +1,486 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(number1, number2) {
+  return Math.max(number1, number2);
+}
+
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


clarabesset:
For doesWordExist, you might want to use "includes" as such - `function doesWordExist(array, word) {
  if (array.length == 0) {
    return false;
  } else if (array.includes(word)) return true;
  else return false;
}`

clarabesset:
For the last iteration howManyTimes, I'll give you a hint because I think you can find the solution: since you are counting how many times a word appears, you'll probably be using a loop to go through the array.  Then using a variable you increment each time you're going inside the loop, you can get the count of how many times a word is present in the array :) - Come to me if you need any help!

PR #1403, opened 2020-01-10
[PAR-WDFT] - Kathleen D. ongoing

kathd:
Would you please give the solution for the Bonus exercise? :) Thank you!


clarabesset:
We will give the labs solutions (which are only proposals, there is not only one right way to do things) one week after each lab, to give everyone a chance to finish them at their own pace :) 

But, really good job on this lab! So you know for future reference - writing (!wordArr.length) is the same as writing (wordArr.length === 0) since 0 (length of this array) is a falsy value. Therefore you could write a shorter boolean condition :) This is totally optional, but boolean condition are nice!

kathd:
Cool! Thanks for the tip! :)

PR #1405, opened 2020-01-10
[PAR-FT-0120] - María.MUNIZAGA - ongoing

mmunizaga:


Jepassaispar:
Great job María ! Jasmin is indeed very happy !

Everything is working fine but I reviewed few details for you to improve : 

in your findLongestWord function you create a longestWord array to add words inside and return the first value of the array. You could actually accomplish the same things without the array : 
```
function findLongestWord(wordArr){
  if (wordArr.length === 0) {
    return null;
  }
  var longestWord = wordArr[0];
  for (let i = 0; i < wordArr.length; i++) {
    if (wordArr[i].length > longestWord.length) {
      longestWord = wordArr[i];
    }
  }
  return longestWord;
}
```

All your function's params  (ex: `averageWordLength (wordsArr1)` and `function uniquifyArray (wordsArr2)`) can have "wordsArr" or "toto" as parameter, the paramater can be named absolutely anything, the important part is what you put as parameter when you call the function.
 
For your bonus, you have the logic, you are just missing the horizontal calculation.

Anyway, everything works fine, you got it ;) if you have any question feel free to check with me !

PR #1410, opened 2020-01-12
[AMS-PT-012020] - Jonas B.

jonasbarre:
…would love to see a solution for the bonus in class

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,15 +1,82 @@
 // Iteration #1: Find the maximum
 
+const maxOfTwoNumbers = function (num1, num2) {
+  // if num1 is bigger than num2, return num1
+  if (num1 > num2) {
+    return num1
+    // if num2 is bigger than num1, return num2
+  } else if (num2 > 1) {
+    return num2
+  } else {
+    // if they are equal, return error message
+    console.log('Your numbers are equal.')
+  }
+}
+
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+let findLongestWord = function (array) {
+  var longest = "";
+  // if there is only one item in the array, return that item
+  if (array.length === 1) {
+    return array[0];
+    // if the array is empty or undefined, return undefined
+  } else if (array.length < 1 || array == undefined) {
+    return undefined;
+  } else {
+    // loop through each item in the array and check for the length of each item. Each time the item's length is longer than the one stored in the 'longest' variable, overwrite it.
+    for (var i = 0; i < array.length; i++) {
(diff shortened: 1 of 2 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,83 @@
 // Iteration #1: Find the maximum
 
+const maxOfTwoNumbers = function (num1, num2) {
+  // if num1 is bigger than num2, return num1
+  if (num1 > num2) {
+    return num1
+    // if num2 is bigger than num1, return num2
+  } else if (num2 > 1) {
```
LloydC: I think you meant to write num1but made a typo instead with 1

starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,83 @@
...
+}
+
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+let findLongestWord = function (array) {
+  var longest = "";
+  // if there is only one item in the array, return that item
+  if (array.length === 1) {
```
LloydC: I feel this 'if' statement is unnecessary. Even with a single element in your array, your 'else' statement would be handle it so you don't really need this one.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,83 @@
...
+findLongestWord(words);
+
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumArray(array) {
+  // !array.length is the same as writing array.length === 0
+  let sum = 0
+  if (!array.length) {
```
LloydC: Same as above, you don't need this 'if' statement. If your array is empty then the return sum will be returned with it's original value which is 0.

starter-code/src/functions-and-arrays.js
```diff
@@ -52,6 +153,27 @@ const wordsFind = [
   'disobedience'
 ];
 
+function doesWordExist(array, searchWord) {
+  // if array is empty, return false
+  if (!array.length) {
+    return false;
+    // if array only contains 1 item and this item is the search word, return true
+  } else if (array.length === 1 && array[0] === searchWord) {
```
LloydC: Again you don't need this 'else if' statement. Your 'else' statement is enough

starter-code/src/functions-and-arrays.js
```diff
@@ -67,6 +189,23 @@ const wordsCount = [
   'matter'
 ];
 
+function howManyTimes(array, searchWord) {
+  let counter = 0
+  // if array is empty, return 0
+  if (!array.length) {
```
LloydC: Again, you don't need the 'if' statement here.

PR #1411, opened 2020-01-12
[PAR-FT-012020] - Youval BUGAJSKI ongoing

YuvalBu:


TatijanaRajcic:
Hey Youval! Your work is nice :) 
Here are some piece of advice:
- For your function maxOfTwoNumbers, you could write it like this:
```
  if (a > b) {
    return a
  } 
 return b;
}
```
You don't need the else here. If (a > b) it will return a, otherwise the function will return b.

To make your code even shorter, you could use a ternary operator : 
```
  return a > b ? a : b;
}
```

- For your function findLongestWord, you don't need the special (words.length === 1), your loop covers it
- For your function sumArray: same remark :) do not need the special case. You could also use reduce to make it even shorted:
```
function sumArray(numArray) {
  return numArray.reduce((total, num) => { return total + num; }, 0);
}
```

- in your averageNumbers function, you could call your function sum, so that you don't have to repeat code :)
- In your averageWordLength function, to make it even shorted you could use map and then averageNumbers:

```
function averageWordLength(wordsArray) {
  if ( wordsArray.length === 0 ) { return null; }
  let wordsLengthArray = wordsArray.map( (word, index) => { return word.length } );
  return averageNumbers(wordsLengthArray);
};
```

- For your function doesWordExist, you could make it simplier:
```
  return wordsArray.includes(word);
}
```

- For your function howManyTimes, you could write your conditions shorter:
array.includes(word) === false is the same as !array.includes(word)
as both of your two first conditions return 0, you could use the || operator:
if (array.length === 0 || array.includes(word) ) { ... }

Good work Youval :) Those are only improvements to make your code shorter!
Don't hesitate to tell me if you try the bonus :)

PR #1412, opened 2020-01-12
Done (without bonus)

fhoman:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,130 @@
...
+
+
+
+
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot', 'christmas'];
+
+
+function findLongestWord(wordArr) {
```
lotofcaffeine: Hey, Your code is pretty neat! One alternative to creating function is arrow function, in the future when you start to create functions inside Express or Classes, you will see how this will help you.
 I really like the tutorial from scrimba:  https://scrimba.com/p/p4Mrt9/c7vd3cd

fhoman: Thanks, i will look into it!

PR #1413, opened 2020-01-12
done [PAR-WDFT] Andrew NB - ongoing tati

andrewnunezbibby:


TatijanaRajcic:
Hey Andrew! Nice work ! :) It's good practice to use the ternary operator so well done for question 1!
- For your sum, you could have used reduce to make it even shorter:
```
function sumArray(numArray) {
  return numArray.reduce((total, num) => { return total + num; }, 0);
}
```
- For your averageNumbers function, instead of writing the calculation for the sum again, you could use your function sumArray inside :)
- Nice use of Set for uniquifyArray, first time I see this type of solution :)
- In your function doesWordExist you do not need 2 conditions, this is enough;
```
doesWordExist = (arr, word) => {
  if (arr.indexOf(word) !== -1) return true;
  else return false;
}
```
- If you try the bonus, don't hesitate to send us a message! You're almost there :) 
Nice job!

PR #1415, opened 2020-01-12
[PAR-FT-012020] Wendy functions-and-arrays - ongoing

Wendev13:


Jepassaispar:
I really like that your functions are long at first and very short at the end, usually it is the other way around ;)
I especially liked your howManyTimes function, it looks incredible 🗡 
Great job Wendy :D

PR #1421, opened 2020-01-13
Done

danielhoogerwerf:
Did not understand the bonus question, so I had to skip it.

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -9,22 +9,48 @@ const findLongestWord = wordArray => {
 	wordLength = "";
 	wordArray.forEach(function(word) {
 		wordLength += word.length;
-	});
+  });
+  // the ... in Math.max allows an array (or in this case a string) as input. It's called Spread I believe...
 	result = wordLength.indexOf(Math.max(...wordLength));
 	return result >= 0 ? wordArray[result] : null;
 };
 
-console.log(findLongestWord(words));
+//console.log(findLongestWord(words));
 
-// findLongestWord(words);
+findLongestWord(words);
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+const sumArray = arg1 => {
+  calculateSum = 0;
+  arg1.forEach(function(numArr) {
+	  calculateSum += numArr;
+  });
+  return calculateSum;
+}
+
+console.log(sumArray(numbers));
+
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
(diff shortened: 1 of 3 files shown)


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,57 @@
...
+	wordArray.forEach(function(word) {
+		wordLength += word.length;
+  });
+  // the ... in Math.max allows an array (or in this case a string) as input. It's called Spread I believe...
+	result = wordLength.indexOf(Math.max(...wordLength));
+	return result >= 0 ? wordArray[result] : null;
+};
+
+//console.log(findLongestWord(words));
+
```
LloydC: It works but if you have a word that's more than 9 characters, it would break. I'd recommend using an approach where you have an empty string longestWord, check each word of the array and if it's length is bigger than the length of longestWord then replace longestWord with it.

starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,57 @@
...
+const averageNumbers = arg1 => {
+	let amountOfItems = 0;
+	let calculateSum = 0;
+	for (let i = 0; i < arg1.length; i++) {
+		amountOfItems = i + 1;
+	}
+	arg1.forEach(function(numArr) {
+		calculateSum += numArr;
+	});
+	return amountOfItems > 0 ? calculateSum / amountOfItems : null;
```
LloydC: You can also use 'arg1.length', if it's less than one then return null and otherwise return calculateSum/arg1.length

starter-code/src/functions-and-arrays.js
```diff
@@ -25,6 +66,18 @@ const wordsArr = [
...
 ];
 
+const averageWordLength = wordArray => {
+	let wordLength = 0;
+	let amountOfItems = 0;
+	wordArray.forEach(function(word) {
+		amountOfItems += 1;
+		wordLength += word.length;
+	});
+	return wordLength > 0 ? wordLength / amountOfItems : null;
```
LloydC: Same as above, you can use 'wordArray.length' here

starter-code/src/functions-and-arrays.js
```diff
@@ -52,6 +115,17 @@ const wordsFind = [
   'disobedience'
 ];
 
+const doesWordExist = (arg1, arg2) => {
+	let result = false;
+	for (let i = 0; i < arg1.length; i++) {
+		//console.log(`word: ${arg1[i]}, keyword ${arg2}`);
+		if (arg1[i] === arg2) { result = true; break; } else { result = false; }
+	}
+	return result;
```
LloydC: You can also do this: return arg1.includes(arg2)//true or false

PR #1430, opened 2020-01-15
done

MichielSnoek:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,64 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 > num2) {
+    return num1
+  } else {
+    return num2
+  }
```
LloydC: You can also do it with the ternary operator shorthand notation like return num1 > num2 ? num1 : num2

starter-code/src/functions-and-arrays.js
```diff
@@ -52,6 +125,18 @@ const wordsFind = [
   'disobedience'
 ];
 
+function doesWordExist(arr, word) {
+  if(arr.includes(word)) {
+    return true;
+  } else {
+    return false;
+  }
```
LloydC: You can use a short expression like return arr.includes(word), it should return true or false already by itself

PR #1461, opened 2020-02-01
BER - WDPT - Pia

pgmdf:
done iteration 1 and 2


starter-code/src/functions-and-arrays.js
```diff
@@ -52,7 +156,27 @@ const wordsFind = [
...
+      continue;
+    }
+  }
+  return trueOrFalse
+}
+
+doesWordExist(wordsFind, "starting")
+
 // Iteration #7: Count repetition
+
```
numbing: awesome job, i really like your solutions  if i wanna be picky i could say its better to your functions has "return" instead of console.log() then when you wanna check the result you could console.log the function itself

PR #1463, opened 2020-02-01
ber wdpt camille and marie

mariehardy:


starter-code/src/functions-and-arrays.js
```diff
@@ -67,6 +191,22 @@ const wordsCount = [
...
+  }
+  console.log(result7)
+}
+
+howManyTimes(wordsCount, 'matter')
+
+
 // Iteration #8: Bonus
 
 const matrix = [
```
numbing: nice job, its really good that you finished the bonus as well... well done

PR #1464, opened 2020-02-01
BER-WEDPT-01-2020

dkoushha:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,93 +1,446 @@
...
+function howManyTimes(someArr, word) {
+  let count = 0;
+  for (let i = 0; i < someArr.length; i++) {
+    if (someArr[i] === word) {
+      count += 1;
+    }
+  }
+  return count;
+}
+
```
numbing: woow really good, i really like  your solutions

PR #1468, opened 2020-02-01
BERLIN - WDPT - Edison

edisonabdiel:


starter-code/src/functions-and-arrays.js
```diff
@@ -1,16 +1,44 @@
...
+var sum = numbers.reduce(function(a, b) { return a + b; }, 0);
+
+console.log(sum)
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function avrg(arr) {
+  const sum = arr.reduce((a, b) => a + b, 0);
```
numbing: as i said before please dont use reduce for now, first we need to learn basic js then we gonna cover everything.
it took so much of your time and you are half done

starter-code/src/functions-and-arrays.js
```diff
@@ -22,9 +50,21 @@ const wordsArr = [
...
-  'palace'
+  'palace',
 ];
 
+function arrAvg (wordsArr){
+  let sum = 0
+  for (let i = 0; i < wordsArr.length; i++){
+    sum += wordsArr[i].length
+  }
+  console.log(sum / wordsArr.length )
```
numbing: you shouldn't console.log() here this function should " return "

PR #1483, opened 2020-02-15
 LIS-PT-022020 - Stefano S. & André R.

oddouu:
@ironhack-LIS-PTWD
We think we managed to finish all the first 8 iterations. We weren't really sure about the last one though, but the test appears to be working.

P.S. It seems like there is a 8.1 bonus in the lab description on Git Hub, however it doesn't show up in the Jasmin Specs. Should we do it anyway?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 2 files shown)


ironhack-LIS-PTWD:
Hi guys,
if you like you ca do the last iteration even though there is no test but since its just a Bonus its up to you :) 
Well done on finishing all the iterations!
Paula

PR #1488, opened 2020-02-16
LIS - PTWD - FEB - Julia Carvalho

juliamendesc:
@ironhack-LIS-PTWD

Hi Girls! Could not really finish the challenge, my brain has already melted trying to find so many solutions with it.

Here is what troubled me the most:
1) Array of strings - Bonus: // Jasmine wants 5.11 (don't know which mixed array it uses), but Ironhack git says 5.7, which makes this CORRECT!!! But I would be curious to understand where is Jasmine seeking the variables that would make 5.11.

2) uniquifyArray - tried with switch, with for loop, but it is comming out null. // JASMINE SOLUTION WITH DIFFERENT WORDS - which I also don't know where it is the variable they use.

3) Find elements - function apparently did not work

Thank you!

### DIFF
--- a/starter-code/src/functions-and-arrays.js
+++ b/starter-code/src/functions-and-arrays.js
@@ -1,18 +1,143 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b) {
+  if (a > b) return a
+  if (a < b) return b
+  if (a === b) return a,b;
+}
+maxOfTwoNumbers(1,2)
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-// Iteration #3: Calculate the sum
+function findLongestWord(word) {
+  let longest = "";
+  if (word <= 0) {
+    return null;
+  }
 
+//   //ALTERNATIVE
+// if (!Array.length) {
+//   return null;
+// }
+
+  // } else if (words = "") {
+  //   return words;
+  // }
+   for (let i = 0; i < word.length; i++) {
+    if (word[i].length > longest.length) {
+      longest = word[i];
+    }
+  } return longest
+  }
+
+// Iteration #3: Calculate the sum
+function sumNumbers(numbers) {
+  let sum = 0;
+    if (numbers.length == 0){
+    return 0;
+  } 
+  for (let i=0; i <numbers.length; i++) {
+    if (numbers.length == 1) {
+    return sum += numbers[i];
+  } else if (numbers[i] === 0) {
+    return 0;
+  }
+  for (let i=0; i < numbers.length; ++i) {
+    sum += numbers[i]; // add each element in an array to total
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hi Julia,

1. to finish iteration array of strings you were just missing minor thing.
when you are checking this :` if (typeof mixedArr[i] == "object" || "array")`
you just have to change it to this: ` if (typeof mixedArr[i] === "object" || typeof mixedArr[i] === "array")`

here is my full solution to this iteration: 
```
function sum(numArray) {
  let sum = 0;
  for (i = 0; i < numArray.length; i++) {
    if (typeof numArray[i] === "string") {
      sum += numArray[i].length;
    } else {
      sum += numArray[i];
    }
    if (typeof numArray[i] === "object" || typeof numArray[i] === "array") {
    }
  }
  return sum;
}
```

if you want to see exactly which arrays Jasmine uses for testing, go to the file  `/tests/functions-and-arrays.spec.js` and check the corresponding test. 
for example this is the test for iteration array of strings:
```
describe('Bonus: Calculate the average of a mixed elements array', () => {
  it('should create a function named avg', () => {
    expect(typeof avg).toBe('function');
  });

  it('should return null if receives an empty array when called', () => {
    // should it return null or zero?
    expect(avg([])).toBe(null);
  });

  it('should return the average of the array', () => {
    // false is counted as 0
    expect(avg([6, 12, 'miami', 1, 'barca', '200', 'lisboa', 8, false])).toBe(5.11);
    // true is counted as 1
    expect(avg([6, 12, 'miami', 1, 'barca', '200', 'lisboa', 8, true])).toBe(5.22);
  });
});
```

2.  Have a look at this solution and let me know if it makes sense to you. It's important to understand the use of [`indexOf()` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf)

```
function uniquifyArray(wordsArray) {
  if (!wordsArray.length) {
    return null;
  } else {
    let newArray = [];
    for (i = 0; i < wordsArray.length; i++) {
      if (newArray.indexOf(wordsArray[i]) === -1) {
        newArray.push(wordsArray[i]);
      } else {
        continue;
      }
    }
    return newArray;
  }
}
```

3. solution for  does word exist

```
  if (!array.length) {
    return null;
  } else if (array.includes(word)) {
    return true;
  } else {
    return false;
  }
}



```

hope this helps! if you have any questions for the last two solutions, we can go through that in class today. 
Cheers Paula :)

juliamendesc:
@ironhack-LIS-PTWD

Hi Paula and Lissy, 

the comments were very helpful! I am still with a few queries on the indexOf. 

I should always use the indexOf with the -1 / 0 related to it, right?

Thanks!

ironhack-LIS-PTWD:
You are welcome :) 
yes exactly. If indexOf("item") returns -1 it means that the item you are checking is not included in the array. it it returns 0 or greater, than it is the actual index of the item you are checking.

PR #1494, opened 2020-02-18
AMS - Web dev - FEB - FT

Charlesvdb:


MFn-Scarboy:
- Technical Requirements (9/10)

- Creative Code (9/10)

- Code Quality (8/10)

Iteration #3
Misnamed function caused all tests to fail

Iteration #5
Returning 'null' instead of '[]'

Iteration #6
The 'break's are unnecessary.
Returning 'null' instead of false.
Your 'for' loop only runs once (try to see why).

Couple oversights, overall pretty good, indentation could be cleaner

PR #1499, opened 2020-02-18
[Ams-WD-FT-Feb] Fábio Costa

fabioMatosCosta:
Finish the 7 iterations, but not the bonuses.
:)


mswidzinska:
- Technical Requirements (10/10)
- Creative Code (10/10)
- Code Quality (9/10)
Well done!
A small thing to improve: you can consider using better naming that 'let a =0' or  'let a = false'.  It should be more logic and meaningful.

PR #1507, opened 2020-02-19
AMS  FEB FT FEB - Yvana Wagter

Y-VW:


MFn-Scarboy:
- Technical Requirements (9/10)

- Creative Code (10/10)

- Code Quality (10/10)

Iteration #1 
The double return is redundant
else
    return num1;
    return num2;
}

Iteration #3 
Misnamed function (caused all tests to fail)

Iteration #5
Returning 'null' instead of '[]'

Iteration #6
Returning 'null' instead of 'false'

Couple oversights, aside from those, clean code, couple of very efficient solutions.

--Joel

PR #1538, opened 2020-03-24
BER-FTWD-Yosra

yosrayousfi:
I finished the lab exercises  and then added extra tests for the bonus questions


src/functions-and-arrays.js
```diff
@@ -1,73 +1,504 @@
...
+    }
+  });
+  return result;
+};
+const sum = arr => {
+  if (hasUnsuporttedDataType(arr)) {
+    throw new Error("Unsupported data type sir or ma'am");
+  }
+  let counter = 0;
+  for (let i = 0; i < arr.length; i += 1) {
```
fenrew: I would personally have put in the "hasUnsporttedDataType" checking here:
if(typeof arr[i] == "object" ) throw new Error("Unsupported data type sir or ma'am")

Makes it a bit cleaner in my opinion, however it's just a personal preference

src/functions-and-arrays.js
```diff
@@ -1,73 +1,504 @@
...
+  //   (acc, cur) => (typeof cur == "string" ? acc + cur.length : acc + cur),
+  //   0
+  // );
+};
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+const averageNumbers = arr =>
+  testIfInvalid(arr) ? null : sumNumbers(arr) / arr.length;
```
fenrew: Nice! This is so clean since you are re-using the functions made previously!

PR #1541, opened 2020-03-24
BER-FTWD-Olha

olhadanylchenko:
done except the last one

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,450 @@
 // Iteration #1: Find the maximum
-
+const maxOfTwoNumbers = (num1, num2) => {
+  if (num1 > num2) return num1;
+  else if (num2 > num1) return num2;
+  else return num1;
+};
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+const findLongestWord = words => {
+  let longestWord = "";
+  if (words.length === 0) return null;
+  else {
+    for (let i = 0; i < words.length; i++) {
+      if (words[i].length > longestWord.length) {
+        longestWord = words[i];
+      }
+    }
+    return longestWord;
+  }
+};
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+const sumNumbers = array => array.reduce((acc, curval) => acc + curval, 0);
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,450 @@
 // Iteration #1: Find the maximum
-
+const maxOfTwoNumbers = (num1, num2) => {
+  if (num1 > num2) return num1;
```
fenrew: You can instead write:
if(num1 >= num2) return num1

This way you will not need your: 
else return num1

src/functions-and-arrays.js
```diff
@@ -1,73 +1,450 @@
...
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+const findLongestWord = words => {
+  let longestWord = "";
+  if (words.length === 0) return null;
```
fenrew: Since you are returning:
return null

You wont need the else statement since if the condition is true you return out of the function anyways

src/functions-and-arrays.js
```diff
@@ -1,73 +1,450 @@
...
+    }
+    return longestWord;
+  }
+};
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+const sumNumbers = array => array.reduce((acc, curval) => acc + curval, 0);
```
fenrew: Nice!

Also, if you dont set any default values on reduce it will automatically start at 0, so there is no need to set the default value in this particular case

src/functions-and-arrays.js
```diff
@@ -1,73 +1,450 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+const sum = array =>
+  array.reduce((acc, curval) => {
+    if (typeof curval === "number" || typeof curval === "boolean") {
```
fenrew: Niiice!

src/functions-and-arrays.js
```diff
@@ -1,73 +1,450 @@
...
+      return acc + curval;
+    } else if (typeof curval === "string") {
+      return acc + curval.length;
+    } else if (typeof curval === "object") {
```
fenrew: You can technically just have an else statement instead of else if here

src/functions-and-arrays.js
```diff
@@ -1,73 +1,450 @@
...
+  "poison",
+  "communion",
+  "simple",
+  "bring"
 ];
 
+const uniquifyArray = array => {
+  if (array.length === 0) return null;
+  else
+    return array.reduce((acc, curval) => {
```
fenrew: This right here is just beautiful!

olhadanylchenko:
Thank you @fenrew 😊  https://frontendmasters.com/courses/functional-js-fundamentals/  -- this course helped me sooo much to understand how reduce works under the hood. 
Still get confused with explicit/implicit returns -- I will review and adjust it based on your comments. :)

PR #1544, opened 2020-03-24
BER-FTWD-[Karl]

karlkristopher:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,119 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers (a, b) {
+  if (a > b) {
+    return a;
+  }
+  return b
```
fenrew: Nice!

src/functions-and-arrays.js
```diff
@@ -1,19 +1,119 @@
...
+
+
+function findLongestWord(arr) {
+  
+
+  if (typeof arr == 'undefined' || arr.length < 1){
+    return null;
+  } 
+
+  let longestWord = 'a';
```
fenrew: I would rather have this to be: let longestWord = ''
To prevent any potential bugs

src/functions-and-arrays.js
```diff
@@ -1,19 +1,119 @@
...
+
+  if (typeof arr == 'undefined' || arr.length < 1){
+    return 0;
+  } 
+
+  let sum = 0;
+
+  for (let i = 0; i < arr.length; i++) {
+    
+    if (typeof arr[i] === 'number') {
```
fenrew: you can actually have an or satetment here: 
typeof arr[i] === 'number' || arr[i] == true

because: 5 + true === 6

src/functions-and-arrays.js
```diff
@@ -1,19 +1,119 @@
...
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arr) {
+
+  if (typeof arr == 'undefined' || arr.length < 1){
+    return null;
+  } 
+
+  return sumNumbers(arr) / arr.length;
```
fenrew: Nice re-use of your functions!:)

PR #1545, opened 2020-03-24
BER-FTWD-Robert

robertbozsik:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,94 @@
...
+  /*
+  if (a > b) {
+    return a;
+  }
+  else {
+    return b;
+  }
+  */
+} 
+console.log("Iteration #1: Find the maximum:", maxOfTwoNumbers(2, 4));
```
Jeff-Moraes: well done

src/functions-and-arrays.js
```diff
@@ -1,19 +1,94 @@
...
 
+// let wordsLength = [7, 7, 7, 9];
+// console.log(wordsLength.indexOf(9)); // 3
+
+function findLongestWord(anyArray) {
+  lengthWords = [];
+  for (let i = 0; i < anyArray.length; i++) {
+    lengthWords.push(anyArray[i].length);
+  }
+
```
Jeff-Moraes: here you could just add
`if (words.length === 0) return null;`
to check if it is an empty array.

src/functions-and-arrays.js
```diff
@@ -1,19 +1,94 @@
...
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(anyArray) {
+  let sumAll = 0;
+  for (let i = 0; i < anyArray.length; i++) {
+    sumAll += anyArray[i];
+  }
+  return sumAll;
+}
```
Jeff-Moraes: great job here

src/functions-and-arrays.js
```diff
@@ -1,19 +1,94 @@
...
+//console.log(sumNumbers(numbersAvg));
+//console.log(numbersAvg.length);
+console.log("Iteration #4: Calculate the average / Level 1:", sumNumbers(numbersAvg) / numbersAvg.length);
+
+
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
+function averageWordLength(anyArray) {
+  lengthOfWordArray = [];
```
Jeff-Moraes: same as before, just check if the array is empty.
`if (words.length === 0) return null;`

src/functions-and-arrays.js
```diff
@@ -29,9 +104,52 @@ const wordsUnique = [
...
+*/
+
+
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
+// console.log(wordsFind.includes("apple"));
+
+function doesWordExist(anyArray, word) {
+  if (anyArray.includes(word)) {
```
Jeff-Moraes: `if (words.length === 0) return null;` ;)

src/functions-and-arrays.js
```diff
@@ -47,6 +165,19 @@ const wordsCount = [
...
 
+function howManyTimes(anyArray, word) {
+  wordCount = 0;
+  for (let i = 0; i < anyArray.length; i++) {
+    if (anyArray[i] === word) {
+      wordCount += 1;
+    }
+  }
+  return wordCount;
+}
```
Jeff-Moraes: Very well done!!

PR #1546, opened 2020-03-24
WDFT-BER-Yung-Ting

yung-ting:
Done with a few bugs.
Most of the iterations are done, but they encounter the same issue of not returning null when called with empty array. Any tips? Thanks a lot!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,20 +1,74 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers (a, b) {
+  if (a > b) {
+      return a;
+  }
+  else if (a < b) {
+      return b;
+  }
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,20 +1,74 @@
...
 // Iteration #2: Find longest word
+function findLongestWord (array) {
+  let length = 0;
+  let longestWord;
+  for (let i = 0; i < array.length; i++) {
+      if (array[i].length > length) {
+          length = array[i].length;
+          longestWord = array[i]; 
+      }
+      return longestWord;
```
Jeff-Moraes: here you should put the `return longestWord;` out of the "for loop", so it should be between the line 24 and 25.

PR #1549, opened 2020-03-24
BER-FTWD-Chantel

chantel-taki:


src/functions-and-arrays.js
```diff
@@ -1,36 +1,113 @@
...
+  if (array.length === 0) {
+    return null;
+  }
+  let word = "";
+  for(let i=0; i < array.length; i++) {
+    if (word.length < array[i].length) {
+      word = array[i];    
+    } if (array.length <= 0) {
+      callback(err , []);
+    }
```
Jeff-Moraes: these lines are not necessary, as you have already checked if the array is empty on line 14.

src/functions-and-arrays.js
```diff
@@ -1,36 +1,113 @@
...
+  "playground",
+  "poison",
+  "communion",
+  "simple",
+  "bring"
 ];
+function uniquifyArray(wordArr) {
+  if (wordArr.length === 0) return null; {
+    
+  }
```
Jeff-Moraes: you don't need theses curly brackets. ;)

src/functions-and-arrays.js
```diff
@@ -1,36 +1,113 @@
...
+  return newArray;
+}
+console.log(uniquifyArray(wordsUnique));
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
+function doesWordExist (array, word) {
+  if (array.length === 0) return null; {
+
+  }
```
Jeff-Moraes: same here.. you don't need theses curly brackets. ;)

PR #1551, opened 2020-03-24
done

paulamoellendorf:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,72 @@
...
+  return null;
+}
+  let longestWord='';
+  for(i=0;i<arr.length;i++){
+    if (longestWord.length < arr[i].length){
+    longestWord=arr[i];
+    } 
+  }
+  return longestWord;
+}
```
Jeff-Moraes: well done

src/functions-and-arrays.js
```diff
@@ -1,19 +1,72 @@
...
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+function sumNumbers(arr){
+  let sum=0;
+  for(i=0;i<arr.length;i++){
+    sum+=arr[i];
+  }
+return sum;
+}
```
Jeff-Moraes: great

src/functions-and-arrays.js
```diff
@@ -1,19 +1,72 @@
...
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
-
+function averageNumbers(arr){
+  if(arr.length == 0){
+    return null;
+  }
+let average = sumNumbers(arr)/ arr.length;
+return average;
+}
```
Jeff-Moraes: nicely done.

src/functions-and-arrays.js
```diff
@@ -29,9 +82,31 @@ const wordsUnique = [
...
+    return null;
+  }
+  for(i=0;i<arr.length; i++){
+    if(indexOf(arr[i]).length>1){
+      arr.splice(i,1);
+    }
+  }
+ return arr;
+}
+
```
Jeff-Moraes: I think the `.slice()` method is not a good approach to this case.
For this interaction, you can take a look at Jan solution:

`function uniquifyArray(ar) {
  if (ar.length === 0) return null;
  let uniqueArray = [];
  for (let el of ar) {
    if (uniqueArray.indexOf(el) === - 1) {
      uniqueArray.push(el)
    }
  }
  return uniqueArray;
}`

src/functions-and-arrays.js
```diff
@@ -29,9 +82,31 @@ const wordsUnique = [
...
+  if(arr.length ==0){
+    return null;
+  }
+for(i=0; i<arr.length; i++){
+  if(word== arr[i]){
+    return true;
+  } 
+}
+return false;
+}
```
Jeff-Moraes: nice approach!

PR #1552, opened 2020-03-24
BER DIONI almost done

dionifss:
I will keep working a bit more tommorow and repush it ;)

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -147,7 +147,7 @@ const wordsUnique = [
 function uniquifyArray(words) {
   let newArray = [];
   for (i = 0; i < words.length; i++) {
-    if (console.log(newArray.indexOf(words[i])) == -1) {
+    if (newArray.indexOf(words[i]) == -1) {
       newArray.push(words[i]);
     }
   }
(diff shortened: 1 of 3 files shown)


dionifss:
All done less the last one ;), if the push is not correct just let me know

src/functions-and-arrays.js
```diff
@@ -1,73 +1,500 @@
-// Iteration #1: Find the maximum
+function maxOfTwoNumbers(n1, n2) {
+  if (n1 > n2) {
+    return n1;
+  } else if (n2 > n1) {
+    return n2;
+  } else {
+    return n1;
```
fonsogms: You could have saved and else if like this: 
if (n1 > n2) {
    return n1;
  } else{return n2}

src/functions-and-arrays.js
```diff
@@ -1,73 +1,500 @@
...
+
+const mixedArr = [6, 12, "miami", 1, true, "barca", "200", "lisboa", 8, 10];
+
+function sum(mixedArr) {
+  let sumatory2 = 0;
+  for (i = 0; i < mixedArr.length; i++) {
+    if (typeof mixedArr[i] === "object") {
+      return "Unsupported data type sir or ma'am";
+    }
+    if (typeof mixedArr[i] === "boolean") {
```
fonsogms: no need to at this statement, as true + a number is equal to= 1+number :)

src/functions-and-arrays.js
```diff
@@ -1,73 +1,500 @@
...
-const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
+function averageWordLength(wordsArr) {
+  let sumatory4 = 0;
+  for (i = 0; i < wordsArr.length; i++) {
+    sumatory4 += wordsArr[i].length;
+  }
+  if (sumatory4 == 0) {
+    return null;
+  }
+  sumy1 = sumatory4 / wordsArr.length;
```
fonsogms: avoid creating a global variable that is being used by more than one function, and create that variable inside of the function instead

PR #1556, opened 2020-03-25
berlin-ftwd-sebastian

SebastianNagl:


src/functions-and-arrays.js
```diff
@@ -29,8 +147,30 @@ const wordsUnique = [
...
+  }
+}
+
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
+let doesWordExist = (arr, word) => {
+  if (arr.length === 0) {
+    return null;
+  } else {
+  if (arr.includes(word)) {
```
fonsogms: no need to make this else statement cause if the first condition happens the function will stop and return null :)
Example 
```
if(arr.length===0){return null;}
if(arr.includes(word)){return true}
else{return false}
```

PR #1557, opened 2020-03-25
BER-WDFT-Jony

jonystn:
Without the bonus


src/functions-and-arrays.js
```diff
@@ -1,18 +1,94 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  } else if (a < b) {
+    return b;
+  } else {
```
fonsogms: you could also do one else stament returning b and save some code :)
if (a > b) {
    return a;
  } else  {
    return b;
  }

PR #1558, opened 2020-03-25
almost done

marciafurtado:
Didn't finish and have many questions :)
I'll keep trying anyway.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -42,11 +42,12 @@ let sum = 0;
   for (let i = 0; i < arrayOfNumbers.length; i++){
     sum += arrayOfNumbers[i];
   }
-  if(arrayOfNumbers.length === 0 || sum === 0){
+  if(sum.length === 1){
+    return sum;
+  } else if (arrayOfNumbers.length === 0 || sum === 0){ // I figured it out that sum is already zero, so that's why it wasn't working 
     return 0;
-  } else if(sum.length === 1){
-    return sum; // not working
   }
+    
 } 
 console.log(sumNumbers(num23));
 
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,103 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers (num1, num2){
+  if(num1 > num2) {
+    return num1;
+  } else if (num1 < num2){
+    return num2;
+  } else {
+    return num1; 
+  }
+}
+
 // Iteration #2: Find longest word
+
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,105 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers (num1, num2){
+  if(num1 > num2) {
+    return num1;
+  } else if (num1 < num2){
+    return num2;
+  } else {
+    return num1;
```
fonsogms: You can use only one else statement and still make it work.
```
  if(num1 > num2) {
    return num1;
  } else{
    return num2;
  }
```

src/functions-and-arrays.js
```diff
@@ -1,19 +1,105 @@
...
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+function findLongestWord (arrayOfWords){
+let longestWord = '';
+
+  for( let i = 0; i < arrayOfWords.length; i++){
+    if(longestWord.length === arrayOfWords.length){
+      continue;
+    } else if(arrayOfWords[i].length > longestWord.length){
+     console.log(longestWord = arrayOfWords[i]);
```
fonsogms: so here you don't need to do `console.log(longestWord=arrayOfWords[i])` you can do ` longestWord=arrayOfWords[i]`

src/functions-and-arrays.js
```diff
@@ -1,19 +1,105 @@
...
 // Iteration #2: Find longest word
+
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+function findLongestWord (arrayOfWords){
+let longestWord = '';
+
+  for( let i = 0; i < arrayOfWords.length; i++){
+    if(longestWord.length === arrayOfWords.length){
+      continue;
```
fonsogms: if(longestWord.length === arrayOfWords.length){
      continue;}
this condition is not necessary,  :)

src/functions-and-arrays.js
```diff
@@ -1,19 +1,105 @@
...
+
+  for (let i = 0; i < arrayOfNumbers.length; i++){
+    sum += arrayOfNumbers[i];
+  }
+  if(sum.length === 1){
+    return sum;
+  } else if (arrayOfNumbers.length === 0 || sum === 0){ // I figured it out that sum is already zero, so that's why it wasn't working 
+    return 0;
+  }
+
```
fonsogms: nice job!

PR #1559, opened 2020-03-25
BER-FWD- Daniela

daniela-kirilova:


src/functions-and-arrays.js
```diff
@@ -46,7 +125,18 @@ const wordsCount = [
...
-
+function howManyTimes(array, name){
+  let found = 0
+  if(repeat.length === 0){
+    return null
+  }for (let word of array){
+    if(word === name){
+      found++
+    }
+    return found++
```
fonsogms: I think here you just wanted to return found? instead of found ++? :)

PR #1560, opened 2020-03-25
first four tasks

Deuker:


src/functions-and-arrays.js
```diff
@@ -1,16 +1,65 @@
...
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+let sum = 0;
+let totalSum = 0;
+
+
+function sumNumbers () {
+  for (let i=0; i<=numbers.length-1; i++){
+    let sum = numbers[i] + totalSum;
```
fonsogms: Nice job!
Few comments here:

1. here you could have used only sum to return the whole sum of the numbers. total sum was not needed.
2. See when you were calculating sum you were recreating a new sum in every loop `let sum` but if you just did this `sum= sum+numbers[i]` you could still have gotten the total at the end of the loop and save your self some code :)
and then in the return you just `return sum`



also you could have calculated sum like this sum

src/functions-and-arrays.js
```diff
@@ -1,16 +1,65 @@
...
+  }
+  return totalSum;
+}
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+let sum = 0;
+let totalSum = 0;
```
fonsogms: same comments regarding the total sum :)

PR #1561, opened 2020-03-25
BER-FTWD-Nate

mcfwesh:


src/functions-and-arrays.js
```diff
@@ -1,20 +1,107 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b){
+  if (a > b){
+    return a;
+  }
```
fonsogms: well done!

src/functions-and-arrays.js
```diff
@@ -1,20 +1,107 @@
...
+  return b;
+}
 
 // Iteration #2: Find longest word
+function findLongestWord(arr){
+  let maxWordCount = 0;
+  let checkedWords = [];
```
fonsogms: Cool approach!
You didnt need an array tho,
You could have use a variable called `let longestWord=""`
and then when comparing everyword with longest word `if(arr[i].lenght>longestWord.length)`
you assign the new longestWord to longestWord `longesWord=arr[i]`

PR #1562, opened 2020-03-25
BER-FTWD-Sami

Alba-Tros:
Done

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,558 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(arg1, arg2) {
+    if (arg1 > arg2) {
+        return arg1;
+    } else if (arg1 < arg2) {
+        return arg2;
+    }
+    return arg1;
+}
 
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+    "mystery",
+    "brother",
+    "aviator",
+    "crocodile",
+    "pearl",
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,558 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(arg1, arg2) {
+    if (arg1 > arg2) {
+        return arg1;
+    } else if (arg1 < arg2) {
```
fonsogms: you could have done `else{ return arg 2}`
and saved some code with the else if and the other return :)

src/functions-and-arrays.js
```diff
@@ -1,73 +1,558 @@
...
+    "pearl",
+    "orchard",
+    "crackpot"
+];
+
+function findLongestWord(arrOfWords) {
+    const longestWord = arrOfWords.sort(function(a, b) {
+        return b.length - a.length;
+    });
+    return longestWord[0];
```
fonsogms: nice approach! :)

src/functions-and-arrays.js
```diff
@@ -1,73 +1,558 @@
...
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(arrayOfNumbers) {
+    let sumOfNumbers = 0;
+    for (let i = 0; i < arrayOfNumbers.length; i++) {
+        sumOfNumbers += arrayOfNumbers[i];
+    }
+    return sumOfNumbers;
```
fonsogms: well done!

PR #1563, opened 2020-03-26
PAR WDFT 0320 - Guillaume Marcellin - ongoing

guiohm432:
Once again when i forked and then git clone I had the folder and files on which I worked with in January.
So I finish the Lab I was far from finishing last time but it seems that's now the lab has some modifications.
Hope that's not a problem.


ta-web-paris:
[Michel] OK. Nice job. Below some tips:

- Iteration 1: Alternative: keep in mind the Math.min(a,b) function.
- Iteration 3: Better call the result "sum" instead of "count". You can set sum to 0 inside the function. You don't need to care about length 0 (since you return "count" at the end), nor 1 (1st iteration of your for-loop)
- Iteration 4.1: Reuse function sumArray(arr) (iteration 3) and divide it by the length of arr. Keep in mind the .toFixed(n) function (to set n decimals after . -> ex: a= 4.911; a.toFixed(1) -> 4.9)
- Iteration 4.2: Transform the word arr into a length arr (via ForEach), and then reuse 4.1 :)
- Iteration 5: Alternative: use a forEach loop and  if (indexOf(word) === lastIndexOf(word)) -> push into an array of uniques.
- Iteration 6: return arr.includes(word) (remove the if-else) since it returns a boolean
- Iteration 8.1 : prod is not called. Your end condition should not be fixed to 16: prefer "arr[0].length" for X and "arr.length" for Y.
- Iteration 8.2 : reuse 8.1 and add global comments (for ex, the direction you calculate Hor, Vert or Diags)

Anyway, good job. I think you got familiar with for-loops :) Try now to reuse more by making your code DRYer (DRY: don't repeat yourself)

PR #1565, opened 2020-03-26
PAR WDFT March Vladimir Leloup ongoing

Vlp90:
Exercice done

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -47,6 +47,10 @@ function sumNumbers(num) {
 	return result;
 }
 
+
+function sum(num) {
+  
+}
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Hey Vladimir :)

Great job ! A few points for you :
- Iteration 1 = you can delete the return word in the end of the function, it's not useful (it's greyed out because ton function does not need it)
- Iteration 2 = you can write "return --the value--" instead of "return longestWord = --the value--"  it's shorter ;)
- Iteration 4 : your code is good ! check this shorter one if you want to go deeper :
>> function averageNumbers(array) {
  if (array.length === 0) {
    return null;
  } else {
    let sum = 0;
    for (i = array.length - 1; i >= 0; i--) {
      sum += array[i];
    }
    return sum / array.length;
  }
}

Keep it up ! 
Tell me if you have any question !

Nina

PR #1571, opened 2020-03-26
[PAR-WDTF-032020] – Rebecca Norén 🚁 ongoing

rebeccanoren:
I completed all of the mandatory tasks and passed the Jasmine tests :)


TatijanaRajcic:
Hey Rebecca ! Thanks for your work! Every mandatory test  passed, so here are only recommandations to make it even better:
- Iteration 1: nice use of Math.max! Otherwise you could habe used what we call ternaries:

```
  return a > b ? a : b;
}
```
If a>b then return a otherwise return b :)

- Iteration 2: nice! another way of writing `arr.length === 0` would be `!arr.length`. It's shorter :D 

- Iteration 3: be careful ! you have a mix up between "numbers" which is the array you are passing as an argument to the function, and "arr" that you are using inside the function, but shouldn't. Here, it works because "arr" is defined, but it's not the actual array that you are passing as a argument. So just replace "arr" with "numbers"!. For the first case, you could right "return 0" directly. Today you are going to see Higher Order Functions, you'll see one function called "reduce". You will be able to do the entire operation, with just one line, using "reduce" :)

- Iteration 4.1: nice use of your sum Function inside the average one :) 

- Iteration 4.2: nice! today you gonna see a higher order function called "map" which could be useful here, to do the exact same thing but shorter :)

```
function averageWordLength(arr) {
  if ( arr.length === 0 ) { return null; }
  let wordsLengthArray = arr.map( (word) => { return word.length } );
  return averageNumbers(wordsLengthArray);
}; 
```
map is making a copy of the original array you are applying map to, and modifies every element according to the callback function you are passing to it. Here, with arr.map you are replacing every word with word.length. Which means that wordLengthArray is now a array full of lengths (instead of words). Therefore, you can apply the function you just wrote before "averageNumbers"

- Iteration 5: nice use of indexOf :) 

- Iteration 6: "includes" already return true or false, so you don't have to add extra conditions. That would be enough:

```
  return arr.includes(word);
}
```

- Iteration 7: niiiice use of guards! (condition && whatIDisplayIfThisConditionIsTrue). This afternoon you gonna see anoooooooother higher order function, "filter". You will be able to use it here!

```
function howManyTimes(wordsArray, word) {
  return wordsArray.filter(el => el === word).length;
}

```

Applying filter to an array gives you a copy of the original array, with the elements that respect the "filter rule" which is basically the callback function you are passing to the "filter" function.

Nice work Rebecca!! If you try and do the bonuses, don't hesitate to send me a message :)))

PR #1584, opened 2020-03-27
PAR WDFT Nelly

LuxiaAnon:
Hi !
I tried hard but, I don't know there is something I didn't get for the unique array :( 
And I don't understand why jasmine do not like my 3.1 bonus because I test every cases of the specs on repl and it worked good ❓ ❗️ 
Aaaand I must confess I get scared by reading the last bonus one 🐲 

Have a nice week end 🏠

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,139 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers (number1,number2){
+  if(number1>number2){
+    return number1
+  }else if (number2>number1){
+    return number2
+  } else{
+    return number2
+  }
+  }
+  maxOfTwoNumbers(2, 1);
+  maxOfTwoNumbers(5, -7);
+  maxOfTwoNumbers(1, 3);
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Hey Nelly, good job !
Very nice use of callback function in iteration 4 :)

A few points : 
- Iteration 2 : you don't need the variable "taille", because to compare words lengths you can compare array[i].length et longword.length (if you define longword = " " at the start)

- Iteration 3 : you don't need to define special cases for booleans (true = 1, false = 0 by default !) and you don't need to calculate the average at the end because you want the total length !  Here is a code that works (with a bonus check for objects, that are not valid data for this exercise)
>>function sum(arr) {
  if (arr.length === 0) {
    return 0;
  }
  let sum = 0;
    if (typeof arr[i] === "object") {
    } else if (typeof arr[i] === "string") {
      sum += arr[i].length;
    } else {
      sum += arr[i];
    }
  }
  return sum;
}

- Iteration 5 : careful your lastIndexOf method searches for the LAST appearance of the word, so that could give you some doubles. You can use this instead : 
>> for (let i = 0; i < arr.length; i++) {
    if (uniquifiedArray.indexOf(arr[i]) === -1) {
      uniquifiedArray.push(arr[i]);
    }
  }

- The bonus is extra spicy, we will push the solution later :) 

Nice work here ! Keep it up :)
Hit me up if you have any question !

Nina

PR #1585, opened 2020-03-28
[PAR - WDFT - 0320] Florian Jomain

florian-jomain:


ta-web-paris:
Hey Florian ! 

Excellent work except for the last iteration, your reduce isn't working properly and its syntax is very strange ("arr[word]++" is not possible for example haha) you can read the doc again, there is an accumulator to use, but not this way ! you can do like this instead (without reduce) : 

>> function howManyTimes(arr, word) {
  let count = 0;
    if (arr[i] === word) {
      count++;
    }
  }
  return count;
}

I let you look online for reduce since you used a lot of nice methods, you should try and practice for this one !

Great job :) 

If you have questions, hit me up !

Nina

ta-web-paris:
Hey Florian !

Excellent work (everything works not like I said before, sorry haha)
If you have questions, hit me up !

Nina

PR #1586, opened 2020-03-28
[PAR WDFT 032020] William Hermozo

williamhzo:
I didn't manage to finish the bonus iteration #8 (matrix) - I will go back to it


ta-web-paris:
Hey William ! 

Nice job !!
A few points for the future :
- It. 1 : be careful using things such as your "greatest = a" without "var" or "let" or "const", in some strict environments it won't work (greatest is undefined) so you should declare variables ! Here it works :)

- It. 2 : you don't need "maxLength" you can use "longestWord.length" instead ("" is empty so length = 0) it's shorter and cleaner ;)

- It. 3.1 : no need to specify 1 or 0 for booleans (they are already 1 or 0)

Otherwise very very nice code :)
Hit me up if you have questions !

Nina

PR #1587, opened 2020-03-29
PAR WDFT 202003 - Anne Patana - Week1 Day4 Lab

apatana:
Test case "should return null when called with an empty array" does not pass in Jasmine for the following iterations (and I'm not sure how to make it pass):

Find the longest word
Calculate the average of an array of numbers
Calculate the average of an array of strings
Unique array
Find elements

Test case "should return true if the word we are looking for is in the array" does not pass in Jasmine for the following iteration (even though the code itself will return "true" in this case):
Find elements

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -2,28 +2,27 @@
 
 function maxOfTwoNumbers (num1, num2) {
   if (num1 > num2) {
-   return num1
+    return num1
   } else if (num1 < num2) {
-      return num2
-    } else {
-      return num1
-    }
+    return num2
+  } else {
+    return num1
   }
-maxOfTwoNumbers(2, 2)
+}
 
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Hey Anne !

- For the "empty array" test you can use this little snippet at the beginning of your functions :
>> if (arr.length === 0) {
    return null;
  }

That way, if the array is empty, the function will return null (otherwise it will try and test values inside of the array, but they don't exist so you get "undefined" or an error !)

- I don't know why it's not working for the "true" test, the code is correct... Maybe it expects a direct boolean value ? In that case you could write : 
>> if (array.includes(word)) {
return true
} else return false

or in the new ternary syntax we saw this morning (hehe) it would be : 
>> (array.includes(word)) ? true : false

Otherwise it's very good ! Nice job :)
Hit me up for potential questions !

Nina

PR #1597, opened 2020-03-29
[PAR-FT-032020] Maxime Paolone 

mxpln:
:sun_with_face:

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,453 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Hey Maxime :)

Nice job ! A few points for future code : 

- You should put your "if (! array.length)" checks at the beginning of the functions, otherwise it will run the code until it reaches that point for an empty array (and here it's fine but in a very long function it takes time for nothing, if you put it at the start boom ! more efficient)

- It. 4 : do not put "console.log" before your "total +=", it's confusing ! If you need to test ok, but then comment it and write just the code needed for the function (it makes more sense)

Otherwise everything is great, good job !
Hit me up for questions if you have any !

Nina

PR #1598, opened 2020-03-29
Ironhacker Paris WD FT March

sdubois90:


ta-web-paris:
Hey Sebastien ! 

Nice job ! 
A few points for your future code : 

- it 2 : you should put temporary variables inside the function (it's not wrong if you don't but it makes more sense that way, and it's cleaner to read)
- you should avoid "console.log" instead of "return" : it's not the same thing (console.log(null) is not valid) and it may throw errors in stricter codes. you should remove them once you have tested the value
- be careful with indentation, for the same reasons as above ! in your jobs a lot of different people will look at your code and it can get confusing if every code is presented in a different way. 

- it 4 : you can use "array.length" instead of using a counter to calculate the average :)

- it 6 // 7 : why the newWord ?? noooo if i want to search "cat" instead if will search for "machine" anyway. you have a "word" parameter, use it ! your parameters are ARRAY and WORD, so code your function to look for WORD in ARRAY (you can even have 6 parameters if you need them, but if you have, use them all !)

PR #1600, opened 2020-03-29
PAR-FT-032020 Yuval Bugajski

YuvalBu:


ta-web-paris:
Great job Yuval ! 
For the bonus I don't know if Franck wants to review it in class, otherwise it will be pushed in his folder soon ! Nice use of reduce and map :) 

Hit me up for questions if you need,

Nina

PR #1601, opened 2020-03-30
PAR WDFT 0320 Teddy ROGERS

TeddyRod:


ta-web-paris:
Hey Teddy !

Good job ! Nice use of methods (includes, push...) and clean code ! Here is some feedback, i keep the lab open since it's not finished but there is absolutely no pressure to finish it, if you have free time at some point you can, if not it's okay you did most of it ! 

A few points : 
- When you test empty arrays you shouldn't write "if (array === "")" because an empty array is not an empty string ! You can write "if (array === [])" or "if (!array.length)" ("!" means "not", so "if no array length" => empty) 

- Does your iteration 5 works in Jasmine ? I can't tell if it does by reading it, but I know in some cases it will not work (it's depending on length match between i and j, there could be special cases) but here is a code that works (the - - - at the beginning of the lines is to make the space because the indentation doesn't work in this comment !)

>> function uniquifyArray(arr) {
  -- if (arr.length === 0) {
  ----   return null;
 --  }
 --  const uniquifiedArray = [];
 --  for (let i = 0; i < arr.length; i++) {
 ----   if (uniquifiedArray.indexOf(arr[i]) === -1) {
 ------      uniquifiedArray.push(arr[i]);
 ----    }
 --  }
 --  return uniquifiedArray;
}

Hit me up if you have questions :)

Nina

ta-web-paris:
Actually no I close it haha

PR #1606, opened 2020-03-30
[PAR-FT-03/2020] - Benjamin Dean

bdean05:
My functions

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,11 +1,42 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(a, b) {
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Hey Benjamin :)

Nice job on this !
I leave the lab open for you to work on it if you have some free time (but focus mainly on this week's labs, this is bonus for you to practice and improve) just a quick thing : you shouldn't return 0 or null if the array is undefined, because undefined is like an error, so we want to catch it ! But smart to use 0 or null for empty arrays !

Clean code, I'm sure in a week you can finish this one in 20 minutes 

Hit me up if you have questions !

Nina

PR #1608, opened 2020-03-30
Louis Alibert - PAR "uniquify"

lalib19:
Jasmine was telling me Iteration 4 level 2 wasnt working properly but I got stuck on it for a long time and couldnt figure out what was wrong with it.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,82 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(a,b) {
+    if(a > b) {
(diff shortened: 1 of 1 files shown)


ta-web-paris:
Hey Louis :)

I think I found the problem for the 4.2 : 
- you have "wordLength" as a parameter for the averageWordLength function, you should have "wordsArr" as a parameter instead ! 

I leave the lab open since it's not finished but no pressure, if you have some free time you can finish it to practice, as you want. Focus on this week's labs it's the priority :)

Hit me up for questions !

Nina

PR #1611, opened 2020-03-30
Maxime Ferret PAR 0320

mferret:
Stuck with functions, after reinforcement (31/03) this lab will be submitted properly finished within few days.

Thanks Tatijana for the reinforcement

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5502
+}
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hey Max ! 

A few points to explain further : 
- It 1 : all good
- It 2 : you can pass an array as parameter for your findLongestWord function, and then do the logic with "array" instead of "words". That way, when you call the fonction, you can put any array variable as an argument and it will adapt the logic to that new value ! Here how it works : 

image

If you have any questions, please ask me ! You can try and work on this lab when you have a bit of time, but please focus on the labs of this week first !  And don't hesitate if you're stuck I can help ! 

Nina

PR #1617, opened 2020-03-31
[BCN-FT-032020] Hélène Stephan

HeleneStephan:
Work in progress, made with Ariadna, not finished yet.
I can't think about anything but sleeping right now so it's impossible to complete it tonight...

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,443 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(x, y) {
+  if (x > y) {
+    return x;
+  } else {
+    return y;
+  }
+}
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+function findLongestWord(words) {
+  var longestWord = "";
+  if (words.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < words.length; i++) {
+    if (words[i].length > longestWord.length) {
+      longestWord = words[i];
+    }
+  }
+  return longestWord;
+}
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,443 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(x, y) {
```
tawebbcn, line 2: Perfectly done! :)

src/functions-and-arrays.js
```diff
@@ -1,73 +1,443 @@
 // Iteration #1: Find the maximum
-
```
tawebbcn, line 2: Iteration 1. Perfectly done! :)

src/functions-and-arrays.js
```diff
@@ -1,73 +1,443 @@
...
+  var sum = 0;
+  if (numbers.length === 0) {
+    return 0;
+  }
+  for (let i = 0; i < numbers.length; i++) {
+    if (typeof numbers[i] === "number") {
+      sum += numbers[i];
+    } else if (
+      typeof numbers[i] === "string" ||
+      typeof numbers[i] === "boolean"
```
tawebbcn, line 55: For booleans, it should have check the value, same as in the case of number. Not the length of the value.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,443 @@
...
+
+function averageWordLength(wordsArr) {
+  if (wordsArr.length === 0) {
+    return null;
+  }
+  var wordString = wordsArr.join("");
+  return wordString.length / wordsArr.length;
+}
+
+function avg(arr) {
```
tawebbcn, line 98: this function should have been very similar to the function sum, that checks every element for it's type on a for loop.

tawebbcn:
Overall very good job! You finished most iterations with their bonuses! Very clean good and great use of previous declared functions and the .join method. Keep up the good work :)

PR #1618, opened 2020-03-31
[BCN-PT-032020] Siiri S.

fetaplop:
I only finished the 1st one


tawebbcn:
The first Iteration was done correctly and it seems you tried hard on the following ones, however, there was too much of the Lab unfinished.

PR #1619, opened 2020-03-31
[BCN-FT-Maxime Lagrange + Alejandro Ollé]

alex-olle:
We added some code to the latest version we updated before, such as new task completed.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,488 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num1 < num2) {
+    return num2;
+  } else {
+    return num1, num2;
+  }
+}
+
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(array) {
+  if (array.length === 0) {
+    return null;
+  } else if (array.length == 1) {
+    return array[0];
+  } else {
+    let longestWord = "";
+
+    for (let i = 0; i < array.length; i++) {
+      if (array[i].length > longestWord.length) {
+        longestWord = array[i];
+      }
+    }
+    return longestWord;
+  }
+}
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(array) {
+  let sum = 0;
+  if (array.length === 0) {
+    return 0;
+  } else if (array.length === 1) {
+    sum += array[0];
+    return sum;
+  } else {
+    for (let i = 0; i < array.length; i++) {
+      if (array[i] === 0) {
+        sum = 0;
+      } else {
+        sum += array[i];
+      }
+    }
+    return sum;
+  }
+}
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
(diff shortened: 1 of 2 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,488 @@
...
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(array) {
+  if (array.length === 0) {
+    return null;
+  } else if (array.length == 1) {
```
tawebbcn, line 27: this check for " array.length == 1, return the single word " is actually redundant. You can see that if you remove it everything checks ok and this is because your for loop will also check if the array only has a single word without any issue. Also, remember to use "===" instead of "==" for comparisons unless you actually want to convert data types for some reason ( which usually you won't even do with "==" )

src/functions-and-arrays.js
```diff
@@ -1,73 +1,488 @@
...
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(array) {
+  let sum = 0;
+  if (array.length === 0) {
+    return 0;
+  } else if (array.length === 1) {
```
tawebbcn, line 49: Same as last comment. This is validation is redundant as it is already worked on the for loop for single number arrays.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,488 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(array) {
+  let average = 0;
+  let sum = 0;
+  if (array.length == 0) {
+    return null;
+  } else if (array.length === 1) {
```
tawebbcn, line 73: Same as last comment. This is validation is redundant as it is already worked on the for loop for single number arrays.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,488 @@
...
+  "fuel",
+  "palace"
+];
+
+function averageWordLength(array) {
+  let average = 0;
+  let sum = 0;
+  if (array.length === 0) {
+    return null;
+  } else if (array.length === 1) {
```
tawebbcn, line 103: Same as last comment. This is validation is redundant as it is already worked on the for loop for single string arrays.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,488 @@
...
+  "simple",
+  "bring"
 ];
 
+function uniquifyArray(array) {
+  if (array.length === 0) {
+    return null;
+  } else {
+    let newArray = [];
+    for (let i = 0; array.indexOf(array[i]) != -1; i++) {
```
tawebbcn, line 135: This was a good attempt to solve it! creating an empty array outside of the for loop and pushing elementsinto it!
Couple of  pointers:
1 - the condition in the for loop is a little messy, I understand the logic trying to avoid repeated words but doing a regular loop and working with the results inside the for loop would be better approach.
2 - For example. Inside the for loop you could create an if statement that only add the current string ( array[i] ) to the newArray IF that same string is not currently included in newArray. Using the .indexOf method.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,488 @@
...
+  "starting",
+  "matter",
+  "eating",
+  "matter",
+  "truth",
+  "disobedience",
+  "matter"
 ];
 
+function howManyTimes(array, word) {
```
tawebbcn, line 184: Good progress on this iteration! 
- You could use the || operator if the first condition and do ( array.length === 0 || !array.includes(word) ) since they both return 0.
- for the other two checks, you could create a counter variable and do a for loop or forEach that adds 1 everytime the word being checked inside the loop is the same as the word passed as an argument in the function.

tawebbcn:
Overall very good  job. You almost finished the Lab which is hard for the first ones.
- Good that you found and started working with .includes. 
- Try to watch redundancy in your code. 
- The logic behind all was good and easy to understand. 
- All well separated for readability. 
Keep up the good work!

PR #1621, opened 2020-03-31
lab-js-functions-and-arrays-Alberto Pagoria

alpagor:
Did the lab until late night together with Lucia just did 50% and problems with jasmine.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,434 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
+  } else {
+    return num1 && num2;
+  }
+}
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
-
+//const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words2 = [];
+function findLongestWord(array) {
+  var max = array[0];
+  //max = crocodile
+  if (array.length <= 0) {
+    return null;
+  } else {
+    for (let i = 0; i < array.length; i++) {
+      if (array[i].length > max.length) {
+        max = array[i];
+      }
+    }
+    return max;
+  }
+}
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(array) {
+  let sum = 0;
+
+  for (let i = 0; i < array.length; i++) {
+    sum += array[i];
+  }
+  return sum;
+}
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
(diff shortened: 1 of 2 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,434 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
+  } else {
+    return num1 && num2;
```
tawebbcn, line 8: Even if this this return (num1 && num2) doesn't alter the result. It also doesn't make much sense. You could either just pick one variable to return (the iterarion asks you to return any) or apply >= to one of the conditions above instead.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,434 @@
...
+  }
+}
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
-
+//const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words2 = [];
+function findLongestWord(array) {
+  var max = array[0];
```
tawebbcn, line 15: instead of max, you could name the variable something like longestWord just to be more specific.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,434 @@
...
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
-const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+//const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+const numbersAvg = [];
+
+let sum = sumNumbers(numbersAvg);
+
+function averageNumbers(array, sum) {
```
tawebbcn, line 48: It was a really good approach to try to use the previous function result inside this one (to not redo that part) but you were missing a couple of pointers:
- First, by declaring a variable sum outside the function (which in other case would be totally fine) the system is having conflicts with Jasmine, so that is what is breaking your code.
- Also, you can see that the second argument in the function (sum) has the same name as the variable created outside the function. So the function will try to get that argument value when the function is invoked and not the result from the function sumNumbers(numbersAvg).
- A better way to do it would be to remove the variable declaration outside, remove the second argument and just call the function sumNumbers(array) inside the function averageNumbers and use the result.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,434 @@
...
+    return sum;
+  }
+};
+
+//console.log(sumWordLength(wordsArr));
+
+let sum = sumWordLength(wordsArr);
+//sum = 53
+
+function averageWordLength(array, sum) {
```
tawebbcn, line 87: Same feedback as before iteration. The approach is very food but the execution could use some work. Instead of arguments and variables, just try to invoke the previous function inside this one with the argument array.

tawebbcn:
It was fine work overall, you manage to fully finish some iterations which is good and you tried hard on the others. I would recommend you to review a little more the lecture on functions and how arguments work and are passed. Remember this is very hard at first. Keep up the good work!

PR #1624, opened 2020-03-31
BCN-FT-JosephBuckley/Arslane

jbuckley213:
Javacript functions and arrays completed by Joseph Buckley and Arslane

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,480 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(numOne, numTwo) {
+  if (numOne > numTwo) {
+    return numOne;
+  } else if (numTwo > numOne) {
+    return numTwo;
+  } else {
+    return numOne;
+  }
+}
 
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  if (arr.length == 0) {
+    return null;
+  } else if (arr.length == 1) {
+    return arr[0];
+  }
+
+  var longestWord = "";
+  for (var i = 0; i < arr.length; i++) {
+    if (longestWord.length < arr[i].length) {
+      longestWord = arr[i];
+    }
+  }
+  return longestWord;
+}
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(nums) {
+  if (nums.length == 0) {
+    return 0;
+  } else if (nums.length == 1) {
+    return nums[0];
+  }
+  var sum = 0;
+  for (var i = 0; i < nums.length; i++) {
+    sum += nums[i];
+  }
+  if (sum === 0) {
+    return 0;
+  } else {
+    return sum;
+  }
+}
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-// Level 2: Array of strings
-const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(numOne, numTwo) {
+  if (numOne > numTwo) {
+    return numOne;
+  } else if (numTwo > numOne) {
+    return numTwo;
+  } else {
```
tawebbcn, line 7: If you use >= in one of the above cases then you would not need to check this last condition. Less code for good :)

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  if (arr.length == 0) {
```
tawebbcn, line 24: Remember always to use triple equal "===" unless you are trying to compare different types of variables . Double equal "==" will give you the same result now but might give you some issues in the long run.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+    return null;
+  } else if (arr.length == 1) {
```
tawebbcn, line 26: You can actually remove this condition to check if the array only has one element. This will be checked in the for loop so it is redundant code.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+}
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(nums) {
+  if (nums.length == 0) {
+    return 0;
+  } else if (nums.length == 1) {
```
tawebbcn, line 46: Same as before. Redundant condition.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+    return nums[0];
+  }
+  var sum = 0;
+  for (var i = 0; i < nums.length; i++) {
+    sum += nums[i];
+  }
+  if (sum === 0) {
```
tawebbcn, line 53: This conditional is highly redundant. You can see the logic says: if sum equals 0, return 0, else return sum. But in the case sum is 0 i could also return sum which is also 0 ;) It all works, of course, but you can easily reduce your code removing the conditional and just do: return sum.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+  }
+}
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(avgNums) {
+  if (avgNums.length == 0) {
+    return null;
+  } else if (avgNums.length == 1) {
```
tawebbcn, line 66: Same as before. Redundant condition to check single element array.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+    return avgNums[0];
+  }
+  var avg = 0;
+  for (var i = 0; i < avgNums.length; i++) {
+    avg += avgNums[i];
+  }
+  avg = avg / avgNums.length;
+  avg = Math.sqrt(avg * avg);
```
tawebbcn, line 74: Not sure why you used Math.sqrt here but it is actually not needed.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+  "smart",
+  "chaos",
+  "fuel",
+  "palace"
+];
 
+function averageWordLength(avgWords) {
+  if (avgWords.length == 0) {
+    return null;
+  } else if (avgWords.length == 1) {
```
tawebbcn, line 95: Same as before. Redundant condition to check single element array.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,480 @@
...
+    return avgWords[0].length;
+  }
+  var sum = 0;
```
tawebbcn, line 98: Try to be careful when naming your variables, "sum" doesn't make sense in this function does it?

tawebbcn:
Excellent work. You did almost everything without any mayor issues. Particularly good work on the last 3 iterations where you used more advanced techniques like double loops, splice method and counters in a very clean way. I would only recommend you to check for redundancies in your code and using the correct comparison operators.

PR #1626, opened 2020-03-31
[BCN-FT-032020] Daniel Hamekasi

wangagat:


src/functions-and-arrays.js
```diff
@@ -1,73 +1,517 @@
...
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  if (arr.length === 0) {
+    return null;
+  }
+  var longest = arr[0];
+  for (var i = 0; i < arr.length; i++) {
```
tawebbcn, line 27: The conditional inside the for loop works good but you could refactor it a little bit better.

```suggestion
  if (arr[i].length > longest.length) {
      longest = arr[i];
    }
```

Removes the continue and else part to make it shorter and more concise.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,517 @@
...
+    if (longest.length >= arr[i].length) {
```
tawebbcn, line 28: The conditional inside the for loop works good but you could refactor it a little bit better.

```suggestion
  if (arr[i].length > longest.length) {
      longest = arr[i];
    }
```

Removes the continue and else part to make it shorter and more concise.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,517 @@
...
+}
+
+//Bonus 3.1
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arr) {
+  if (arr == "") {
```
tawebbcn, line 57: Although this works, for best practices we would recommend to use something like ( arr.length === 0 ) to check if an array is empty. Using double equal "==" means you are using coercion and converting data types from an array to a string and it might be a little bit confusing to read and create issues in the long run.

tawebbcn:
Overall almost perfect job. I see you finished all iterations and even tried hard on the last bonus which was the hardest of all! If you want, you can go see the solution on your student portal for your peace of mind :) Other than that, good use of the forEach instead of loops and using previously declared functions. Keep up the good work!

PR #1628, opened 2020-03-31
[BCN-FT-ALEX A.] - done

skanndar:


src/functions-and-arrays.js
```diff
@@ -1,73 +1,560 @@
...
+  "orchard",
+  "crackpot"
+];
+
+function findLongestWord(arr) {
+  if (arr.length === 0) {
+    return null;
+  }
+  var longest = arr[0];
+  for (var i = 0; i < arr.length; i++) {
```
tawebbcn, line 27: The conditional inside the for loop works good but you could refactor it a little bit better.

if (arr[i].length > longest.length) {
longest = arr[i];
}

Removes the continue and else part to make it shorter and more concise.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,560 @@
...
+
+  for (i = 0; i < arr2.length; i++) {
+    if (typeof arr2[i] === "string") {
+      total += arr2[i].length;
+    } else if (typeof arr2[i] === "number") {
+      total += arr2[i];
+    } else if (arr2[i] === true) {
+      total++;
+    } else {
+      //HOw to throw an error?
```
tawebbcn, line 65: This was not really needed at all and only for you to go more into Jasmine and see how it works in the files, but if you want to know you can, just add: throw new Error("Unsupported data type sir or ma'am") and it will check correct :)

src/functions-and-arrays.js
```diff
@@ -1,73 +1,560 @@
...
+
+function sum(arr2) {
+  var total = 0;
+
+  for (i = 0; i < arr2.length; i++) {
+    if (typeof arr2[i] === "string") {
+      total += arr2[i].length;
+    } else if (typeof arr2[i] === "number") {
+      total += arr2[i];
+    } else if (arr2[i] === true) {
```
tawebbcn, line 62: You can add this condition with an || operator to the second condition to reduce code. The numeric value of the boolean true is always 1 so it will match the same statement.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,560 @@
...
+      total++;
+    } else {
+      //HOw to throw an error?
+    }
+  }
+  return total;
+}
+
+// function sum(arr2) {
```
tawebbcn, line 71: remember to remove unused code snippets (even the commented ones like this) when doing pull request to make your code cleaner.

src/functions-and-arrays.js
```diff
@@ -1,73 +1,560 @@
...
+//   });
+//   return total;
+// }
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arr) {
+  if (arr == "") {
```
tawebbcn, line 100: Although this works, for best practices we would recommend to use something like ( arr.length === 0 ) to check if an array is empty. Using double equal "==" means you are using coercion and converting data types from an array to a string and it might be a little bit confusing to read and create issues in the long run.

tawebbcn:
Excellent work! only a couple of bonuses left. Very good use of double for loop, counters and methods. We will send you the solution for you to check the two bonuses you were having difficulties.

skanndar:
Hey,
thank you for the feedback.

On line
160 I make a comment:
//When trying in repl.it it gives the expected result '5.7', however,
jasmin doesnt take it, Expected 5.111111111111111 to be 5.11.

I couldn't figure out why. Any idea?

skanndar:
when fixing this:

+  if (arr == "") {

Although this works, for best practices we would recommend to use something
like ( arr.length === 0 ) to check if an array is empty. Using double equal
"==" means you are using coercion and converting data types from an array
to a string and it might be a little bit confusing to read and create
issues in the long run.


Jasmin goes from green to red. I don't understand why.

tawebbcn:
- For the avg function Jasmine check you could use the following methods to transform the number to the required one.

return parseFloat(avg.toFixed(2));

- For the validation check on averageNumber function, instead of (arr == "") just write this in the validation of the if statement:

(arr.length === 0)

Try to write instead of copy/pasting in case your are having issues with format.
Both worked correctly on our side, let us know if it didn't for you.

PR #1635, opened 2020-05-02
[PTMAD042020] Roman Baos

Pumpelgit:
Iteration 1 complete. Got mixed up with Visual Studio Code Intelsense snippet. Added (in comments) "if condition" way just in case, but I prefered to use the Mathf.Max method.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -61,6 +61,16 @@ function sum(data) {
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,111 @@
...
+//Bonus
+
+function avg(arr) {
+  if (arr.length<=0) {
+    return null
+  }
+  let result = 0;
+  for (let i = 0; i < arr.length; i++) {
+    
+    if( typeof arr[i] == 'boolean') {
```
plaso, line 96: Aqui igual que en el sum

PR #1668, opened 2020-06-02
[BER-WDFT] Marco Muser

marcomuser:


src/functions-and-arrays.js
```diff
@@ -1,16 +1,77 @@
...
+       } else {
+         onlyNumbersArr.push(1);
+       }
+    }
+  }
+  
+  let sum = 0;
+  for (let j = 0; j < onlyNumbersArr.length; j++) {
+    sum += onlyNumbersArr[j];
+  }
```
Jeff-Moraes, line 65: nice solution, another way to get the sum here is reusing the sumNumbers() function that you have created before.

PR #1669, opened 2020-06-02
Verena WDFT Berlin June 2020

verenamay:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,77 @@
...
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(numbers) {
+  let total=0;
+  for(var i in numbers) {
+    total +=numbers[i];
+    
+  } return total
+}
```
Jeff-Moraes, line 49: great solution ;)

src/functions-and-arrays.js
```diff
@@ -1,19 +1,77 @@
...
+    if(words[j].length == tlength){
+       wholeArr.push(words[j]);
+    }
+  }
+ if(wholeArr.length == 1){
+   return wholeArr[0]
+ }else{
+    return wholeArr
+}
+}
```
Jeff-Moraes, line 36: the logic that you used is on the right way, but it is a bit simpler than what you tried.
here is one solution for this interaction
```
let longestWord = "";
      longestWord = words[i];
    }
  }
  return longestWord;
```

src/functions-and-arrays.js
```diff
@@ -1,19 +1,77 @@
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(numbersAvg) {
+  let total=0;
+  for(var i in numbersAvg) {
+    total +=numbersAvg[i];
+} let average=0;
+average = total/numbersAvg.length; 
+return average;
+}
```
Jeff-Moraes, line 63: to solve this error
`should return null if receives an empty array when called`
you can just check if the numbersAvg that you are receiving in the function has some value inside of it
```
if (numbersAvg.length === 0) return null;
```

src/functions-and-arrays.js
```diff
@@ -1,19 +1,77 @@
...
+
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
+function averageWordLength(wordsArr) {
+  avg = wordsArr.join('').length/wordsArr.length;
+  return avg
+}
```
Jeff-Moraes, line 71: same here if the wordsArr is empty
```
if (wordsArr.length === 0) return null;
```

verenamay: Thanks for your feedback, I will try the other iterations on my own as well :)

PR #1671, opened 2020-06-02
BER WDFT 0629 Gosha

MRzsztk:


src/functions-and-arrays.js
```diff
@@ -1,18 +1,84 @@
...
+  let l=0;
+  let longestWordIndex=0;
+  for (let i=0; i<wordArr.length; i++) {
+    if (wordArr[i].length>l) {
+      l=wordArr[i].length;
+      longestWordIndex=i;
+    }
+  }
+  return wordArr[longestWordIndex];
+}
```
Jeff-Moraes, line 19: to solve this error
`should return null if receives an empty array when called`
you can just check if the wordArr that you are receiving in the function is empty
```
if (numbersAvg.length === 0) return null;
```

src/functions-and-arrays.js
```diff
@@ -1,18 +1,84 @@
...
+    if (typeof element == "string") {
+      replaceArr.push(element.length);
+    }
+    if (element === false) {
+      replaceArr.push(0);
+    }
+    if (element === true) {
+      replaceArr.push(1);
+    }
+  }
```
Jeff-Moraes, line 50: to solve this error
`should throw an error when unsupported data type (object or array) present in the array`
you can just check if the element type is == 'object' and throw the Error
```if (typeof element == 'object') throw new Error("Unsupported data type sir or ma'am");```

File: src/functions-and-arrays.js [hunk 3]
```diff
@@ -1,18 +1,84 @@
...
+  let sum=sumNumbers(lengthsArr);
+  console.log(`${sum} letters in total, over ${lengthsArr.length} words.`);
+  let avg=sum/lengthsArr.length;
+  return avg;
+}
+
+//bonus: mixed average
+const mixedArr = [6, 12, 'miami', 1, true, 'barca', '200', 'lisboa', 8, 10];
+function avg(arr) {
+  return mixedAvg=sum(arr)/arr.length;
```
Jeff-Moraes, line 80: this is the response from Jasmine
`Expected 5.111111111111111 to be 5.11.`
so for that, you can use the method .toFix() like this
```return Number((sum(array) / array.length).toFixed(2));```

PR #1673, opened 2020-06-02
BER-WDFT-PaolaGaray

PaolaGaray:
The following iterations are done:
- Find the maximun
- Find the longest word
- Calculate the sum of array of numbers
- Calculate the average of an array of numbers
- Calculate the average of an array of strings
- Unique array
- Find elements
- Count repetition

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,144 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(num1, num2){
+  if(num1 > num2){
+    return num1;
+   } else {
+     return num2;
+   }
+};
+
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,144 @@
...
+}
+
+
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
+function doesWordExist(arr, word) {
+if (arr == ''){return null};
+if (arr.includes(word)){return true} else {return false}
+}
```
Jeff-Moraes, line 93: in this case, you can also do
```return arrayOfWords.includes(word);```
because .includes() method return true or false ;)

PaolaGaray: thanks!! 🙏

PR #1674, opened 2020-06-02
done

IevaKa:
Done

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -145,8 +145,6 @@ function howManyTimes(wordsCount, word){
   });
   return result;
 }
-
-
 // Iteration #8: Bonus
 
 const matrix = [
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,78 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 >= num2){
+    return num1
+  } else {
+    return num2
+  }
+}
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+function findLongestWord(words){
+  let letters = wordsLength(words);
+  let longestIndex = 0;
+  if(words.length === 0){
+    return null
+  } else {
+    for(let i = 0; i < letters.length; i++){
+      if(letters[i] > letters[longestIndex]){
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,79 @@
...
+    return null
+  } else {
+    let OfwordsLength = wordsLength(wordsArr);
+    let average = sumNumbers(OfwordsLength)/wordsArr.length;
+    return average;  
+  }
+}
+
+
+
```
timbotimber, line 76: there are missing variables in this function. Another solution might be to declare a variable with the value 0 then use a for loop to count the length of each word in the array. You can then return the total variable divided by the length of the number of elements in the array... Something like this: 

  if (wordsArr.length === 0) {
    return null;
  }
  let total = 0;
    total += wordsArr[i].length;
  }
  let average = total / wordsArr.length;

  return average;
}

Let me know if that makes sense!

src/functions-and-arrays.js
```diff
@@ -29,8 +89,37 @@ const wordsUnique = [
...
+    wordsUnique.forEach((wordUnique) => {
+      if(uniqueArray.indexOf(wordUnique) === -1){
+        uniqueArray.push(wordUnique);
+      }
+    });
+    return uniqueArray;
+  }
+}
+
+
```
timbotimber, line 106: Nice! Could also add the index of the array to the final if statement within the forEach for clarity: 

if (uniqueArray.indexOf(wordsUnique[i]) === -1) {
      uniqueArray.push(wordsUnique[i]);

src/functions-and-arrays.js
```diff
@@ -29,8 +89,37 @@ const wordsUnique = [
...
+    let result = false;
+    wordsFind.forEach((wordFind) => {
+      if(word === wordFind){
+        result = true;
+      } 
+    });
+    return result;
+  }
+}
+
```
timbotimber, line 122: awesome! Super clean.

src/functions-and-arrays.js
```diff
@@ -47,27 +136,59 @@ const wordsCount = [
...
+  }
+  
+  for(let y = 0; y < results.length; y++){
+    if(results[y] > finalResult){
+      finalResult = results[y];
+    }
+  }
+  return finalResult;
+}
+
```
timbotimber, line 194: Awesome you attempted this! I couldn't get the solution to run though. Did it work for you? Maybe we could go through it if you like?

PR #1677, opened 2020-06-02
Simon Ertel  [WDFT June 2020 Berlin]

SiEr0212:
Simon Ertel [WDFT June 2020 Berlin]

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,97 @@
-// Iteration #1: Find the maximum
+console.log("hello");
 
-// Iteration #2: Find longest word
+// Iteration #1: Find the maximum
+function maxOfTwoNumbers(a, b) {
+if (a > b) {
+  return a;
+}
+if (b > a) {
+  return b;
+}
+if (a === b) {
+  return a || b;
+}
+}  
+
+ //Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+function findLongestWord(words) {
+  var longestWord = "";
+  if (words.length === 0) {
+    return null;
+  }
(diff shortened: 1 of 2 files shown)


src/functions-and-arrays.js
```diff
@@ -29,9 +108,35 @@ const wordsUnique = [
...
+    }
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i] === word) {
+        return true;
+    }
+}
+return false;
+
+}
+
```
timbotimber, line 139: also looks good! Nice one

src/functions-and-arrays.js
```diff
@@ -47,6 +152,30 @@ const wordsCount = [
...
+      else if (numOfWords === 5){
+        return 5;
+      }
+  
+      
+      } 
+
+}
+
+
```
timbotimber, line 178: I can't make this one work as expected. We can go through it if you like? I would have probably: 

- created a variable to check for true or false
- run a for loop to compare the words in the array to the second argument 
- stop the loop and return the check variable depending on the answer

```
function doesWordExist(wordsFind, word) {
  let check = true;
  for (let i = 0; i < wordsFind.length; i++) {
    if (wordsFind[i] === word) {
      check = true;
      break;
      could do this with the .include method
    }
    check = false;
  }
  return check;
}

```

SiEr0212: Hey Tim, thanks a lot for your feedback. Was really helpfull. 
Yeah didn't declare the i's, thx for the hint. I wondered why it didn't work.
A lot became clearer this morning during the recap of the lab with Jan. I understood the concept of the for of loop and used it in the iterations.
Now the only thing where i would need help to go through would be the last bonus #8.(just copy pasted the solution from Jan)
Should we perhaps discuss it tomorrow afternoon?

PR #1678, opened 2020-06-02
done without bonus

Tyouxik:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,64 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers (num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2
+  } else {
+    return num1, num2
+  }
+}
```
timbotimber, line 10: Niiice. Could refactor slightly without the else if you want it to be shorter

src/functions-and-arrays.js
```diff
@@ -1,19 +1,64 @@
...
+  console.log(longestWord)
+  if (words.length === 0) return null;
+  if (words.length === 1) return words[0];
+  for (let i = 0; i < words.length; i++) {
+    if (longestWord.length < words[i].length) {
+      longestWord = words[i];
+    }
+  }
+  return longestWord 
+}
```
timbotimber, line 25: looks good! Do you need the `if (words.length === 1) return words[0];`? might be able to refactor without

src/functions-and-arrays.js
```diff
@@ -1,19 +1,64 @@
...
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers (numbers) {
+  let sum = 0;
+  for (let j = 0; j < numbers.length;j++) {
+    sum += numbers[j];
+  }
+return sum
+}
+
```
timbotimber, line 38: perfect

src/functions-and-arrays.js
```diff
@@ -1,19 +1,64 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers (arr) {
+  if (arr.length === 0) return null
+  let average = sumNumbers(arr) / arr.length;
+  return average;
+}
+
```
timbotimber, line 48: Nice!

src/functions-and-arrays.js
```diff
@@ -1,19 +1,64 @@
...
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
+function averageWordLength (words) {
+  let wordLength = [];
+  for (word of words) {
+    wordLength.push(word.length);
+  }
+return averageNumbers(wordLength);
+}
+
```
timbotimber, line 61: Did this one run for you? It may have been better to use a for loop (as opposed to for of) and add the values to an empty variable as opposed to an empty array.

src/functions-and-arrays.js
```diff
@@ -29,9 +74,27 @@ const wordsUnique = [
...
+  let uniquifiedArray = []
+  if (array.length === 0) return null;
+  for (word of array) {
+    if (uniquifiedArray.indexOf(word) === -1){
+      uniquifiedArray.push(word);
+    }
+  }
+  return uniquifiedArray
+}
+
```
timbotimber, line 87: nice!

PR #1680, opened 2020-06-02
[WDFT BERLIN June 2020] Ivan Mickovski

imickovski:
Still not finished 4/8.


imickovski:
Actually is 7/7 without the bonuses.

PR #1681, opened 2020-06-02
first steps

Romulo91:
Not done but I#m orking on it.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,89 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers (num1, num2){
+  if (num1 > num2) { 
+    return num1
+  }     
+  else { 
+    return num2
+  }
+}
+
+
 // Iteration #2: Find longest word
+
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+function findLongestWord (wordArray) {
+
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,90 @@
...
 
+function maxOfTwoNumbers (num1, num2){
+  if (num1 > num2) { 
+    return num1
+  }     
+  else { 
+    return num2
+  }
+}
+
```
fonsogms, line 11: nice!

src/functions-and-arrays.js
```diff
@@ -1,19 +1,90 @@
...
+    for ( i = 0; i < wordArray.length; i ++ ) {
+      if (wordArray[i].length > currendWord.length) {
+        currendWord = wordArray[i]; 
+       
+      }
+      }
+      return currendWord
+      console.log(currendWord)
+  } else {
+    return null;
```
fonsogms, line 32: no need to do else, you could do a return nul after the if :)

src/functions-and-arrays.js
```diff
@@ -29,6 +100,22 @@ const wordsUnique = [
...
+  let arr = [];  
+  for (i = 0; i < unique.length; i ++){
+    if (arr.indexOf(unique[i]) == -1){
+        arr.push(unique[i])
+        }
+  }
+  return arr 
+}
+}
+
```
fonsogms, line 116: nicely done!

PR #1683, opened 2020-06-03
Yuri [Berlin WDFT June 2020]

yuri-la-torre-ortiz:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,67 @@
...
+  }
+  return sum
+}
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+averageNumbers = (arrayNumbers) => {
+  average = sumNumbers(arrayNumbers)/arrayNumbers.length
+ return average
```
fonsogms, line 51: you forgot to do a if statement when an array is empty and return null
example:
```
if(arr.length==0){
return null
}
```

src/functions-and-arrays.js
```diff
@@ -1,19 +1,67 @@
...
+}
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
+
+averageWordLength = (stringArray) => {
+  numberArray = []
```
fonsogms, line 58: same comment as above :)

src/functions-and-arrays.js
yuri-la-torre-ortiz: Thanks a lot!  Yeah, I noticed that this morning during the review. 

Thanks for the feedback.  It gives closure.  :D

PR #1684, opened 2020-06-03
WDFT Berlin Nina Smolnikova - half-done - 34/53

nina-sm:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,126 @@
...
+  var longestWord = "";
+  if (words.length === 0) {
+    return null;
+  }
+  if (words.length === 1) {
+    return words[0];
+  }
+ 
+  for(var i = 0; i < words.length; i++){
+    if(words[i].length > longestWord){
```
fonsogms, line 24: here you need to compare the length of both strings/words  and then make the longestWord a new strring so something like this 
```
      longestWord = words[i];
    }
  
  ```

PR #1685, opened 2020-06-03
WDFT Berlin - Leo Costa - partially solved challenges - w1d2

Leoprogramming:
Partially solved exercise done in around 1.5H


src/functions-and-arrays.js
```diff
@@ -1,15 +1,59 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  let result = num1 >= num2 ? num1 : num2;
+  return result;
+}
```
fonsogms, line 5: Nice one

PR #1686, opened 2020-06-03
Vitor Pavan [Web Full Time Course Berlin July]

vitorpavani:
Here is my partial code.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,90 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers (number1, number2){
+  if(number1>number2){return number1;}
+  else {return number2;}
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,90 @@
...
+  if(number1>number2){return number1;}
+  else {return number2;}
+}
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+function findLongestWord (words){
+  let longestword = '';
+  
+  if(words==''){
```
fonsogms, line 12: when empty array you can use an else statement like this:

```
if(array.length==0){
// do something
}
```

PR #1714, opened 2020-06-08
[Remote-PT-062020] Giovanna Izzo

Hachiko9:
Hava problems with the last test for "Bonus: Calculate the sum" and for "Bonus: Calculate the average of a mixed elements array".


valmgisbert:
WOW YOUR CODE, your use of ternaries and different methods blew me away. `filter()`??? I'm swooning. Good use of conditionals, and for your approach for every iteration. Be careful with syntax errors or extra space, and practice spacing for an easier to read code. Amazing work!

PR #1716, opened 2020-06-08
[Remote-PT-062020] Debora Crescenzo

NubivagantYansa:
Exercise completed


valmgisbert:
STARTING WITH A ONE-LINER, WHAT A RIDE. You can refactor the itr #2 but the nested loops show me you understand them; also nice to see you try arrow functions, plus different methods! Love it! Remember in arrow function one-liners you don't need the `{}` (see `averageWordLength` alternative func). Wonderful job!

PR #1717, opened 2020-06-08
ex 1ºPedroLamy

PedroMMLamy:


valmgisbert:
Your first iteration doesn't have any logic in it. Nice use of loops. I am a bit confused as to why you used `alert()` though. Whenever you can, keep working on this!

PR #1721, opened 2020-06-09
[REMOTE - WDPT - JUNE 2020] LYNN W-B

lynn-w-b:
will try to also complete bonus iterations now but wanted to submit this work before class :0)


valmgisbert:
Wonderful use of switch cases and conditionals. In some iterations you could refactor them to lose variables you don't really use. Nicely structured code, practice spacing a bit more for an even more pleasing code. Great work!

PR #1722, opened 2020-06-09
[REMOTE WDPT 0620 ] Cristina Matamoro 

cris-developer:
src/functions-and-arrays.js


valmgisbert:
Very well-structured code; practice your spacing for an even more easy-to-read work. Your grasp on the concepts is seen on your iterations. Whenever you can, return to the bonuses or practice arrow functions on them. Nice job!

PR #1725, opened 2020-06-09
[REMOTE-WDPT-0602] - Alejandro Gonzalez

alejanglez:
**lab-javascript-functions-and-arrays**

No bonuses reached :/


valmgisbert:
You had a solid way of approaching the iterations, and very nice use of different `for` loops. Also good that you console.log() to check your progress, but you don't really need it when invoking the function; that already does the job. Be careful with syntax errors, and practice your spacing for an easier-to-read code! Whenever you can, come back for the bonuses or refactor your code. Good job!

PR #1726, opened 2020-06-09
[REMOTE-WDPT-0602] - Andre Bacalhau

AndrewMastercode:


valmgisbert:
Good job! you can go back to some iterations and refactor them to make it more concise and practice other methods. Practice your spacing for an easier to read code. Whenever you can, return for the bonuses. Keep up the good work!

PR #1752, opened 2020-06-16
[PAR-FT-0620] Menhaj safi

Menhaj01:


TatijanaRajcic:
Hey Menhaj! Thanks for your lab! Some remark:
- maxOfTwoNumbers: you can simplify it, you don't need that many conditions, that would be enough:
```
  if (a > b) {
    return a;
  } else {
    return b;
  }
}
```
- findLongestWord: you don't need the special case array.length === 1, this is handled with your loop
- averageNumbers: be careful, you are dividing avrgNumber many many times, since you wrote this line `avrNumber = sum / array.length;` inside your loop, whereas you should execute it AFTER you finished your sum, so AFTER the sum, which means AFTER the loop. Also!!! You shouldn't write =+ but += : otherwise you are not doing a sum, but you are assigning a value to sum => `sum = +array[i];` this means: sum is equal to the positive value of array[i]. It's not the same at all as `sum += array[i]; `which is what you want. At the very end, what you get is:
```

  var avrNumber = 0;
  var sum = 0;
  if (array.length === 0) {
    avrNumber = null;
  } else {
      sum += array[i];
    }
    avrNumber = sum / array.length;
  }

  return avrNumber;
}
```
- averageWordLength: same thing as beforen you should make your final division OUTSIDE the loop:

```
  var total = 0;

  if (array.length === 0) {
    total = null;
  } else {
      total += array[i].length;
    }
    total = total / array.length;
  }

  return total;
}
```

- avg: it does not work because you should take into account the type of data that you have in your array

- Here are some proposals for the questions that you did not have time to finish, have a look at it and tell me if it helps you understand:

```
function uniquifyArray(arr) {
  if (arr.length === 0) {
    return null;
  }

  const uniquifiedArray = [];

    if (uniquifiedArray.indexOf(arr[i]) === -1) {
      uniquifiedArray.push(arr[i]);
    }
  }

  return uniquifiedArray;
}

function doesWordExist(arr, wordToFind) {
  if (arr.length === 0) {
    return null;
  }

  let found = false;

    if (arr[i] === wordToFind) {
      found = true;
      break; // No need to continue the loop if the item is found.
    }
  }

  return found;
}

// Iteration #7: Count repetition
function howManyTimes(arr, word) {
  let count = 0;

    if (arr[i] === word) {
      count++;
    }
  }

  return count;
}
```

PR #1775, opened 2020-07-07
[PTAMS062020] Mahmut Sarcan

mhmtdl:


src/functions-and-arrays.js
```diff
@@ -1,18 +1,58 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(a,b) {
+  if(a>b) {
```
guido-an, line 3: Use spaces: `if(a > b) {`

guido-an: Same thing in the following lines

src/functions-and-arrays.js
```diff
@@ -1,18 +1,58 @@
...
+    return a;
+  }else if(b>a){
```
guido-an, line 5: Just an _else_ in this case is enough: 
`function maxOfTwoNumbers(a, b) {`
  `if(a > b) {`
    `return a;`
 ` } else {`
    `return b;`
 ` }`
`}`
`maxOfTwoNumbers(15, 30)`

PR #1776, opened 2020-07-07
[WEB-REM-2007] Elise Dujardin

elisedjn:
Everything works on the Jasmine tests but I was not able to manage the columns and the diagonals calculation on bonus iteration 8. 
If I get time tonight, I'll go back to it and try to solve it before tomorrow, but I also wanna try the other bonus lab about Cluedo (I used to love that game 😅)


IH-WebDev-TA-Remote:
Awesome work on the Lab! Yes, I think Jasmine is not checking for diagonal line. We will share the solution later. But still amazing job, keep it up :)

PR #1777, opened 2020-07-07
[WEB-REM-2007] Britta Helm

brittahelm:
Don't think I got the throw error right, and I'm still working on the bonus quest but wanted to submit in time :)


IH-WebDev-TA-Remote:
Awesome work on the Lab! Keep up the good work :) The throw error is really tricky and you would even need to go into the Jasmine program to see what is required. We will share the solution later. Keep up the good work!

PR #1779, opened 2020-07-07
WEB - REM - 2007 Jorge Berrizbeita

JoelRocaMartinez:
This LAB was super challenging. I would lie if I would say that I could do this on my own. I managed to go so far as I did thanks to google and the documentation. I definitelly lack a huge amount of knowledge to do a LAB like this on my own.


IH-WebDev-TA-Remote:
Very Good work overall. Don't worry, Labs are supposed to be challenging and make you look for resources online and go again through your class notes. No developer can work without online resources, notes or previous projects ;) The important think is that you learn what you can and can't do with the tools you have.

PR #1780, opened 2020-07-07
[WEB - REM - 2007] Bodei

BodeiBrouwer:
This is how far I got with the lab on day 2.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,84 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers ( a, b) {
+  if (a > b) {
+    return a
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Good job on the Lab, I see you made a good effort on it but you need to be careful with some things: 
- On it. 2. you are sometimes pointing to the parameter arr and sometimes directly to the words array. You should always point to the parameter. Also, the max variable would make more sense that starts as an empty string, not an array.
- On it. 4. You are returning avg inside every loop iteration. It would make more sense that you try to return the avg outside the loop once you have the full average.

For the rest,  will send the solution so you can check it, but overall great effort. This Lab is challenging! keep up the good work :)

PR #1781, opened 2020-07-07
JS fans and arrays 

kmstamey:
finished as much as possible


IH-WebDev-TA-Remote:
Good job on the Lab though it seems you didn't tried on the last 2 iterations. We will send the solutions later but still try to do those on your own when you have a chance and give your 100%!
Also remember to change the title on your pull requests to [WEB-REM-2007] yourname

PR #1782, opened 2020-07-07
[WEB-REM-2020] Alvaro Sánchez Lamadrid

AlvaroSLam:
Iteration 1, 2, 3, 4 completed

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,15 +1,56 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(a, b) {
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Good effort on the Lab but there was still a lot to do in there. We will send the solutions later but keep working on it when you have the chance :) 
On the sumNumbers iteration, you don't need to call the function. Jasmine is checking that for you. If you remove the function invocation everything works. I also noticed you are forgetting to check some cases like empty values or arrays with negative numbers. 
Try to give your 100% and take note of any lab doubt you may have for Friday and we can take a look at it.

PR #1783, opened 2020-07-07
[WEB-REM-2007] Marta Lourido

MartaLourido:
Hello!

here is my javaScript functions and arrays lab!

all the best,

Marta Lourido


IH-WebDev-TA-Remote:
Good job on the Lab! everything is clean and works but don't be afraid to go for those bonuses ;) Keep it up!

PR #1785, opened 2020-07-07
[WEB-REM-2007] Maeva Merrou

maevamerrou:
As discussed with Jorge, let's have a quick chat tomorrow so we can work together about the resolution of the unfinished tasks.
Thanks a lot!
Maeva

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,77 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(a, b){
+  if (a>b) {return a};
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Good job on the lab!
Most of the red tests between the iterations can be solved creating a condition to return null if the array is empty. Something like ( if (words.length === 0) return null ). 
For the uniquifyArray iteration the logic seems a little odd. You are trying to compare the element only with the next or the one after that, but you need to find a way to compare it to all of the elements. We will send the solution later but still if there is something that is not clear enough let me know :)

PR #1786, opened 2020-07-07
[Remote-WDFT-070620] Alejandro Hernández

alejandrohj:
Hi TA,

I only got 2 little failures on Jasmine. One of them is due to the decimals of the average and I have asked to Manish How to fix it, the other one is the only task I haven't been able to figure out, It's one of the unique array iteration, when you pass me an array with the same elements.

On my side, I have to said that I have completed the iterations 8 and 8.1.

I would like some feedback.

Regards,

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,86 @@
 // Iteration #1: Find the maximum
-
+const maxOfTwoNumbers = (numOne, numTwo) => {
+  if(numOne > numTwo){
+    return numOne
+  }
+  if(numTwo > numOne){
+    return numTwo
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Awesome work on the Lab! the problem you were having with 4.1 is that it takes a mixed array, so you want to call the sum function, not the sumNumbers function. Then everything works.
For the uniquifyArray function, splicing the original array is not recommended as that will mutate the array while the loop is still going. It is better to create a new array and start pushing elements you want in there and just return that. We will post the solution later for you to check but overall amazing work. Keep it up!

PR #1787, opened 2020-07-08
[WEB-REM-2007] Gonzalo

gonzalo-development:
Except fromt the bonus, think all is almost done

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,21 +1,61 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(a,b){
+if(a > b){
+  return a;
+}else{
+  return b;
+}
+} 
+maxOfTwoNumbers(4,5)
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
-
+var max= [];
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Good job on the lab, the logic on the functions is correct but there are some stuff breaking jasmine that need changing in order for the solutions to work.
- First of all, console.log is a tool for developers to check in the console if the code is working, but functions should return a value (using the return keyword) instead. Also, when working with Jasmine, you don't need to invoke the function, Jasmine will do that for you and check the tests.
- on iteration 2, name of the function should be findLongestWord, not maxWord. You know max will be a string, not an array. so it makes more sense you initialise it as an empty string. The variable declaration of max should be inside the function, since it is where it would be used and returned.
- on iteration 3, the sum function it is just a function that is returning the sum of two parameters. As per instructions you need to create one function that will run the whole process of summing the elements of an array (passed via parameters) and returns the sum.
- on iteration 4 you are declaring the variable avgTotal but you are not returning anything.
- on iteration 6, you wrote the function twice. first one works but second one doesn't. Remember to clean your code if there is something you are no longer using.
Overall good job but read more careful the repository instructions and work along with the Jasmine testing system. These solutions should be functions that return the result. The elements that you work with should be in the parameters and every piece of code you write should be in the scope of that function.

PR #1853, opened 2020-08-27
[PAR-WDFT-248] Luca Marongiu

AyloSrd:
[PAR-WDFT-248] Luca Marongiu


N10T:
Hello Luca,

**Iteration 1**
-
- 👍 


**Iteration 2**
-
- prefer to compare with "===" than "==" you are not to comparing the type with "=="
- 👍 


**Iteration 3**
-
- ** prefer to compare with "===" than "==" you are not to comparing the type with "=="**
- forEach ... HOF👍  https://en.wikipedia.org/wiki/Higher-order_function
- line 21 & line 28, you are shadowing the variable "sum"
- "things" => use more explicit words like "param"


**Iteration 4**
- 
- 👍 


**Iteration 5**
-
- You have a shorter way ;) 👍 


**Iteration 6**
-
- array.includes can be easier 👍 


**Iteration 7**
-
- 👍 


🥇 ** Bonus 8 ** 🥇
- 
- good logic! 💯 
but you have an extra step


### Very good job

PR #1854, opened 2020-08-27
[PAR-WDFT-248] Vincent PROUM

VProum:


Fanny-Sbn:
Hello Vincent, thanks for submitting your lab.

All the compulsory iterations work ! It's super great. You even use ternary operators 👌 

Your greatestProduct function doesn't work, you made it more complicated than it is. Your on the right path with nested loops. 

What you want to do is simply calculate all the 4 adjacent numbers horizontally and vertically. 
Basically you need to do your nested loops and you can do this part of your code 
"horiz = matrix[i][j] * matrix[i][j + 1] * matrix[i][j + 2] * matrix[i][j + 3]" **if the j < matrix.length -3**
and this part of your code 
"verti = matrix[i][j] * matrix[i + 1][j] * matrix[i + 2][j] * matrix[i + 3][j];" **if i < matrix.length -3**

Then you just need to find which one has the highest result. 

I let you think about it ;) (you'll receive a proposal for the solution)

PR #1855, opened 2020-08-27
[PAR-WDFT-248] LISA PONCET

lisapnct:
I think I like Javascript even though I didn't manage to do all the bonuses... I passed all tests except 5!


Fanny-Sbn:
Hello Lisa, thanks for submitting your lab ! 

**Iteration 2**
- 
Good it's working but try to use if, else if **and else** instead of else if for your last condition. This **else** will take into account of the other possibilities, in other words you don't need to add another condition after this **else** statement, it's useless. 

**Iteration 3**
-

- Same comment for the if, else if and else statement than the previous iteration. 
- What are you checking with the second else if statement? If all the values in the array are equal to zero ? It's useless, if all the values are equal to zero the reduce method you're using in the next statement will do the job. 
- Nice you already know HOF 👌 

Bonus : 

You probably misunderstood what was the expected result. You should create a function that no matter what was passed in the array( string, numbers or boolean) except object return the sum of all the elements. So you need to use the concept of polyphormism (one example : transforming strings in number by calculating their length)

**Iteration 4**
- 
**Level 1**
-Same comment with if, else if, else statement
-Just a tip: you could have reuse your sumNumbers function in averageNumbers instead of rewriting exactly the same code than before. 

**Level 2**
Same comment with if, else if, else.
Actually you don't need a else if statement you can remove it and simply keep your if statement and replace the last else if statement with a else statement. 

**Level 3**
-Your function just return null if the array is empty, it's doing nothing else...

**Iteration 5 & 6**
-
Good 👌 

**Iteration 7**
-
Your last statement check the two previous one so you can remove them, you can write something like : 

function howManyTimes (arr, word) {
  if (!arr.length) {
    return 0;
  } else {
    return arr.reduce((a, v) => (v === word ? a + 1 : a), 0)
  }
}

**Iteration 8**

You're checking Jasmine tests but the function isn't doing what it is meant for. Re-read the bonus instructions (and not only Jasmine test)

Good job 👍

PR #1857, opened 2020-08-27
[PAR-WDFT-248] Jorell Langmar

JorellLangmar:
Made all exercises except the last bonus that I'll do at home


Fanny-Sbn:
Hello Jorell, thanks for your lab, it's really great ! 

Just a few comment:
-When naming a variable used camelCase so that it can be more readable ( sumnum -> sumNum) 
-Iteration 3 Bonus : you don't need to declare 3 different variables, you can have only one "let sum=0" and add it into your code instead of the other ones and then return it instead of total, you'll save a few lines. 
-Iteration 4 : As I told you yesterday, you can reuse your sumNumbers() function
-Iteration 6 : 👍  Have a look to includes() method
-Iteration 7 : line 195 "incr = incr +0" is useless, you can remove it


Good job ! 🥇

PR #1859, opened 2020-08-27
[PAR-WDFT-248] Frederico Ratsimandresy

Ratsimandresy:
Didn't finish the bonuses but i will finish them!


Fanny-Sbn:
Hello Rico, thanks for submitting your lab. 

**Iteration 2**
-
Your function is working but try to use if, else if and else statement instead of using 3 times a “if”

**Iteration 3**
-
Same comment for if, else if and else statement. Try to name your variable with explicit name you could have name your “reSult1” -> “sum”. And the use of camel case in your variable reSult1 isn’t correct, the camel case is used to distinguish different words like “newArray” or “sumNumbers”
**Iteration 4**
-
**Level 1**
-Your second if statement is useless, and your if statement inside your for loop either. You could have done a if, else statement and declare only one variable that you can return directly. 
According to what you’ve written: 
function averageNumbers(arr) {
  let total = 0;
  if (arr.length === 0) {
    return total = null;
  }
  else {
        total += arr[i];
    }
  }
  return total/arr.length
}

**Level 2**
Exactly the same comment than the level 1, you just need to adapt the code. 

**Iteration 5 & 6**
-
Good 

**Iteration 7**
-
You haven’t done iteration 7 and it’s not a bonus, try it! 😊

Keep on going!

Ratsimandresy:
Hello Fanny,

Thank you very much for the feedback. Will do the changes.
By the way, i did finish the Iteration 7, here is the code (you can check
my code):


function howManyTimes(arr, str) {
let res = 0;
if (arr.length === 0 && !arr.includes(str)) {
res = res;
} else {
const elem = arr[i];
if (elem === str) {
res++;
}
}
}

return res;
 have nice day!

N10T:
Hey Rico,

- res = res 👎 , res = 0 👍 
- this looks good 👍

PR #1860, opened 2020-08-27
[PAR-WDFT-248] Bey Bouziane

Bouziane95:


N10T:
Hey Bouziane,

**Iteration 1**
-
- 👍 


**Iteration 2**
-
- LINE 8-9: why looping in an array of 1 element? you don't need the for loop
- 👍 


**Iteration 3**
-
- LINE 27 - 33: 2 conditions for the same action, use a "or" operator ||
- reduce 👍 


**Iteration 4**
- 

### averageNumbers
- LINE 46 - 60: 2 conditions for the same action, yon don't need the line 46 else if statement
- 👍 
### averageWordLength
- 👍 


**Iteration 5**
-
- [...new Set(array)] 👍  
- my proposal : 
function uniquifyArray(array){
  if (array.length == 0) return null
  return [...new Set(array)];
} 

**Iteration 6**
-
- a shorter way : 
function doesWordExist(array, searching){
  if (array.length == 0) return null;
  return array.includes(searching)
}


**Iteration 7**
-
- 👍 


### good job

PR #1861, opened 2020-08-27
WDFT 248 / Philippe Buteau

s3chs:
Hey there! Sorry I did not push the code that much.
I will do it for every iteration by now, i promise!


Fanny-Sbn:
Hello Philippe, thanks for submitting your lab. 

**Iteration 2 & 3**
-
Good 

**Iteration 4**
-
**Level 1**
Your function is working, but in your for loop try to calculate only the sum and then return the sum/arr.length for performance 😉
**Level 2**
Same comment than level 1

**Iteration 5**
-
Good

**Iteration 6**
-
The method includes() check if a string exist in an array and return true or false, then you don’t need to do a for loop neither return true or false. 
  if (arr.length == 0) {
    return null;
  }
    return arr.includes(word)
}

For instance purpose, you can even write it in a line with ternary operator (it’s doing exactly the same thing that the one above) 

  return !arr.length ? null : arr.includes(word)
}

**Iteration 7**
-
Good, you just need to remove sumWord +=0, it’s useless. 
Good job! Try the bonus, if you have time 😉

PR #1862, opened 2020-08-27
[PAR-WDFT-248] Fabien Marchand

fabienMarchand:
Find my version of this lab


Fanny-Sbn:
Hello Fabien, thanks for submitting your lab. 

**Iteration 2**
-
Your else if and else statement are useless if you initialize your testWord variable as an empty string.

  var testWord = "";
  if (words.length == 0) {
    return null;
  }
  for (const word of words) {
    if (testWord.length < word.length) {
      testWord = word;
    }
  }
  return testWord;
}

**Iteration 3 – Bonus**
-
Perfect, but you don’t need to check if the type is Boolean, it’s doing automatically the conversion (false =0 and true =1). 

**Iteration 4**
-
Nice you reused the function sumNumbers(). 

**Iteration 5 &6&7**
-
Good

**Bonus**
-
The Jasmine tests are ok but you’re function greatestProduct() doesn’t work as expected. With your current for loops, you can’t check all the 4 adjacent numbers of your matrix…

You need to do this part of your code:
“calcH = matrix[i][j] * matrix[i][j + 1] * matrix[i][j + 2] * matrix[i][j + 3];" if the j < matrix.length -3
and this part of your code
"calcV = matrix[i][j] * matrix[i + 1][j] * matrix[i + 2][j] * matrix[i + 3][j];" if i < matrix.length -3


Good job !

fabienMarchand:
Thanks Fanny i will put the modification

PR #1863, opened 2020-08-27
[PAR-WDFT-248] Olivo Raphaëlle

RaphOL:
lab 3 exercices done


Fanny-Sbn:
Hello Raphaëlle, thanks for your lab 😊

**Iteration 2**
-
Good, but the your second if statement (if (words.length === 1) longestWord = words[0];) is useless.

**Iteration 4**
-
**Level 1**
Good 
Tip: you could have reused your sumNumbers() function to save a few lines and don’t repeat yourself. 
  if (numbers.length === 0) return null;
  return sumNumbers(numbers) / numbers.length
}

Good job!  Have you tried the bonus? 😉

Fanny-Sbn:
Hello Laura, thanks for your lab !

**Iteration 2**
-
You should use if, else if and else statement, that way you don’t need to write a condition after the else statement. 

In that case you just need and if and else statement :

  let word = "";
  if (arr.length === 0) {
    return null;
  }
for (let i=0; i<arr.length; i++) {
      if (word.length < arr[i].length)
      word = arr[i];
    }
  return word;
}

**Iteration 3**
-
You’re over complicating your function. The first else if statement is useless. You just need an if, else statement. In the else statement, you just need to do your for loop and add each elements of the array you’re iterating to your variable sum. In that case, no need to parseInt() the elements neither checking if the sum is equal to 0 or is greater than 0. You just have to return the sum. 

function sumNumbers (arr2) {
  let sum = 0;
  if(arr2.length === 0) {
    return 0;
  } else{
    for (let i=0; i<arr2.length; i++) {
      sum += arr2[i];
    }
      return sum;
  }
}


**Iteration 4**
-

**Level 1**

Good, the else if statement is useless. 

Tip: you could have reused your sumNumber() function to save a few lines and don’t repeat yourself. 
function averageNumbers(arr4) {
  if (arr4.length === 0) {
return null;
}
  return sumNumbers(arr4) / arr4.length
}

**Level 2**

Your first else if statement is useless and so your variables str and strLength. Your else take the case if there is only one element in the array. 

function averageWordLength(arr42){
  var sum42 = 0;
  if(arr42.length === 0) return null;
    for (let i=0; i<arr42.length; i++){
    sum42 += arr42[i].length;
    }
    return sum42/arr42.length;
  }

**Iteration 5**
-
Use an if else statement. You’re not using word5 and arrThree variables, so you can remove them. 
You don’t need to write the first element of the array when declaring the variable arrTwo, you can declare it as an empty array. 
function uniquifyArray(arr5) {
  let arrTwo = [];
  if (arr5.length === 0) return null;
  for (let i = 0; i < arr5.length; i++) {
    if (!arrTwo.includes(arr5[i])) {
      arrTwo.push(arr5[i]);
    };
  }
  return arrTwo;
}

**Iteration 6*
-
The first else if statement is useless. Here you should have an if, else statement. 
The method includes() check if a string exist in an array and return true or false, then you don’t need to do a for loop neither return true or false. 
function doesWordExist(arr6, word) {
  if (arr6.length === 0) {
    return null;
  }
    return arr6.includes(word)
}

**Iteration 7**
-
Your function is overcomplicated and doesn’t print the score (you need to return it). You just need to check if the word that you have passed as argument is equal to the element you’re iterating on in the array and if it is, add one to your variable score, all the other steps are unnecessary. You really don’t need to create a variable to push the elements in it. 


function howManyTimes(arr7, word7) {
  var score = 0;
  if(arr7.length === 0) return 0; 
    for (let i = 0; i<arr7.length; i++) {
      if (arr7[i] === word7) {
        score++;
      }
    }
   return score;
  }

You need to review a bit this lab. Wax on wax off! 😉

PR #1865, opened 2020-08-27
[PAR-WDFT-248] Ailie McCorkindale

RaspberryChaos:


N10T:
Hello AIlie,

**Iteration 1**
-
- 👍, but on short function like this try to use a fat arrow function  

  return a > b ? a : b;
}

const maxOfTwoNumbers2 = function (a,b) {
  return a > b ? a : b;
}

const maxOfTwoNumbers3 = (a,b) => a > b ? a : b;

**Iteration 2**
-
-  👍  


**Iteration 3**
-
- reduce 👍 
- LINE 48:
      case 'boolean':
        nums[i] === true ? total += 1 : total += 0;
you can do it like this : 
       case 'boolean':
         total += nums[i]
true is equal to one when you "+" it with a number and false is equal to 0, but your way is more explicit 👍 


**Iteration 4**
- 
- reduce 👍 


**Iteration 5**
-
-  👍 


**Iteration 6**
-
- good job with the -1 of indexOf 👍 
return array.indexOf(word) !== -1 ? true : false;
you dont need a ternary : 
return array.indexOf(word) !== -1
or you need it for this : 
const doesWordExist = (array, word) =>array.length ? array.indexOf(word) !== -1 : null;


**Iteration 7**
-
**2 conditions, same return**
  if(!array.length) return 0;
  if(array.indexOf(word) === -1) return 0;
**use the or operator :** 
  if(!array.length || array.indexOf(word) === -1) return 0;

🥇 **BONUS** 🥇 
-

- Nice JOB! 💯

PR #1866, opened 2020-08-27
[PAR-WDFT-248]  NICOLAS ENAULT

nenault:


N10T:
Hello Nico,

**Iteration 1**
-
- You used the longest way, but on short function like this try to use a fat arrow function  

function maxOfTwoNumbers(number1, number2) {
  if (number1 > number2) {
    return number1;
  } else if (number1 < number2) {
    return number2;
  } else {
    return number1;
  }
}

function maxOfTwoNumbers2(number1, number2) {
  if (number1 > number2) return number1;
  if (number1 < number2) return number2;
  return number1;
}

function maxOfTwoNumbers3(number1, number2) {
  if (number1 < number2) return number2;
  return number1;
}

const maxOfTwoNumbers4 = function (a,b) {
  return a > b ? a : b;
}

const maxOfTwoNumbers5 = (a,b) => a > b ? a : b;

**Iteration 2**
-
-  Be very careful with that kind of JS weirdness "arr == 0" 


**Iteration 3**
-
- keep in mind "==" is not equal to "===" 
- it's an extra step to use an array, but it's working too 👍 
- nice use of your already declared function sumNumbers


**Iteration 4**
- 
- nice use of your already declared function sumNumbers
- 👍 

**Iteration 5**
-
-  includes 👍 


**Iteration 6**
-
- LINE 133: No need loop, when you are not using your index
      if (arr.includes(word)) {
        return true;
      } else {
        return false;
      }
    }

My proposal:
  if (arr == 0) {
    return null;
  } else {
    return arr.includes(word)
  }
}

**Iteration 7**
-
- LINE 149: no need to return 0

🥇 **BONUS** 🥇 
-
- Nice JOB! 💯

PR #1867, opened 2020-08-27
[PAR-WDFT-248] Laurent Vuillaume

figura01:
done, not totaly working but mostly :p


N10T:
Hello Laurent,
-

**iteration 1**
- 👍 
- simpliest way : 
const maxOfTwoNumbers = (a, b) => {
  if (a > b) {
    return a;
  } else {
    return b;
  }
}

with a ternary : 
🥇  const maxOfTwoNumbers = (a, b) => a > b ? a : b;
 
**iteration 2**
-
- "==" and "===" are no equal, prefer to use "===" to compare the type and the value
- In findLongestWord function, you have a "res" and a "result" variable, can be tricky to understand
- my proposal:
  var longestLen = 0, longestWord = null;
  if (words.length === 0) return;

  for (let i = 0; i < words.length; i += 1) {
    if (words[i].length > longestLen) {
      longestWord = words[i];
      longestLen = longestWord.length;
    }
  }
  return longestWord;
}

 
**iteration 3**
- 
- the indentation of you else/if is hard to understand:
- this condition is always true because an array is truthy.
- my proposal with your function : 
const sumNumbers = (arr) => {
  let res;
  if (arr.length === 0) { res = 0 }
  if (arr.length === 1) { res = arr[0] }
  if (arr.length > 1) {
        let sum = arr[0];
        for (let i = 1; i < arr.length; i += 1) {
          sum += arr[i];
        }
        res = sum;
    }
  return res;
};

 
**iteration 4**
-
- why not using your function "sumNumbers" ?
- my proposal : 
  var sum = 0;

  if (!numbers.length) return null;
  if (numbers.length === 1) return numbers[0];

  for (let i = 0; i < numbers.length; i++) sum += numbers[i];

  return sum / numbers.length;
}
 
**iteration 5**
-
- Easier way : 
function uniquifyArray(words) {
  if (!words.length) return;

  const finalWords = [];

  for (let i = 0; i < words.length; i++) 
    if (finalWords.indexOf(words[i]) === -1) finalWords.push(words[i]);

  return finalWords;
}
 
**iteration 6**
-
- find is working 👍 
- My proposal with includes function : 
function doesWordExist(arr, elementToFind) {
  if (arr.length === 0) return false;
  return arr.includes(elementToFind);
}

 
**iteration 7**
-
- why not retuning "count"?   "return count;"
    if (count === 0) {
      return 0
    } else {
      if(count === 1) {
        return 1;
      } else {
        if(count === 5) {
          return 5;
        }
 
**Bonus 8 🥇 **
-
- good start but you are not answering the question, only to the Jasmine unit testing

Try to have understood how to use the if, without the else sometimes, and use the return without creating a lot of variables.

PR #1878, opened 2020-08-29
[PAR-WDFT-248] Timothée NICOLE

timothee-nicole:


Fanny-Sbn:
Hello Timothée, thanks for submitting your lab.

**Iteration 1**
-
No need to write else here

**Iteration 2**
-
👍 
Try to indent your code properly to make it easier to read.
Be careful == is not equal to ===.

**Iteration 3**
-
 👍 

Bonus: You have the good logic, but what about the boolean ? 

**Iteration 4**
-
**Level 1**
👍 
Just a tip : instead of writing (numbersAvg === false), prefer (!numbersAvg)

**Level 2**
👍 

**Bonus**
Same comment than for bonus - iteration 3

**Iteration 5**
-
Prefer to write : 

const uniquifyArray = (wordsUnique) => {
  let newTab = [];
  if (wordsUnique.length == 0) {return null} 
  else {
      for (let i = 0 ; i < wordsUnique.length ; i++) {
           if (!newTab.includes(wordsUnique[i])){
                newTab.push(wordsUnique[i])}
        } 
    }
  return newTab
  }

**Iteration 6**
-

The method includes() checks if a string exist in an array and return true or false, then you don’t need to do a for loop neither return true or false.
if (arr.length == 0) {
return null;
}
return arr.includes(word)
}

**Iteration 7**
-
No need to write else continue

**Iteration 8**
-
You're trying to check the jasmine tests but the function doesn't do what it supposed to do ! Re-read the instructions from the PDF.

Good job !

PR #1881, opened 2020-08-31
[PAR-WDFT-248]Tatiana Matzenbacher

tatimatz:


N10T:
Hello Tatiana,
 
**iteration 1**
-
- 👍 

**iteration 2**
-
- 👍 
- just have to add this condition before your loop
if (words.length === 0) { return null }

**iteration 3**
-
- 👍 

**iteration 4**
-
- have to add a condition when you have only one element in the array
- you can re-use the function "sumNumbers"

**iteration 5**
-
- uniquifyArray is not working for Jasmine
- you give no conditions to return null
- try to finish this iteration if you have time

**iteration 6**
-
- one condition to add

**iteration 7**
-
- try to finish this iteration if you have time

Good start, you just have to finish your functions before the pull request

PR #1882, opened 2020-08-31
[PAR-WDFT-248] LAURA MANTAL-D'ABOVILLE

Lauradabo:


N10T:
Hello Laura,

 
iteration 1
-
- 👍 
- with a fat arrow function and ternary : 
const maxOfTwoNumbers = (nb1, nb2) => nb1 > nb2 ? nb1 : nb2;

**iteration 2**
-
- 👍 
- be careful "==" is not equal to "==="

**iteration 3**
-
- 👍 

**iteration 4**
-
- 👍 

**iteration 5**
- 👍 

**iteration 6**
-
- you don't need a loop, you are not using "i"
- my proposal
function doesWordExist(words, word) {
  if (words.length == 0) return null;
  if (words.length == 1) return true;
  return words.includes(word)
}

**iteration 7**
-
- 👍 

**Bonus 8 🥇 **
-
- very good job 🥇

PR #1886, opened 2020-09-15
[rem2009] lovro hodak

LovroHodak:


IH-WebDev-TA-Remote:
Very good job! all mandatory iterations passing except the last test of iteration 6. The problem is inside the loop. on the first iteration you are either returning false or true, but either way it is stopping the function to continue running. So it will always check only the first item, even if it doesn't match. Solution would be returning true if it encounters the word, but running the "return false" code after the loop finishes.
Still, very good job overall! and code is clean and easy to read. Very well done!

PR #1887, opened 2020-09-15
[REM 2009] Ross Homaidan

ibirossi:
I found this really difficult.  I know I should have been able to do most of it with what we have learnt so far, but I had real problems applying it correctly.  Sorry!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,50 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(a,b){
+  if(a>b)
+  return a;
+    else{
+      return b
+    }
+  }
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
I see overall you have good enough understanding of JS and probably just applying it in this specific scenario with the lab instructions was challenging. 
For example:
-  your code of iteration 2 is great. but as per instructions, you need to wrap everything in a function called findLongestWord with a parameter called words, finally return the longestWord variable. 
- For iteration 3, you forgot to add the parameter numbers to the function. Functions have access to outside variables, however, in this case it is expected that the numbers array will be passed via parameter.
- on iteration 5, just naming issues, you are returning the function uniquifyArray, not the array uniqueArray. Also inside the condition of indexOf, you wanna check that against the array parameter, not the uniqueArray you created.

Some of the other issues are a little similar. I would recommend you to go over the lab again when you have a chance, even checking against the solutions we will send ask us for help if you don't understand anything! And most important, remember to ask questions in the channel! putting those examples in the channel would have been great for everyone to jump in and help you :) Keep up the good work!

PR #1888, opened 2020-09-15
[REM - 2009]Marta Obregon

MartaObregon:


IH-WebDev-TA-Remote:
the approach for iteration 3 is perfect, however, there are some logic errors in the other ones that are preventing the tests from passing. for example:
- in iteration 2 the loop conditions inside the parenthesis are a little odd, try comparing them to the conditions in the iteration 3 loop. Also, because of the condition to return null if the array is empty is inside the loop it will never be executed on empty array. the loop doesn't run is the array doesn't have any elements.
- on iteration 4. Very good approach! however, you want to sum the numbers inside the loop and then divide by the total to get the average after the loop ends, just before returning the value. Just a math issue, not a coding one ;) 

Some of the other issues are similar, I would suggest you come back to the lab, even checking against the solution we will send later to see how that code logic goes against your approach. Also, don't be afraid to post more questions in the as-the-class channels, these would have been great for posting them there and have everyone help out :)

Still, I see some good approaches in there and a lot of effort, keep it up and you will get these in no time!

PR #1889, opened 2020-09-15
[REM-2009] Santiago Andrés Ardiles Páez

SantAndresP:
I couldn't find a way to remove duplicates using only **_for_** loops.


IH-WebDev-TA-Remote:
Awesome! all mandatory done and even some bonuses. Really good job! Even the comments for each iteration done and superb.
There is probably a way to remove the elements from the same array using for loops and splice, however, creating a new array without the repeated elements is a perfectly done approach.
Again, logic is very good and code is clean and easy to read. Keep up the amazing work!

PR #1890, opened 2020-09-15
[REM-2009] Sumiya Ushirodone

sumi0820:
Although I managed to outcomes, I was not able to clear some criteria. It would be appreciated if you would review it and give me some advice in terms of how my codes look or how we should approach to solve these kinds of coding challenges.

>Unique array
>should return the correct uniqified array when an array of the same elements passed as argument

>Find elements
>should return true if the word we are looking for is in the array

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,560 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Amazing work! all mandatory iterations done and even some bonus. Regarding your questions:

- iteration 5. The test that is not passing is in case you have one repeated element more than 2 times in the same array, with your code it will only removed case of that element, but it is expected to remove **all** cases except one
- iteration 6 inside the loop you are either return true or false. It is important to understand that once a function encounters the keyword returns, it will end, meaning it won't run any more code after that. Because of this, in this case it will only check the first element of the array. To solve this, only return true if it encounters the word inside the loop and move the "return false" line of code after the full loop finishes running.

But overall very good job, logic is very good, code is clean and easy to read. Keep it up!

PR #1891, opened 2020-09-15
[REM - 2009]Aleksandra Tanasijenko

alekstana:
couldn't finish the "Unique Array"

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,68 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(a,b){
+  if (a > b) { 
+    return a
+  } else {
+      return b
+    }
+}
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Good work on the lab. There seem to be some tests breaking but mostly because of small issues:

- iteration 3 and 4. You need to return the values instead of using console.log. Remember the console is used by developers to check if things are running correctly and to check for errors. However, in web development the values in functions are expected to be passed by using the return keyword. Change those and you will see more tests passing!
- good start on iteration 5. However, the uniqueArray was initialised as a string, but because you know it is gonna be an array it makes more sense to initialise it as an empty array. Also inside the loo good thinking using the indexOf to check if an element is inside an array, however there was the other part of the code that updated the array missing.
- iteration 6. Remember false and true (as booleans) are different types of data than strings, so they should never be written with quotes around them.

Overall good work, there are some iterations fully done and some tests passing. The code seems a little more clean and wasy to read but still try to watch indentation and spacing! I would recommend you go back to this, even with the solution and try to understand the approach of the mandatory iterations.

Keep up the good work!

PR #1892, opened 2020-09-16
[REM-2009] Jordane Ruiz

jordaneruiz:
Second Lab 80% completed.


IH-WebDev-TA-Remote:
Very good job on the lab!, almost all of the tests passing. There are a couple of issues causing the last tests of iteration 5 and 6 to fail.
- iteration 5. the code to push the element in line 156 is not inside the condition of line 155. So it is always running.
- iteration 6. inside the loop you are either returning true or false in each iteration. Whenever a function encounters the keyword return it will return that value and stop executing any other code inside. Meaning, for all cases, this loop will only run once. To solve this you want to return true inside the loop only if you encounter the word. And then, below the loop (so after the whole loop runs) try returning false. This means that the whole loop will run and try to look for the word, but if it doesn't, only after, it will return false.

But again, a couple small issues, the approach to all iterations was good and the code is clean and easy to understand!. Very good job!

PR #1894, opened 2020-09-16
[rem-2009]-Katie Stamey

kmstamey:


IH-WebDev-TA-Remote:
Good work. There seem to be a couple of issues though that are preventing all tests from each iteration from passing though.
- Iteration 2. the loop should be with the < symbol, not <= symbol. Also, remember to declare the variable i, as in "let i...".
- iteration 3. in the for loop condition you are checking if i<number, you are missing the ".length". Also, in L37 since the element will be a number, it should be initialised as a 0 value, not as an empty string.
- iteration 6. the function needs to have two parameters, one is the words array and the other the word to check.

On the other iterations the start of the approach is ok but at some point the logic is breaking. I would suggest to go back to this lab, even with the solution provided to see how your approach was working against the one shared.

Still, it was a very good effort!, and code is well indented and easy to read. Keep up the good work!

PR #1896, opened 2020-09-16
[REM-2009]-zuzana

ZuzanaAK:
This time I haven't done the bonus but I will look at it later

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,55 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(a,b) {
+  if (a>b) {
+    return a
+  } else {
+    return b
+  }
+}
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Code logic is good, but there are some tests that are not passing mainly because of some issues with how to work around Jasmine. for example:
- First, the code seems to be breaking Jasmine because of the code outside the functions. For this lab there is no need to write any code outside the functions created. Try commenting those and the tests will appear again.
- in iteration 2, the function has access to the words variable outside, but the idea is to work around a parameter been passed in the function. so add the parameter words to the function and the tests will pass.
- the functions in iterations 3 and 4 have the same name. This will cause some issues with Jasmine and is overall not good practice.
- iteration 4. once you change the name of the function to the correct one it should work
- iteration 4. on averageWordLength: add the parameter words, also you forgot to declare the variable total before the loop.
- iteration 6. the parameters should be in the same order: the words array, and the word to check. so (words, word)

There are a couple other issues but they are mostly with code logic, I would recommend you to  go back to this lab and check it against the solution provided. Try to see what might be missing in your approach, but still if you need any help, please let us know! Keep it up :)

PR #1949, opened 2020-10-13
[FTAMS102020] Nina van den Braak

ninsajala:
About half way the lab..

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -27,9 +27,7 @@ function findLongestWord(array) {
   for (let i = 0; i < array.length; i++) {
     if (longestWord.length < array[i].length) {
       longestWord = array[i];
-    } else {
-      continue;
-    }
+    } 
   }
   return longestWord;
 }
@@ -67,12 +65,8 @@ function sum(array) {
       sum = sum + array[i];
     } else if (typeof array[i] === "string") {
       sum = sum + array[i].length;
-    } else if (typeof array[i] === "boolean") {
-      if (array[i] === true) {
-        sum++;
-      } else {
-        continue;
-      }
+    } else if (typeof array[i] === "boolean" && array[i] === true) {
+      sum++ 
     }
   }
   return sum;
@@ -86,9 +80,8 @@ function averageNumbers(array) {
   if (!array.length) {
     return null;
   }
-  let sum = sumNumbers(array);
-  let average = sum / array.length;
-  return average;
+  
+  return sumNumbers(array)/ array.length;
 }
 
 // Level 2: Array of strings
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,597 @@
...
+function findLongestWord(array) {
+  let longestWord = "";
+  if (!array.length) {
+    return null;
+  }
+
+  for (let i = 0; i < array.length; i++) {
+    if (longestWord.length < array[i].length) {
+      longestWord = array[i];
+    } else {
```
LloydC, line 30: Good job Nina! You don't need the `else{...}` here because there's no action defined in its scope and the function will keep working fine with just the "if" statement

src/functions-and-arrays.js
```diff
@@ -1,73 +1,597 @@
...
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(array) {
+  let sum = 0;
+
+  for (let i = 0; i < array.length; i++) {
+    if (typeof array[i] != "number") {
```
LloydC, line 45: You could have an alternative conditional `if(typeof array[i] === "number"){sum += array[i]}` then if you wanted to calculate the sum of strings as well you could add `else if(typeof array[i] === "string"){sum += array[i].length}`

src/functions-and-arrays.js
```diff
@@ -1,73 +1,597 @@
...
+        typeof array[i] === "string" ||
+        typeof array[i] === "boolean"
+      )
+    ) {
+      throw new Error("Unsupported data type sir or ma'am");
+    } else if (typeof array[i] === "number") {
+      sum = sum + array[i];
+    } else if (typeof array[i] === "string") {
+      sum = sum + array[i].length;
+    } else if (typeof array[i] === "boolean") {
```
LloydC, line 70: A cleaner way to write line 70 would be `if (typeof array[i] === "boolean" && array[i] === true) {...}`

src/functions-and-arrays.js
```diff
@@ -1,73 +1,597 @@
...
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(array) {
+  if (!array.length) {
+    return null;
+  }
+  let sum = sumNumbers(array);
+  let average = sum / array.length;
+  return average;
```
LloydC, line 91: I feel it would also work if you had `return sumNumbers(array)/array.length`, you usually want to use variables when you plan to pass a value to another process/construct in your code but if that's not the case you can just return it

src/functions-and-arrays.js
```diff
@@ -1,73 +1,597 @@
...
+  "machine",
+  "subset",
+  "trouble",
+  "starting",
+  "matter",
+  "eating",
+  "truth",
+  "disobedience",
+];
+
```
LloydC, line 167: An alternative solution is ` function doesWordExist(arr, word){
  for (let el of arr){
     if (el === word) return true;
   }
  return false; }`

ninsajala: Thanks for the feedback! Your suggestions made my code a lot cleaner

PR #1950, opened 2020-10-13
[FTAMS102020] Sophie Tichelaar

Sophietic:
til 4.2

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -12,6 +12,7 @@ function maxOfTwoNumbers(num1,num2){
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
 function findLongestWord(words){
+  if (!words.length) return null;
 
   let longestWord = words[0]
   for(let i=0; i<words.length; i++){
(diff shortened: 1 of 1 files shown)


Sophietic:
Mandatory iterations done

src/functions-and-arrays.js
```diff
@@ -1,19 +1,69 @@
...
+findLongestWord(words)
+
+// Iteration #3: Calculate the sum 
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+function sumNumbers(array){
+  let sum =0;
+  for (let i=0; i<array.length; i++){
+    sum = sum + array[i]
```
LloydC, line 35: Nice! Another way to write it would be `sum += array[i]`

src/functions-and-arrays.js
```diff
@@ -47,6 +134,17 @@ const wordsCount = [
...
+  let count =0
+  for(let i=0; i<array.length; i++){
+if (array[i]===word){
+  count++
+}
+
+  }
+  return count
+}
+
```
LloydC, line 147: Nicely done Sophie!

An alternative solution is 
`function howManyTimes(arr, word) {

  let count = 0;
  for (let el of arr) {
    if (el === word) count++;
  }
  return count;
}`

src/functions-and-arrays.js
```diff
@@ -29,9 +79,46 @@ const wordsUnique = [
   'bring'
 ];
```
LloydC, line 81: An alternative solution is `
const uniquifyArray = words => {
  let uniqueArr = [];
  for (let word of words) {
    if (!uniqueArr.includes(word)) uniqueArr.push(word);
  }
  return uniqueArr;
};`

PR #1951, opened 2020-10-13
[FTAMS102020] Elia Chiarucci

eliachiarucci:


src/functions-and-arrays.js
```diff
@@ -28,9 +109,38 @@ const wordsUnique = [
...
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
+const doesWordExist = (wordsFind, word) => {
+  if (!wordsFind.length) {return null;}
+  let found = false;
+  for (let i = 0; i < wordsFind.length; i++) {
+    if (wordsFind[i] === word) {
+      found = true;
+    }
+  }
+  if (found === true) { return true; } else { return false; }
```
LloydC, line 141: Good work Elia! I think if you have on line 141 `return found` it would be simpler

PR #1952, opened 2020-10-13
[FTAMS102020] Emma Northridge

emmanorthridge:
Complete

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,12 +1,42 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 > num2){
+    return num1
+  }else if(num2 > num1){
+    return num2
+  }else{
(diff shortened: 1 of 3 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,73 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 > num2){
+    return num1
+  }else if(num2 > num1){
+    return num2
+  }else{
```
wax-bear, line 7: Is this "else" clause necessary?

wax-bear (commented):
Great stuff Emma!

A few comments from my side:
- Please mind indentation, spacing and consistent use of semicolons.
- It is always helpful to see how other programmers solved the problems you are working on yourself. Perhaps google problems like "finding the highest number" and check Stackoverflow to see some alternative solutions.

PR #1953, opened 2020-10-13
[WDFT-AMS-102020] Yann Quintin 

YannQuintin:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,95 @@
...
+
+// maxOfTwoNumbers(3, 4);
+
+
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-// Iteration #3: Calculate the sum
+let longestWord;
```
LloydC, line 27: `let longestWord` should be defined inside function findLongestWord as it only needs to be used within the scope of that function

PR #1955, opened 2020-10-13
[FTAMS102020] Rich Alossi

alossix:


src/functions-and-arrays.js
```diff
@@ -1,73 +1,522 @@
...
+const words = [
+  "mystery",
+  "brother",
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot",
+];
+let longestWord;
```
wax-bear, line 20: I would place this let in the function body. Imagine you export the function to another file, and it is called there, are you sure "longestWord" is defined?

src/functions-and-arrays.js
```diff
@@ -1,73 +1,522 @@
...
+let sum = (arr) => {
+  if (arr.length === 0) {
+    return 0;
+  } else {
+    for (let i = 0; i < arr.length; i++) {
+      if (typeof arr[i] === "string") {
+        arr[i] = arr[i].length;
+      } else if (arr[i] === true) {
+        arr[i] = 1;
+      } else if (arr[i] === false) {
```
wax-bear, line 61: Is this condition necessary?

src/functions-and-arrays.js
```diff
@@ -1,73 +1,522 @@
...
+        arr[i] = 0;
+      } else if (typeof arr[i] === "object" || typeof arr[i] === "array") {
+        return "Error: Objects and arrays are not valid input.";
+        // I think 'throwing' errors is a JS function we haven't learned yet
```
wax-bear, line 65: Correct. You could replace "return" by "throw".

wax-bear (commented):
Great job Rich!

A few comments:
- Generally your code style looks good.
- Most of the tests pass, which is great! I would encourage you to look at solutions on Stackoverflow, in order to see different ways of going about these challenges.

PR #1960, opened 2020-10-13
[FTAMS102020]issaGruber

GruberI:


wax-bear (commented):
Hey Issa,

Great job coding functions and arrays!

A few general remarks:
- Please mind spacing, indentation and consistent use of semicolons.
- The functions coded work, and that is great! A next step could be to reflect on the current solutions, and then research other implementations. For example, uniquifyArray can be solved in multifold ways. https://stackoverflow.com/questions/9229645/remove-duplicate-values-from-js-array

GruberI:
Thank you!

PR #1970, opened 2020-10-14
[FTAMS102020]Mahalakshmi-Maya-Emad

mahavinay:


src/functions-and-arrays.js
```diff
@@ -1,18 +1,109 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers(number1, number2){
+  return Math.max(number1, number2);
```
wax-bear, line 4: Elegant!

src/functions-and-arrays.js
```diff
@@ -1,18 +1,109 @@
...
+}
+
+
```
wax-bear, line 7: What is the use of these blank lines?

src/functions-and-arrays.js
```diff
@@ -1,18 +1,109 @@
...
 // Iteration #2: Find longest word
+
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+function findLongestWord(words) {
+  let word ='';
+  if (!words.length){
+    return null;
+  } else for (let i = 0; i < words.length; i++) {
```
wax-bear, line 16: I would wrap the for loop by the curly braces (not used right now) of the else clause.

src/functions-and-arrays.js
```diff
@@ -1,18 +1,109 @@
...
+    if (!mixedArr.length){
+      return 0;
+    }
+    else{
+    for (let i = 0 ; i < mixedArr.length ; i++){
+      if (typeof mixedArr[i] === "number"){
+        console.log(sum)
+      sum += mixedArr[i]
+      //console.log("number",sum)
+      } else if ((typeof mixedArr[i] === "boolean") && (mixedArr[i] = "true")) {
```
wax-bear, line 52: I see two issues on this line. Firstly, I think you'd like to compare "===" mixedArr[i] with "true" rather than asigning "true" to mixedArr[i]. Secondly, you want to see whether mixArr[i] is _true_. true is a boolean, whereas "true" is a string.

src/functions-and-arrays.js
```diff
@@ -47,6 +171,25 @@ const wordsCount = [
...
 
+
+function howManyTimes (wordsCount, searchWord)  {
+  count = 0;
+  if(!wordsCount.length){
+    return 0
+  }
+  words.forEach(item => {
+    if (item == searchWord) 
+      count+=1;
```
wax-bear, line 182: You could use "count++" here.

src/functions-and-arrays.js
```diff
@@ -47,6 +171,25 @@ const wordsCount = [
   'matter'
 ];
 
+
+function howManyTimes (wordsCount, searchWord)  {
+  count = 0;
+  if(!wordsCount.length){
+    return 0
+  }
+  words.forEach(item => {
```
wax-bear, line 180: Where is words coming from? Would it perhaps be better to submit the wordsArray as an argument?

src/functions-and-arrays.js
```diff
@@ -47,6 +171,25 @@ const wordsCount = [
   'matter'
 ];
 
+
+function howManyTimes (wordsCount, searchWord)  {
```
wax-bear, line 175: I would add/replace an argument. The words array is not included as an argument right now. Plus, do we need the wordsCount as an argument?

src/functions-and-arrays.js
```diff
@@ -47,6 +171,25 @@ const wordsCount = [
...
+  count = 0;
+  if(!wordsCount.length){
+    return 0
+  }
+  words.forEach(item => {
+    if (item == searchWord) 
+      count+=1;
+      else
+      count = 0
```
wax-bear, line 184: Imagine the program is iterating over the words array. It finds an instance of the word you are searching for. Now, the count is 1. The next word is not the word searched for. Now, the count is set back to 0. Is this what you would like to happen?

wax-bear (commented):
Great stuff Mahalakshmi!

A few comments from my side:
- Almost al non-bonus functionalities work, which is great! It would be great if all cases in "howManyTimes" would pass as well.
-  Please mind indentation, spacing and consistent use of semicolons.

PR #1972, opened 2020-10-14
[FT-AMS-102020] Emad Ghnim

Emad-doz:


src/functions-and-arrays.js
```diff
@@ -1,18 +1,112 @@
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+function averageNumbers (arr){
+  let sum = 0;
+  if (!arr.length){
+    return null;
+  }
+  for (let i = 0 ; i < arr.length ; i++){
+    sum += arr[i]
+  }
+  return sum / arr.length
```
LloydC, line 65: You could also have used your sumNumbers function `return sumNumbers(arr)/arr.length`

PR #1974, opened 2020-10-14
[FTAMS102020] AlejandroPliakasBenitez

apliakas:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,148 @@
...
+const averageNumbers = arrayOfNumbers => {
+  let total = 0;
+
+  for(let i = 0; i < arrayOfNumbers.length; i++) {
+    total += arrayOfNumbers[i];
+  }
+
+  return Math.round((total / arrayOfNumbers.length) * 10) / 10;
+}
+
```
LloydC, line 95: You could shorten your code if you made use of your `sumNumbers` function

`return Math.round(sumNumbers(arrayOfNumbers)/arrayOfNumbers.length * 10)/10`

PR #1977, opened 2020-10-15
[FTAMS102020] Sabela Pena Garcia

defSabe:
Thanks!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -47,37 +47,30 @@ function sumNumbers(arr) {
 // Level 1: Array of numbers
 //const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers(arr) {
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,76 @@
...
 
 // Iteration #3: Calculate the sum
 
-const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+//const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+ 
+function sumNumbers(arr) {
+  let sum = 0;
+  for (let i = 0; i < arr.length; i++){
+      sum = arr[i] + sum;
```
LloydC, line 41: Good work Sabela! 

You can also write  use `sum += arr[i]` instead of `sum = arr[i] + sum`

PR #1979, opened 2020-10-16
[FTAMS102020] CEYDAULUBAS-lab-javascript-functions-and-arrays

ceydaulubas:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,65 @@
...
+    return number1
+  } else {
+    return number2;
+  }
+}
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+function findLongestWord(words) {
+  let longestWord = '';
+  if (words.length == '') {
```
LloydC, line 13: Your condition on line 13 is incorrect because the length of an array is always a number. What you have in your compare right now is an empty string, it should be

`if(words.length === 0) {...}`

src/functions-and-arrays.js
```diff
@@ -1,19 +1,65 @@
...
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(arr) {
+  if (arr.length === 0) {
+    return null;
+  }
+  return sumNumbers(arr) / arr.lengt;
```
LloydC, line 46: You're missing a "h" character at the end of line 46, in the future always read proof your code or run it before you submit.

src/functions-and-arrays.js
```diff
@@ -29,9 +75,35 @@ const wordsUnique = [
...
+}
+
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
+function doesWordExist(wordExist, word) {
+  if (wordExist.length == 0) {
+    return null
+  }
+  for (let i = 0; i < wordExist.lengt; i++) {
```
LloydC, line 98: Same here, you're missing character "h" on `wordList.length`

src/functions-and-arrays.js
ceydaulubas: When I ran my code I didn't get any errors. Therefore, I wasn't aware that I forgot to add "h". Any suggestions for the plugin?
By the way, I will share my code with you again after I fix my code.

PR #1980, opened 2020-10-16
[AMS-FT-102020] Ahmet Demirlicakmak

codebrokerad:
All iterations are completed, but please urge me if there are shorter and more feasible solutions for each iterations.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,51 +1,115 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers (h1,h2){
+  if (h1 > h2) { 
+  highestNum = h1;
+} else if (h1 == h2){
+  highestNum = h1;
+} else {
+  highestNum = h2;
+}
+return highestNum;
+}
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,51 +1,115 @@
...
+    return null;
+  }
+  for (let i in array){
+    if(array[i] == word){
+      return true;
+    }
+  }
+ return false;
+}
+
```
LloydC, line 96: Good job Ahmet!

Here's another solution to give you a different perspective
`function doesWordExist(arr, word) {
  return arr.includes(word);
}`

src/functions-and-arrays.js
```diff
@@ -1,51 +1,115 @@
...
-];
+function uniquifyArray(array){
+  if (!Array.isArray(array) || !array.length) {
+    return null;
+  }
+  let uniqueArray = [];
+  for(let i in array){
+    
+      if(uniqueArray.indexOf(array[i]) === -1) {
+
```
LloydC, line 76: You can also use this conditional `if (!uniqueArray.includes(array[i])){...}`

PR #1982, opened 2020-10-16
[FTAMS102020] Alex Kantsadis

teof95:
iteration 7 does not fully work. dont know why!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,21 +1,21 @@
 // Iteration #1: Find the maximum
 function maxOfTwoNumbers(number1, number2) {
-  if (number1 > number2) {
-    return number1;
-  } else {
-    return number2;
-  }
+    if (number1 > number2) {
+        return number1;
+    } else {
+        return number2;
+    }
 }
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
 function findLongestWord(words) {
-  if (words.length > 0) {
-    return words.sort(function (a, b) { return b.length - a.length })[0];
-  } else {
-    return null
-  }
+    if (words.length > 0) {
+        return words.sort(function(a, b) { return b.length - a.length })[0];
+    } else {
+        return null
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,175 @@
...
+    } else {
+        let total = 0;
+        for (let i = 0; i < numbersAvg.length; i++) {
+            total += numbersAvg[i];
+        }
+        let avg = total / numbersAvg.length;
+        return avg
+    }
+}
+
```
LloydC, line 56: You can replace line 48-53 with `return sumNumbers(numbersAvg)/numbersAvg.length`

src/functions-and-arrays.js
```diff
@@ -1,73 +1,175 @@
...
+    'sharp',
+    'playground',
+    'poison',
+    'communion',
+    'simple',
+    'bring'
 ];
 
+function uniquifyArray(data) {
+    if (wordsUnique.length == 0) {
```
LloydC, line 101: It should be `if (data.length == 0) {...}` as you want your function to work for any array

src/functions-and-arrays.js
```diff
@@ -1,73 +1,175 @@
...
+
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
+function doesWordExist() {
+    let includesTwenty = wordsFind.includes("machine");
+    return includesTwenty
+    console.log(includesTwenty)
+}
+
```
LloydC, line 120: Your function doesWordExist needs to be built with arguments, that way it will work for any array you give the function

`function doesWordExist(array, word) {
    let includesTwenty = array.includes(word);
    return includesTwenty
}`
`doesWordExist(wordsFind, "machine")`

src/functions-and-arrays.js
```diff
@@ -1,73 +1,175 @@
...
+    else
+        for (let i = 0; i < wordsCount.length; i++) {
+            let result = 0;
+            if (wordsCount[i] === "matter") {
+                result++;
+            }
+            return result;
+        }
+}
+
```
LloydC, line 150: Same thing with function howManyTimes

`function howManyTimes(arr, word) {
  if (arr.length === 0 || arr == [])
        return "0"
    else
            let result = 0;
            if (arr[i] === word) {
                result++;
            }
            return result;
        }
}`

`howManyTimes(wordsCount, "matter")`

PR #1984, opened 2020-10-20
[WDFTRM-20201020] - Pauline Massé

paulinemss:


src/functions-and-arrays.js
```diff
@@ -1,20 +1,113 @@
...
+  if (!arr.length) { return 0; }
+  let sum = 0;
+  arr.forEach((element) => {
+    if (typeof element === "number") {
+      sum += element; 
+    } else if (typeof element === "string") {
+      sum += element.length; 
+    } else if (element === true) {
+      sum += 1; 
+    } else if (typeof element === "object" || typeof element === "array") {
```
filipe-freire, line 55: In this line, you should only check for the first statement, since an array is just a "special" object! For more information you can check https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures

PR #1992, opened 2020-10-20
[BARCELONA FULL-TIME 2020] ALVARO PEREZ GOMEZ & GE YANG

ohhyangyang:


src/functions-and-arrays.js
```diff
@@ -29,8 +77,34 @@ const wordsUnique = [
...
+  }
+  return arrFiltered;
+};
+
+console.log(uniquifyArray(wordsUnique));
+
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
+var doesWordExist = function (arr, word) {
+  for (let i = 0; i < arr.length; i++) {
```
MartaJank, line 95: There's one piece of test you are not passing in jasmine. Maybe doing something like this you can go around it, although it may not be the best or cleanest way to do it:

var doesWordExist = function (arr, word) {
    if (arr[i] === word) {
      return true;
    } 
  }
  return false
};

You could also use the .includes( ) method.

MartaJank (commented):
Overall, very good job guys! 

Only two things to improve:
- Check with Jasmine to pass all the tests.
- Use more descriptive variable names so anyone, even if they have no idea of what is this code about, could understand it in a second just by having a quick look at it.

PR #1993, opened 2020-10-20
[WDFT-BCN-102020] Jaime Pinto & Lucia Skudlik

luciaskudlik:
done without last Bonus, and 2 Jasmin errors still

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,138 @@
 // Iteration #1: Find the maximum
+const maxOfTwoNumbers = (a,b) => {
+  if (a>b) {
+    return a;
+  } else if (a<b) {
+    return b;
+  } else {
+    return a;
+  }
+};
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+const findLongestWord = (array) => {
+  if (array.length === 0) {
+    return null;
+  }
+
+  let longestWord = array[0];
+  for (let i = 0; i < array.length; i++) {
+    
+    if (array[i].length > longestWord.length) {
+      longestWord = array[i];
+    }
+  };
+  return longestWord;
+};
+
+//console.log(findLongestWord(words));
+
 
 // Iteration #3: Calculate the sum
 
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
+const sumNumbers = (array) => {
+  if (array.length === 0) {
+    return 0;
+  }
+  
+  let sum = 0;
+
+  for (let i = 0; i < array.length; i++) {
+    sum += array[i];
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,139 @@
 // Iteration #1: Find the maximum
+const maxOfTwoNumbers = (a,b) => {
+  if (a>b) {
+    return a;
+  } else if (a<b) {
```
MartaJank, line 5: Maybe it would have been even easier to put just:

if (a > b) {
    return a;
} else {
    return b;
}

But good job 👍

src/functions-and-arrays.js
```diff
@@ -1,19 +1,139 @@
...
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+const findLongestWord = (array) => {
+  if (array.length === 0) {
+    return null;
+  }
+
+  let longestWord = array[0];
```
MartaJank, line 20: Very good on the variable naming, very clear to read and easy to understand :)

src/functions-and-arrays.js
```diff
@@ -1,19 +1,139 @@
...
+        total += array[i].length;
+        break;
+      case "boolean":
+        if (array[i] === true) {
+          total += 1;
+        } else {
+          continue;
+        };
+        break;
+      default:
```
MartaJank, line 82: To pass the test in Jasmine you couldn't pass you could just write something like this here instead of the "break;":

src/functions-and-arrays.js
```diff
@@ -29,9 +149,47 @@ const wordsUnique = [
...
 
+
+const uniquifyArray = (array) => {
+  if (array.length === 0) {
+    return null;
+  }
+
+  let uniqueArray = [];
+
+  for (let i = 0; i < array.length; i++){
```
MartaJank, line 160: Maybe instead of using the for loop you could use the forEach one, it will go around the test Jasmine is not passing:

array.forEach(function (word) {
    if (!uniqueArray.includes(word)) {
      uniqueArray.push(word)
    };
  })

src/functions-and-arrays.js
```diff
@@ -29,9 +149,47 @@ const wordsUnique = [
...
+  if (array.length === 0) {
+    return null;
+  }
+
+  for (let i = 0; i < array.length; i++) {
+    let currentWord = array[i];
+    let lastIndex = array.length -1;
+    if (currentWord === word ) {
+      return true;
+      break;
```
MartaJank, line 185: Try to avoid using "break", it is necessary in the switch loops but you shouldn't use it anywhere else (maybe in super advanced stuff sometimes, but in the bootcamp I am sure you won't need it anywhere but in the switch loops).

Actually, when you delete this break your code still works.

PR #1997, opened 2020-10-20
[WDFTRM-20201020] - Monique de Haard

moniquedehaard:
Finished exercises lab functions and arrays


src/functions-and-arrays.js
```diff
@@ -1,18 +1,102 @@
...
+function maxOfTwoNumbers(num1, num2){
+  if (num1 > num2) {
+    return num1;
+  }
+  if (num2 > num1) {
+    return num2;
+  }
+  if (num1 === num2) {
+    return num1;
+  }
```
filipe-freire, line 11: You could also use the following as an alternative! 
`if (num1 > num2) {
return num1;
} else {
return num2;
}`

filipe-freire (approved):

PR #1999, opened 2020-10-20
Juliane and Irene Group 12

juliane-29:


src/functions-and-arrays.js
```diff
@@ -29,9 +90,37 @@ const wordsUnique = [
   'bring'
 ];
 
+var emptyArray = [];
+
+function uniquifyArray(arr) {
+  if (arr.length === 0){
+    return null;
+  } else if (arr.length > 0) {
```
MartaJank, line 98: You can make it simpler with a forEach and it will then pass all the test:

arr.forEach(function (word) {
    if (!emptyArray.includes(word)) {
      emptyArray.push(word)
    };
  })

Also, maybe it would be better to name the variable "uniqueArray" or something like that because, once you push things into it it won't be empty anymore.

PR #2002, opened 2020-10-20
[BCN-FT-102020]Jamie Garrett Emma Craig

garrettjamie87:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,65 @@
...
+    return number1;
+  }
+}
+
+console.log(maxOfTwoNumbers(3,4));
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+function findLongestWord(array){
```
MartaJank, line 16: You should check if the array is empty, if it is, Jasmine asks you to return null. Something like this:

if (!array.length) return null;

or like this:

if (array.length === 0) {
    return null;
}

src/functions-and-arrays.js
```diff
@@ -1,19 +1,65 @@
...
+  var longestWord = 5;
+for (i=0; i<array.length; i++){
```
MartaJank, line 18: The logic was good but you were returning a number and the exercise asks you for a word, you can fix it like this:

var longestWord = "";
for (i=0; i<array.length; i++){
  if (array[i].length > longestWord.length) {
    longestWord = array[i];

  }
}
 return longestWord;

src/functions-and-arrays.js
```diff
@@ -1,19 +1,65 @@
...
+  }
+  return sum;
+}
+console.log(sumNumbers(numbers));
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(array) {
```
MartaJank, line 46: Here, Jasmine also asks you to return null if the array is empty so include this piece of code:

if (arr.length === 0) {
    return null;
  }

src/functions-and-arrays.js
```diff
@@ -1,19 +1,65 @@
...
+  return sumNumbers(array)/array.length;
+}
+
+console.log(averageNumbers(numbersAvg));
+
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
+
+function averageWordLength(array) {
```
MartaJank, line 56: Same here:

if (arr.length === 0) {
    return null;
  }

PR #2004, opened 2020-10-20
BCN-WEBOCT2020 - Hélène Stephan & Juanmi Sansinenea - Day02

juanmi-sansinenea:


src/functions-and-arrays.js
```diff
@@ -1,18 +1,80 @@
 // Iteration #1: Find the maximum
+var maxOfTwoNumbers = function(num1, num2){
+  if (num1 > num2){
+    return num1;
+  } else {
+    return num2;
+  }
+}
+
```
MartaJank, line 9: Good job! Clean and easy to understand! 👍

src/functions-and-arrays.js
```diff
@@ -1,18 +1,80 @@
...
+mixedArr = ["hi", 5, {fruit1: 'lemon', fruit2: 'orange'}, null];
+var avg =(array) =>{
+  if (array.length === 0) {
+    return null;
+  }
+  var totalLength = 0;
+  for (i = 0; i < array.length; i++) {
+    totalLength += array[i].length;
+  }
+}
```
MartaJank, line 77: To pass all the Jasmine test, you could do something more easy maybe, like this:

let total = sum(array);
let average = total / array.length;

and then return that value, maybe toFixed( ).

src/functions-and-arrays.js
```diff
@@ -29,8 +91,37 @@ const wordsUnique = [
   'bring'
 ];
 
+
+var uniquifyArray = function(array){
+  //var storedLength = array.length;
+  if (array.length === 0) {
+    return null;
+  }
+  for (i = 0; i < array.length; i++) {
```
MartaJank, line 100: You can actually do this with a forEach, it would be easier and it would pass all the tests:

array.forEach(function (word) {
    if (!uniqueArray.includes(word)) {
        uniqueArray.push(word)
    };
})

src/functions-and-arrays.js
```diff
@@ -29,8 +91,37 @@ const wordsUnique = [
...
+          array.splice(j,1);
+       }
+    }
+  }
+  return array;
+}
+
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
+var doesWordExist = (word, array) =>{
```
MartaJank, line 112: Jasmine throws two errors and this is only due to the order of the parameters you're passing.

Just put array first and word second and it will work :)

src/functions-and-arrays.js
```diff
@@ -29,8 +91,37 @@ const wordsUnique = [
...
+  if (array.length === 0) {
+    return null;
+  }
+  if (array.length === 1 && word === array[0]){
```
MartaJank, line 116: This if is actually not necessary, you can actually delete it and the code will still work.

MartaJank (commented):
Good job guys! Below some little comments :)

src/functions-and-arrays.js
juanmi-sansinenea: Thank you, Marta!

PR #2005, opened 2020-10-20
[BCN-WDFT-202010] Claudia Layrisse & Joe Buckley

clayrisse:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,144 @@
...
+// return sum
+// }
+function sumNumbers(numbers) {
+  if (numbers.length === 0) {
+    return 0;
+  }
+  var sum = () => numbers.reduce((acc, currentValue) => acc + currentValue);
+  return sum();
+}
+function sum(numbers){
```
MartaJank, line 62: You could make it with a switch for example:

switch (typeof numbers[i]) {
      case "number":
        total+= numbers[i];
        break;
      case "string":
        total += numbers[i].length;
        break;
      case "boolean":
        if (numbers[i] === true) {
          total += 1;
        } else {
          continue;
        };
        break;
      default:
    }
  }
  return total;

or with forEach:

  let sum = 0;
  numbers.forEach( function (el) {
    if (typeof el === 'string') {
      sum += el.length;
    }
    else if (typeof el === 'object') {
      throw new Error("Unsupported data type sir or ma'am")
    } 
    else {
      sum += el
    };
  })

  return sum;

src/functions-and-arrays.js
```diff
@@ -1,19 +1,144 @@
...
+  var sum = () => length.reduce((acc, currentValue) => acc + currentValue);
+  
+  return sum()/str.length;
+}
+
+function avg (str){
+  if (str.length === 0) {
+    return null;
+  }
+  var onlyNumbers = str.filter(x=>(typeof (x) === 'number'))
```
MartaJank, line 134: You can actually do this exercise just using sum( )

PR #2137, opened 2021-01-12
Hai Nguyen - All done

honghainguyen777:


annawaszak:
Hi Hai!

Excellent job on this lab! Your code is clean, readable and very efficient. 

I have only two improvement suggestions for **iteration 6**

```
function doesWordExist(words, wordNeedToSearchFor) {
  if (words.length === 0 || word.length === 0) { 
    return null;
  }
  let isExist = false;
  for (let word of words) {
    if (words.includes(wordNeedToSearchFor)) {
      return true;
    }
  }
  return false;
}
```

- In the first statement "word" is not defined (same bug in iteration 7)
- Since you used the includes method(), you did actually need the for loop around it. It returns 'true' in any of the words are a match. You would need that look you find chose to put this in the if statement (word === wordNeedToSearchFor).

Well done!
Anna

PR #2140, opened 2021-01-12
obligatory exercises

aniastryjewska:


annawaszak:
hi Ania!

Great job on the lab!

Everything works and good really good, here are just a few suggestions for minor improvements:

**iteration 1**
All good. If you wanted to make it even more efficient, you could have skipped the second else if, like below.

```
  if(a > b) {
    return a
  } else {
    return b
  }
}
```

**iteration 4.2**
Unconventional solution - clever!

**iteration 7**
It's perfectly fine, just wanted you to know that within the if statement, you don't need to  write "numberOfTimes = numberOfTimes+1",  " numberOfTimes += 1" is a shorter version.


I would encourage you to use white spaces and indentations to make your code more readable. 

Well done!
Anna

PR #2143, opened 2021-01-12
Update functions-and-arrays.js

princi8m:
Not finished yet.  Work in progress v2

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -81,6 +81,12 @@ const wordsUnique = [
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
+function doesWordExist(findWord, wordToFind){
+  let n = findWord.includes(wordToFind);
+    return n;
+  }
+
+
 
 // Iteration #7: Count repetition
 const wordsCount = [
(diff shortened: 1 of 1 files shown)


annawaszak:
Hi Michele!

Good job on the lab, it was a difficult one. Below you will find a few suggestions that I hope you'll find helpful.

**iteration 1**
Your code is correct, a more efficient version could be this:

```
function maxOfTwoNumbers (num1, num2) {
  if (num1 > num2) return num1;
  else return num2
};
```
Also, if your 'if statement' is a one liner, you don't need to use the curly brackets :)

**iteration 2**

- The first if statement is redundant (the third if statement will work if the array contains just one word).
- In the second if statement you can just return null. Don't need to save it as the longest word. 
- Return the 'longestWord' instead of console.logging it .

**iteration 4.2**

- At the very end of the function, you should return the 'totWordLength' instead of console.logging it. 
- I would advise to call it aveWordLength at that point to make the code more readable.
```
function averageWordLength (arr1) {
  console.log(arr1.length);
  let totWordLength=0;
  for (i=0; i < arr1.length; i++) {
    totWordLength=totWordLength+arr1[i].length;   
  }
  totWordLength = totWordLength / arr1.length;
   console.log(totWordLength);
}
```

Most of your iterations fail this test "should return null if receives an empty array when called". You have passed it in iteration two though, so you just need to add this bit of code to every iteration that requires it:
```
  if (words.length === 0) {
    return null;
  };
```

**iteration 6**
Very cool and efficient solution!

Thanks,
Anna

princi8m:
Thank you!!!

PR #2144, opened 2021-01-12
Pull Request Corentine Arrays 

LouisianeP:


annawaszak:
Hi Corentine!

Great job on the lab. It was a tricky was and your solutions very very efficient. Here are a few suggestions for minor improvements:

**iteration 2**
- Just FYI, the for loop does not have to be wrapped in the else statement. It could be independent. 

**iteration 3 - bonus**
It does not work - prints out "18miami1truebarca200lisboa810". It's because it cannot handle any type of data besides numbers. I would propose using multiple if statements with combination of 'typeof' operator. 

**iteration 4.1**
You can but don't need to first define the variable and then assign a new value to it. You could 
```
 let avgArray = 0
 let sumArray = sumNumbers(array)
 avgArray = sumArray / (array.length)
```
you can just do this:
```
 let sumArray = sumNumbers(array)
 let avgArray = sumArray / (array.length)
```
**iteration 5**
Unique and interesting solution!

Also, well done on that last bonus - it was a difficult one!


I have just one general note - proper indentations and use of whitespaces would can your code even more readable. You tend to keep everything within the function on the same level regardless of how it's nested. Here are to screenshot to illustrate the difference:
Screen Shot 2021-01-14 at 3 14 22 PM
Screen Shot 2021-01-14 at 3 14 47 PM

Hope this helps! I'm happy to discuss the details with you if you have any questions.

Thanks, 
Anna

LouisianeP:
@annawaszak 
Thanks a lot for your feedback ! Really much appreciated ! 
I will improve the readability for next labs :) promised.

Regarding the last bonus, I think I made something that worked, but I guess that it is not performant since I doubled the loops and that takes a lot of computation effort. 
I would be happy to see a more efficient solution if you have one.

Have a nice day,
Corentine

PR #2181, opened 2021-01-15
add code

NicoZanzan:


annawaszak:
Hi Nico,

Really amazing job on the lab. I wanted to highlight that just because Jan showed a different solutions to some of the iterations, it doesn't mean they were better - just different approach. 

Here are a few suggestion for minor changes to improve your code even more:

**iteration 5**
```
  for (let i = 0; i < ar.length; i++) {
      if (ar.forEach((word) => {
  if (!uniqueWords.includes(word)) {
      uniqueWords.push(word);
  }
```
You should pick either a for loop here OR forEach, but not both. If you're going with the forEach, you don't need to wrap it in an if statement. So it would be either:

```
  for (let i = 0; i < ar.length; i++) {
      if (!uniqueWords.includes(ar[i)) {
          uniqueWords.push(ar[i));
      }
  }
```
OR
```
  ar.forEach((word) => {
      if (!uniqueWords.includes(word)) {
          uniqueWords.push(word);
      }
  }
```

Here are a few hacks:
- if you had an if/else if/else statement that is a one liner, you can skip the curly brackets, like this:
```
   if (!uniqueWords.includes(word)) uniqueWords.push(word);
```

**iteration 6**
You did not have to wrap the if statement in a loop. Includes go through an array element and returns true or false if includes the element you're looking for.

```
for (i = 0; i < array.length ; i++) {
      if (array.includes(word)){
        return true
      } else {
        return false
      }
  }
```
Since includes, already returns 'true or false' you could save yourself some time doing this:
`  return arr.includes(searchedWord);
`

Hope this helps, If you have any questions or want to discuss something in more details, slack me :)

Thanks, 
Anna

PR #2188, opened 2021-01-16
done

deboranis:


ingridpitta:
Congratulations! Some observations :)

You could have used the "Early return", that is, check if your argument does not immediately satisfy a condition and already return. For example, check if the argument is not an empty array before proceeding to the other steps of the algorithm.

PR #2189, opened 2021-01-16
[WD-PT-SAO-01-2021] Lucas Franco - Henrique Morikawa

lhfdv:


ingridpitta:
Congratulations! Some observations :)

You could have used the "Early return", that is, check if your argument does not immediately satisfy a condition and already return. For example, check if the argument is not an empty array before proceeding to the other steps of the algorithm.

If you are not going to have a third condition, or more, within the "if" block you can end with "else", you do not need - nor should - end with "else if" (line 35)

In "function sum (anyArray)" you could have used a switch case to avoid the many "if's"

In "uniquifyArray" it is not necessary to use one for inside the other to resolve this issue. (https://stackoverflow.com/questions/1960473/get-all-unique-values-in-a-javascript-array-remove-duplicates)

Note: not every "if" block needs to have an "else"

Cool that you reused functions and did all the exercise :)

PR #2196, opened 2021-01-19
[WDPT JAN21] Aleksandra Porada: Finished all the iterations and bonuses

hola-ola:


src/functions-and-arrays.js
```diff
@@ -10,8 +10,8 @@ function maxOfTwoNumbers(num1, num2) {
 };
```
IH-WebDev-TA-Remote, line 10: Another good solution would be using the Math.max(num1,num2), you can learn more about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max).

src/functions-and-arrays.js
```diff
@@ -77,8 +81,8 @@ function averageNumbers(numbersAvg) {
...
```
IH-WebDev-TA-Remote, line 81: You could also have used the sum function previously declared instead of typing the sum logic once again. After all, the beauty of functions is that they can be reused! :smile:

IH-WebDev-TA-Remote (commented):
Very nice work Aleksandra! Congratulations on finishing yet another lab! :tada: I've added a few comments that can further improve your solution! Other than that, well done!

src/functions-and-arrays.js
hola-ola: True, good point! Thanks :)

PR #2202, opened 2021-01-20
[BLN-PT-JAN2021] Gizem Wanka

gizeminthecity:
iterations without bonus

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,73 +1,603 @@
 // Iteration #1: Find the maximum
-
+function maxOfTwoNumbers(num1, num2) {
+    if (num1 > num2) {
+        return num1;
+    } else if (num2 > num1) {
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,73 +1,603 @@
...
+    if (arr.length === 0) {
+        return null;
+    }
+
+    let average = 0;
+    let sum = 0;
+
+    for (num of arr) {
+        sum += num;
+    }
```
IH-WebDev-TA-Remote, line 66: You could also have used the previously declared sumNumbers() function (line 40) here, using its value to calculate the averga!

IH-WebDev-TA-Remote (commented):
Congratulations Gizem, everything is well done! :tada: Left a comment regarding code re usability but other than that, nice work!

PR #2203, opened 2021-01-21
[RMT-PT-JAN2021] Ola Pietraszek

ollayka:
Iterations without bonuses

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,18 +1,82 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
+
+maxOfTwoNumbers(7, 9);
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+function findLongestWord(arr) {
+  if (arr.length === 0) {
+    return null;
+  } else if (arr.length === 1) {
+    return arr[0];
+  }
+  let longestWord = "";
+  for (let word of arr) {
+    if (word.length > longestWord.length) {
+      longestWord = word;
+    }
+  }
+  return longestWord;
+}
+
+findLongestWord(words);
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,18 +1,82 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
```
IH-WebDev-TA-Remote, line 8: Here one could also use the Math.max(num1, num2) in order to find the maximum of two numbers! Find more about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max) :smile:

src/functions-and-arrays.js
```diff
@@ -1,18 +1,82 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+function averageNumbers(arr) {
+  if (arr.length === 0) {
+    return null;
+  } else if (arr.length === 1) {
+    return arr[0];
+  } else {
+    return sumNumbers(arr) / arr.length;
```
IH-WebDev-TA-Remote, line 57: Perfect use of the previous sumNumbers function! :+1:

src/functions-and-arrays.js
```diff
@@ -28,9 +92,34 @@ const wordsUnique = [
...
+function doesWordExist(arr, word) {
+  if (arr.length === 0) {
+    return null;
+  } else if (arr.length === 1 && arr.includes(word)) {
+    return true;
+  } else if (arr.includes(word)) {
+    return true;
+  } else {
+    return false;
+  }
```
IH-WebDev-TA-Remote, line 121: Another possible solution would be:
`
for (let occurrence of arrayOfWords) {
    if (occurrence === word) return true;
  }
  return false;
`

IH-WebDev-TA-Remote (commented):
Fantastic job Ola! :smile: Not much more I can add here! Congratulations and keep up the good work! :medal_sports:

PR #2209, opened 2021-01-22
[WD-PT-SAO-01-2021] Michelle Mendonça

michaloumen:


ingridpitta:
Congratulations! Nice that you used Math.max() and "early return". 

In bonus 3.1 (sumOfElements) you could have used a switch, as it would make reading easier and avoid "if's". 

In line 91 (if (! Numbers.lenght) ....) would be "length" is a common mistake in that word :) 

The "averageWordLength" function has no return, instead, it has a console.log (). Tip: you can reuse the previous functions to add the items in the array and then average by dividing the sum result by the number of items in the array.

PR #2210, opened 2021-01-25
[WD-PT-SAO-01-2021] Anderson-Thiago

anderson-santiago-silva:
Done 100%!! =)


ingridpitta:
Congratulations! It was very cool to use the "early return" and that you have reused functions.

PR #2217, opened 2021-02-04
[PAR-WDFT- 102] Augustin LEMAN

august159:
I have pleased Jasmine but not in the bonuses yet.
Also I have changed the color of Jasmine to be more color-blind friendly


ta-web-paris:
That's an amazing idea about Jasmine's colors, I'll take it to the people managing it so they can maybe implement changes globally ! Good thinking about accessibility ! Are you colorblind ? Just out of curiosity :) 

Nina

august159:
I am color-blind indeed. Glad of it can help others.

Thanks for the feedback.

PR #2250, opened 2021-03-23
[FTAMS032021] Andrea Sforza

arsforza:
Completed all required and bonus assignments

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,15 +1,76 @@
 // Iteration #1: Find the maximum
+function maxOfTwoNumbers(number1, number2) {
+  if(number1 > number2)
+      return number1;
+  else if(number1 < number2)
+      return number2;
+  else
+      return number1, number2;
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+function findLongestWord(wordArr) {
+  if(wordArr === null || wordArr.length === 0) return null;
+
+  let longestWord = '';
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -71,3 +157,88 @@ const matrix = [
...
+      if (getProduct(matrix, x, y) === -1 || getProduct(matrix, x, y) <= product)
+        continue;
+      product = getProduct(matrix, x, y);
+    }
+  }
+  
+  return product;
+}
+
+function getProduct(matrix, x, y) {
```
reMRKableDev, line 177: Very nice work here! As a tip, I would suggest using a more descriptive naming convention when it comes to the parameters of your functions. In the case here `matrix` is fine, however, the `x` and `y` could give more detail to the type of incoming data. 

See the suggested example below 👇🏾 

```suggestion
function getProduct(matrix, xCoordinate, yCoordinate) {
```

src/functions-and-arrays.js
```diff
@@ -71,3 +157,88 @@ const matrix = [
...
+  let productH = 1;
+  let productV = 1;
```
reMRKableDev, line 179: The above comment can also be considered here. `productH` and `productV` are not very descriptive just by reading through the code. Consider using one of these conventions to add more description 👉🏾  `horizontalProduct` or `productHorizontal` etc

reMRKableDev (commented):
Nice work @arsforza 🚀 

Your code works well and passes all the Jasmine tests. Kudos for successfully completing the Bonus sections. I made a few comments on naming conventions that you can check out 👍🏾

PR #2251, opened 2021-03-23
[BER-FT-032021] Philipp Dam

Philipphub:


src/functions-and-arrays.js
```diff
@@ -1,73 +1,462 @@
...
+  "communion",
+  "simple",
+  "bring",
 ];
 
+const uniquifyArray = (arr) => {
+  var newArr = {};
+  return arr.filter(function (item) {
+    return newArr.hasOwnProperty(item) ? false : (newArr[item] = true);
+  });
```
annawaszak, line 103: Interesting approach. You could also do indexOf() method here, that way you don't even that newArr variable.

annawaszak (commented):
Hi Philipp,

Apologies for the delay, I didn't press submit on this one. Great job on the lab! I can tell you have a good understanding of the concept covered during lectures. 

I would not be doing my job well if I have not pointed out a few ways to improve your code - please find the comments below. 

The only test you have not passed is "should return null if receives an empty array when called" in iteration 4.1, 4.1, 5 and 6. Interestingly, you have passed the same test in iterations 2 and 3, so I'm confident you can handle it. 

If you have any questions or suggestions, don't hesitate to let me know. 

Thanks,
Anna

PR #2252, opened 2021-03-23
ftwebdevams032021 Jeroen, Silvia & Ayoub

jeroenvanbart:


src/functions-and-arrays.js
```diff
@@ -1,15 +1,44 @@
...
+const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearlllll', 'orchard', 'crackpot'];
+let findLongestWord = (array) => {
+    let word ="";
+    for (let i=0 ; i<array.length; i++ ){
+        if(word.length < array[i].length){
+            word =array[i];
+        }
+    }
+    return word;
+}
```
reMRKableDev, line 17: It seems the Jasmine tests for this function are failing with the following warning 👉🏾  `Find the longest word should return null when called with an empty array`

The test expects there to be some handling for empty arrays. Specifically, it wants the function to return `null` if an empty array is passed.
Screenshot 2021-03-24 at 12 40 49


---
One thing you can do here is to add the empty array handling before any of the function's operations. This way, you ensure that you are taking care of this edge case before anything else.

```suggestion
let findLongestWord = (array) => {
/* Handle empty array edge case here:*/
 // 1. Check if the incoming array is empty.
 // 2. Return "null" if it is.
 
// code goes here...

/* Function's operation if the incoming array is not empty */ 

    let word ="";
    for (let i=0 ; i<array.length; i++ ){
        if(word.length < array[i].length){
            word =array[i];
        }
    }
    return word;
}
```

src/functions-and-arrays.js
```diff
@@ -1,15 +1,44 @@
...
+
+    }
+    return sum;
+ }
+console.log(sumNumbers(numbers));
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
+let averageNumbers = sumNumbers(numbersAvg)/numbersAvg.length;
```
reMRKableDev, line 40: `averageNumbers` should be declared as a function here. For the moment, you are only creating a variable that is storing the results from `sumNumbers(numbersAvg)/numbersAvg.length;`

---

What you want to do instead is something like this 👇🏾 
```suggestion
let averageNumbers = (numbersAvg) =>  sumNumbers(numbersAvg)/numbersAvg.length;
```

reMRKableDev (commented):
Nice work @jeroenvanbart + co 🚀 

I have left some comments you can check out. They should guide you in the right direction and fix the few errors you have so far. 
Be sure to finish all the other iterations as well 👍🏾

reMRKableDev (commented):
Nice work @jeroenvanbart + co 🚀 

I have left some comments you can check out. They should guide you in the right direction and fix the few errors you have so far. 
Be sure to finish all the other iterations as well 👍🏾

PR #2253, opened 2021-03-23
[FTWDAMS032021]- Jamal Nasser

jamal-nasser:


src/functions-and-arrays.js
```diff
@@ -1,16 +1,51 @@
...
+function findLongestWord(words) {
+let longest = '';
+for (let i = 0; i < words.length; i++) {
+if (words[i].length > longest.length)
+longest = words[i];
+}
+return longest;
+}
+
+console.log(findLongestWord(words));
```
reMRKableDev, line 18: It seems the Jasmine tests for this function are failing with the following warning 👉🏾  `Find the longest word should return null when called with an empty array`

The test expects there to be some handling for empty arrays. Specifically, it wants the function to return null if an empty array is passed.

---

One thing you can do here is to add the empty array handling before any of the function's operations. This way, you ensure that you are taking care of this edge case before anything else.

```suggestion
function findLongestWord(words) {
/* Handle empty array edge case here:*/
 // 1. Check if the incoming array is empty.
 // 2. Return "null" if it is.
 
// code goes here...
/* Function's operation if the incoming array is not empty */ 

let longest = '';
for (let i = 0; i < words.length; i++) {
if (words[i].length > longest.length)
longest = words[i];
}
return longest;
}

console.log(findLongestWord(words));
```

src/functions-and-arrays.js
```diff
@@ -1,16 +1,51 @@
...
 
+
+function averageNumbers(numbersAvg){
+    let total=0;
+    for(let i in numbersAvg) { 
+        total += numbersAvg[i]/numbersAvg.length;
+    }
+    return total;
+}
+averageNumbers(numbersAvg);
```
reMRKableDev, line 47: The same implementation for the empty array is expected here. The concept is the same as mentioned in the comment above.

src/functions-and-arrays.js
```diff
@@ -29,9 +64,32 @@ const wordsUnique = [
...
 ];
 
+function  uniquifyArray(wordsUnique) {
+  let newArr = [];
+  wordsUnique.forEach(function(i) {
+    if(!newArr[i]) {
+      newArr[i] = true
+    }
+  })
+  return Object.keys(x)
```
reMRKableDev, line 74: I think you want to get the keys from `newArr` and not `x` here.

```suggestion
  return Object.keys(newArr)
```

src/functions-and-arrays.js
```diff
@@ -1,16 +1,51 @@
...
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+
+function averageNumbers(numbersAvg){
+    let total=0;
+    for(let i in numbersAvg) { 
+        total += numbersAvg[i]/numbersAvg.length;
+    }
+    return total;
```
reMRKableDev, line 45: This solution seems to still give errors in the Jasmine tests.  A better approach would be to divide the `total` with the `numbersAvg.length`

```suggestion
        total += numbersAvg[i];
    }
    return total /numbersAvg.length;
```

---

This is one change, however, you have a few other things to fix. Check the Jasmine responses to see what you need to make better

reMRKableDev (changes requested):
Nice work @jamal-nasser 🚀 

This is a good attempt. I've left some comments you can check out to improve your solutions. Revisit your solutions and make sure the Jasmine tests are all green when you turn them in.

PR #2254, opened 2021-03-23
[FTAMS032021] SilviaRodriguez, Jeroen, Ayoub

silrodcar:


src/functions-and-arrays.js
```diff
@@ -1,20 +1,80 @@
...
+
 // Level 1: Array of numbers
+
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+let total = 0;
+for(let i = 0; i < numbersAvg.length; i++) {
+    total += numbersAvg[i];
+}
+let avg = total / numbersAvg.length;
```
reMRKableDev, line 57: You may have noticed that all your Jasmine tests are still red. The reason is this code block here.

The problem here is that you are not actually creating a function. If you placed these operations in a function, then the Jasmine test suite should run as it should.

```suggestion
const numbers = [2, 6, 9 ,10, 7 ,4 ,1 , 9];

function numbersAvg(arrOfNumbers) {
  let total = 0;
  for(let i = 0; i < arrOfNumbers.length; i++) {
      total += arrOfNumbers[i];
  }
  
  let avg = total / arrOfNumbers.length;
  
  // return some value....
}

```

reMRKableDev (commented):
Nice work @silrodcar 🚀 

This is a good first attempt. I've left you some comments on how you can fix the failing Jasmine suite. Please check it out and finish off the remaining iterations

PR #2257, opened 2021-03-23
Excercises except last bonus finished

AndyMcGN:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,96 @@
...
+  return num2;
+};
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+function findLongestWord(words) {
+  if (words.length === 0) {
+    return null;
+  }
```
wax-bear, line 16: Perhaps create an empty line after this one, for the sake of readability.

src/functions-and-arrays.js
```diff
@@ -1,19 +1,96 @@
...
+  return sum;
+}
+const mixedArr = [6, 12, 'miami', 1, true, 'barca', '200', 'lisboa', 8, 10]
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
+function averageNumbers(numbers) {
+  if (numbers.length === 0){
```
wax-bear, line 62: After the closing parenthesis, a space would not hurt.

src/functions-and-arrays.js
```diff
@@ -1,19 +1,96 @@
...
+    return null;
+  }
+  sum = sumNumbers(numbers);
```
wax-bear, line 65: Use a variable keyword (let/const/var) for declaring variables. In this case I would use const, as "sum" is not reassigned.

wax-bear (commented):
Great stuff Andy!
Please check out the comments for potential improvements.

Best,

Thomas

PR #2263, opened 2021-03-23
[BCN - FT - 032021] Beatriz Avis + Toni Casasayas

BeAvisP:
Done until Bonus 8.

Jasmine has been a bit of a headache...


BeAvisP:
Update bonus 8 and add bonus 8.1

PR #2269, opened 2021-03-23
[BER-WDFT-230321] Lab Javascript Functions and Arrays - Pietro Ballarin

pietroballarin:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,86 @@
...
+
+maxOfTwoNumbers (166, 8)
+
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
+function findLongestWord(wordArr) {
+  if (wordArr.length === 0) {
+    return null;
+  }
```
annawaszak, line 21: Quick tip - providing the "if" statement is a one liner, then you don't need to add the curly brackets and write everything in one line. 
`if (wordArr.length === 0) return null`

src/functions-and-arrays.js
```diff
@@ -1,19 +1,86 @@
 // Iteration #1: Find the maximum
 
+function maxOfTwoNumbers (num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
+    } else {
+      return num1, num2;
+    }
```
annawaszak, line 10: the indentation of the "else" statement should be on the same level as "if" and "else if"

annawaszak: You don't actually need three statements here. Since, it does not matter which number you return if they are equal, you can handle this with if ... else ...

annawaszak (commented):
Hi Pietro!

Amazing job on this lab. You knocked it out of the park! It makes me think you could conquer those bonus iterations :)

I would not be doing my job though if I didn't point out a few improvement tips :) Please find detailed comments below.

Thanks,
Anna

pietroballarin:
@annawaszak 

Thanks for your feedback, Anna! :)

PR #2384, opened 2021-06-01
[Berlin-FT-MAY2021] Raoul Ullens

Ra-Ul-1:
Several tests to be passed still

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,13 +1,39 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a
+  };
+  if (a < b) {
+    return b
+  };
+  if (a === b) {
+    return a, b
+  };
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(words) {
+  let longest = ""
+  for (let word of words) {
+    // Return null when array is empty
+    if(!words.length) {
+      return null
+    };
+    // Return the word when called with a single-word array
+    if(words.length = 1) {
+      return words[0]
+    };
+    // return longest occurence when multiple words
+    if(words[i].length > longest.length) {
+      longest = words[i]
+    };
+  }
+  return longest
+}
 
 
 // Iteration #3: Calculate the sum


src/functions-and-arrays.js
```diff
@@ -1,13 +1,39 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a
+  };
+  if (a < b) {
+    return b
+  };
```
ahmedhany82, line 8: You could also do it as an if-else, or just an if. The equality can be handled as one of the two cases, in the example below, if they are equal, b will be returned. The function ends when the return keyword is reached. What comes after return is not executed so this is why we can do without an else.

```
  if (a > b) {
    return a;
  }
  return b;
 }
```

src/functions-and-arrays.js
```diff
@@ -1,13 +1,39 @@
...
 
-function findLongestWord() {}
-
+function findLongestWord(words) {
+  let longest = ""
+  for (let word of words) {
+    // Return null when array is empty
+    if(!words.length) {
+      return null
+    };
```
ahmedhany82, line 25: This should be before the for loop. The for loop will not iterate if the array is empty and so this line will never be reached and the function will return an empty string if the array is empty.

src/functions-and-arrays.js
```diff
@@ -1,13 +1,39 @@
...
+    // Return the word when called with a single-word array
+    if(words.length = 1) {
+      return words[0]
+    };
```
ahmedhany82, line 29: We can do without this since if the array has 1 element, the for-loop will just iterate once and will give the same result.

src/functions-and-arrays.js
```diff
@@ -1,13 +1,39 @@
...
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a
+  };
+  if (a < b) {
+    return b
+  };
+  if (a === b) {
+    return a, b
```
ahmedhany82, line 10: I think this returns only b. JavaScript does not allow you to return 2 values.

ahmedhany82 (commented):
Hey Raoul,
I added some comments on what you have so far. Perhaps if you have some time, you can continue working on the remaining iterations and maybe give the bonus a try. This lab and Friday's one are fundamental so the more time you spend on them, the easier it will get later on. Feel free to reach out if you have any questions on the lab even after submission. 
Best regards,
Ahmed

PR #2385, opened 2021-06-01
BER-WDFT-MAY2021 Irina

IraAwesomePossum:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,76 @@
...
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  } else if (b > a) {
+    return b;
+  } else {
+    return a || b;
```
ahmedhany82, line 8: This will not behave as you may expect. It will return `a` all the time unless` a = 0` because 0 evaluates to a falsy value and any other value evaluates to true. In this case, it will return `b`. It will still pass the test case because we will not reach this line of code unless a and b are equal, so it won't matter.

If you want to handle the equality case, could change line 3 to be `if(a >= b)` and then it returns `a` if `a` and `b` are equal.

You could do this in a shorter version. Check the lab solution posted by Jan.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,76 @@
...
-function findLongestWord() {}
-
-
+function findLongestWord(words) {
+  if (words.length === 0) {
+    return null;
+  } else {
+    return words.reduce((a, b) => (a.length < b.length ? b : a), '');
+  }
+}
```
ahmedhany82, line 22: Good that you used reduce but take a look on the lab solution to also be familiar with other ways to do it.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,76 @@
...
-function averageNumbers() {}
-
+function averageNumbers(arr) {
+  if (arr.length === 0) return null;
+  let averNums = 0;
+  for (let i = 0; i < arr.length; i++) {
+    averNums += arr[i];
+  }
+  return averNums / arr.length;
+}
```
ahmedhany82, line 56: This works and you could also use the `sum` function defined above.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,76 @@
...
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(arr) {
+  if (arr.length === 0) return null;
+
+  if (arr.length === 1) return arr[0].length;
+
+  return arr.join('').length / arr.length;
+}
```
ahmedhany82, line 67: Nice!

src/functions-and-arrays.js
```diff
@@ -52,16 +87,25 @@ const wordsUnique = [
...
 ];
 
-function uniquifyArray() {}
-
-
+function uniquifyArray(arr) {
+  if (arr.length === 0) {
+    return null;
+  } else {
+    return Array.from(new Set(arr));
```
ahmedhany82, line 94: Nice! you could also the spread operator to turn the set into an array `[...new Set(arr)]`

src/functions-and-arrays.js
```diff
@@ -78,9 +122,10 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
-
-
+function howManyTimes(arr, word) {
+  if (arr.length === 0) return 0;
+  if (arr.indexOf(word) === -1) return 0;
+}
```
ahmedhany82, line 128: For this one, you will need to loop on the array and increment a counter each time you find the target word.

Take a look on the lab solution posted on github.

ahmedhany82 (commented):
Hey Irina,
Good job overall. I added some comments below. I see you used reduce from some iterations which is impressive but make sure you are also familiar with other ways do it. Maybe give the bonus a try if you have some time.  Feel free to reach out if you have any questions.
Best regards,
Ahmed

PR #2388, opened 2021-06-01
 BER - WDFT - MAY21 - Alexandre Oliveira

alexandreoliv:
Solved lab + bonuses

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,41 +1,153 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
 
+function maxOfTwoNumbers(a, b) {
+  console.log(arguments.length);
+  if (arguments.length === 0) {
+    return 0;
+  }
 
+  if (arguments.length === 1) {
+    return arguments[0];
+  }
+
+  if (a >= b) {
+    return a;
+  }
+  else {
+    return b;
+  }  
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  let longestSize = 0;
+  let longestWordPosition = 0;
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -78,37 +210,81 @@ const wordsCount = [
...
+    for (let j = 0; j < matrix.length; j++) {
+      productColumn = matrix[i][j] * matrix[i+1][j] * matrix[i+2][j] * matrix[i+3][j];
+      if (productColumn > maxProduct) {
+        maxProduct = productColumn;
+      }
+    }
+  }
+    
+  return maxProduct;
+}
```
atkat, line 256: Super well done on both of these bonuses. This works perfectly fine, but you can actually put the two nested loops together: 
```
  let result = 0;
  let horizontal = 0;
  let vertical = 0;
  for (let j = 0; j < 20; j++) {
    for (let i = 0; i < 17; i++) {
      horizontal = matrix[j][i] * matrix[j][i + 1] * matrix[j][i + 2] * matrix[j][i + 3];
      vertical = matrix[i][j] * matrix[i + 1][j] * matrix[i + 2][j] * matrix[i + 3][j];
      result = Math.max(horizontal, vertical, result);  // this you can also do the way you did it if you want
    }
  }
  return result;
}
```

src/functions-and-arrays.js
```diff
@@ -78,37 +210,81 @@ const wordsCount = [
...
+    for (let j = 3; j < matrix.length; j++) {
+      product = matrix[i][j] * matrix[i+1][j-1] * matrix[i+2][j-2] * matrix[i+3][j-3];
+      if (product > maxProduct) {
+        maxProduct = product;
+      }
+    }
+  }
+    
+  return maxProduct;
+}
```
atkat, line 287: Maybe you can try putting this one in only one nested loop too? If you don't manage we can send you the solution :). 

Really well done!!!

atkat (commented):
Excellent work!! Please see individual comments.

PR #2391, opened 2021-06-01
[BER-WDFT-MAY2021]- Hanlin Li - Lab June 1st

PemaHL:


src/functions-and-arrays.js
```diff
@@ -1,20 +1,54 @@
...
+  if (a > b) {
+    return a;
+  }
+  if (a < b) {
+    return b;
+  }
+  if (a = b) {
+    return a
+  }
+}
```
ahmedhany82, line 12: This works fine. You can also do it in a shorter way. See below:
```
  if (a > b) {
    return a;
  }
  return b;
 }
```

src/functions-and-arrays.js
```diff
@@ -1,20 +1,54 @@
...
+  if (arr.length === 1) {
+    return arr[0];
+  }
+  let longestWord = "";
+  //iterate over each words in the array
+  for (let i = 0; i < arr.length; i++) {
+    if (longestWord.length < arr[i].length) {
+      longestWord = arr[i];
+    }
+  }
```
ahmedhany82, line 35: I think you missed to `return longestWord` here. This is why the last 2 tests are failing.

src/functions-and-arrays.js
```diff
@@ -52,15 +100,39 @@ const wordsUnique = [
...
+  const newUniqueArr = [];
+  for (let i = 0; i < array5.length; i++) {
+    if (array5.indexOf(array5[i]) === array5.lastIndexOf(array5[i])) {
+      newUniqueArr.push(array5[i]);
+    }
+  }
+  return newUniqueArr;
+}
+console.log(uniquifyArray(wordsUnique))
```
ahmedhany82, line 115: Hmm we discussed this yesterday. It seems this will copy all the unique elements to `newUniqueArr` and will skip the duplicated ones. I think the requirement for this function is that for duplicate elements, we still need to pick just one of them. So this logic will not do it. 

In this case, I think we could use `indexOf` on the `newUniqueArr` like so:
```
 function uniquifyArray(ar) {
   let newUniqueArr = [];
   for (let el of ar) {
     if (newUniqueArr.indexOf(el) == - 1) {
       newUniqueArr.push(el)
     }
   }
   return newUniqueArr;
 }
```

src/functions-and-arrays.js
```diff
@@ -52,15 +100,39 @@ const wordsUnique = [
...
 
 // Iteration #6: Find elements
+
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(keyWord, array6) {
```
ahmedhany82, line 121: I think the function parameters are switched in the test case. It expects the array to be the first argument. Take a look on functions-and-arrays.spec.js

src/functions-and-arrays.js
```diff
@@ -52,15 +100,39 @@ const wordsUnique = [
...
+  if (array6.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < array6.length; i++) {
+    if (array6[i] === keyWord) {
+      console.log('True')
+      break;
```
ahmedhany82, line 128: You need to `return true` here. Remember `console.log` does not return anything. A `return` statement is needed for the function to return a value

src/functions-and-arrays.js
```diff
@@ -52,15 +100,39 @@ const wordsUnique = [
...
+    }
+    else
+    console.log('False')
```
ahmedhany82, line 131: There is no need for an else here. After the for loop ends, if we did not find the keyWord, we should return false.

```
function doesWordExist(array6, keyWord) {
  if (array6.length === 0) {
    return null;
  }
  for (let i = 0; i < array6.length; i++) {
    if (array6[i] === keyWord) {
      return true;
    }
  }
  return false;
}
```

src/functions-and-arrays.js
```diff
@@ -78,9 +150,21 @@ const wordsCount = [
...
-function howManyTimes() {}
-
-
+function howManyTimes(array7, keyWord2) {
+  if (array7.length === 0) {
+    return 0;
+  }
+  if (keyWord2.length === 0) {
+    return null;
+  }
```
ahmedhany82, line 159: I think this check is not needed.

ahmedhany82 (commented):
Hey Hanlin, 
Good job with the lab. I added a few remarks below. I would say try to give the bonus a try if you have some time. Feel free to reach out if you have any questions.
Best regards,
Ahmed

PR #2395, opened 2021-06-01
BER - WDFT - Alan Sanchez

alnwolff:
All the mandatory Iterations are done.
The Bonus are not done yet.


atkat:
Hey Alan,

Good job, you did all the necessary iterations and they're very clean. 

I really just have one thing:

in iteration 6, you can save yourself some typing by just returning true when it's true and false in all other cases (no need for an extra helper variable):
          `for (let el of ar) {
               if (el === word) {
                 return true;
               }
             }
             return false; `

You've done well, do you wanna try to do some of the bonuses maybe?

Best,
Katka

PR #2396, opened 2021-06-01
BER-WDFT-062021 - Andrew Oppo

andrewoppo:
The tests that were run on the lab solutions kept coming up failed even though my console ran the exact same calculations on the exact same code and returned the values that should have made them pass. Somehow, additional values were being added to the totals, but I couldn't figure it out.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,58 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  } 
+  return b;
+}
 
 
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -52,14 +110,39 @@ const wordsUnique = [
...
+  }
+}
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(word, someArr) {
```
atkat, line 133: Also, minor thing but the order of arguments is reversed. It doesn't matter if you know how to feed the function calls. But it won't work for the tests as they have been instructed to call howManyTimes(word, someArr). The same happens in the next function.

atkat (commented):
Very nicely done, just some small things (see comments).

PR #2399, opened 2021-06-01
BER-WDFT-MAY-2021-Zoriglon Ayush

timbstar58:
done without bonus

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,21 +1,47 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2)
+    return num1;
+  else 
+    return num2;
+
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,21 +1,47 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2)
+    return num1;
+  else 
+    return num2;
+
+}
```
ahmedhany82, line 8: You can do it without the `else` but this is just a detail. Take a look on the lab solution posted on github.

src/functions-and-arrays.js
```diff
@@ -1,21 +1,47 @@
...
+  let longestWord = '';
+  for (let i=0; i<arr.length; i++){
+      if (longestWord.length < arr[i].length ){
+      longestWord = arr[i];
+      }
+  }
+  return longestWord;
+}
+  
+findLongestWord(words)
```
ahmedhany82, line 28: You don't need to call the functions, they are called by the test cases,

ahmedhany82 (commented):
Hey Zoriglon,
Good job with the lab. I added some comments below. Maybe if you have some time, try to give the bonus part a try. Feel free to reach out in case you have any questions.
Best regards,
Ahmed

PR #2405, opened 2021-06-01
Ber - WDFT - Juan

Mangonando:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,96 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(a, b) {
+  if (a >= b) {
+    return a;
+  } else  {
+    return b;
+  }   
+}
```
atkat, line 8: General tip: providing your if statement is a one liner, you can skip the curly brackets

` if (arr.length === 0) return null;`

Also, we can actually write it like this (below), this way the function will always return b unless the condition is met. It's ok to write it the way you did, it's just a nice thing to know.

```
  if (a >= b) {
    return a;
  } else  {
    return b;
  }   
}
```
Edit: Oh, I've just noticed you're doing it already in the next one.

src/functions-and-arrays.js
```diff
@@ -52,15 +110,36 @@ const wordsUnique = [
...
+    return null;
+  }
+    let wordSeeker = false;
+    for(let i of wordsFind) {
+      if(i == value) {
+          wordSeeker = true;
+      }   
+    } 
+    return wordSeeker;  
+}
```
atkat, line 142: Here, you can save yourself some typing by just returning true when it's true and false in all other cases (no need for an extra helper variable):
```
for (let el of ar) { 
  if (el === word)  return true;  
   return false;
```

atkat (commented):
Hey Juan,

Good job, you did all the necessary iterations and they're clean.

I really just have one thing:

in iteration 6, you can save yourself some typing by just returning true when it's true and false in all other cases (no need for an extra helper variable):
for (let el of ar) { if (el === word) { return true; } } return false;

You've done well, do you wanna try to do some of the bonuses maybe?

Best,
Katka

src/functions-and-arrays.js
Mangonando: Thanks for your feedback. It is definitely good to learn tips and different ways to write which will be helpful in the future.

src/functions-and-arrays.js
Mangonando: This one looks great. Thanks.


Also, yes, I am looking forward to doing the bonuses but I need to stay up to date with the course first and so far it is demanding my time.

PR #2406, opened 2021-06-01
BER-WDFT-052021-Nelly Neumann -Solved Lab

nellyneumann:
I had a problem with the exercise #7 - Count repetition.
I couldn't resolve it.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,38 +1,68 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2) {
+    return number1;
+  } else {
+    return number2;
+  }
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(wordarray) {
+  if (wordarray.length === 0) {
+    return null;
+  }
 
-
+  let longestword = '';
+  for (let i = 0; i < wordarray.length; i++) {
+    if (wordarray[i].length > longestword.length) {
+      longestword = wordarray[i];
+    }
+  }
+  return longestword;
+}
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
-
-
+function sumNumbers(numbersarray) {
+  let result = 0;
+  for (let i = 0; i < numbersarray.length; i++) {
+    result = result + numbersarray[i];
+  }
+  return result;
+}
 
 // Iteration #3.1 Bonus:
 function sum() {}
-
-
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,38 +1,68 @@
...
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
-
-
+function sumNumbers(numbersarray) {
+  let result = 0;
+  for (let i = 0; i < numbersarray.length; i++) {
+    result = result + numbersarray[i];
```
ahmedhany82, line 33: You could also write this as `result += numbersarray[i];` but this is just a tiny detail.

src/functions-and-arrays.js
```diff
@@ -1,38 +1,68 @@
...
-function averageNumbers() {}
-
+function averageNumbers(numbersarray) {
+  if (numbersarray.length === 0) {
+    return null;
+  }
+let sum = sumNumbers (numbersarray)
+let numberOfElements= numbersarray.length
+let average = sum / numberOfElements
+return average
```
ahmedhany82, line 52: Another tiny detail. Just so that you know, you could do this in one line `return sumNumbers(numbersarray) / numbersarray.length`

src/functions-and-arrays.js
```diff
@@ -1,38 +1,68 @@
...
 
-function averageWordLength() { }
+function averageWordLength(wordsarray) {
+  let wordlist = []
+  for (let i = 0; i < wordsarray.length; i++){
+    wordlist.push(wordsarray[i].length)
+  }
+  let average = averageNumbers(wordlist)
+  return average
+}
```
ahmedhany82, line 65: This works fine but it creates an extra array. Since the length is just a number we can add all the lengths in a variable instead. But the way you did it also works.

```
function averageWordLength(words) {
  let sum = 0;
  for (let i=0; i < words.length; i++) {
    sum += words[i].length;
  }
  return sum / words.length;
}
```

src/functions-and-arrays.js
```diff
@@ -52,16 +82,25 @@ const wordsUnique = [
...
+    return null;
+  }
+  let found = false
+  for (let i = 0; i < wordsFindArray.length; i++){
+    if (wordsFindArray [i] === word){
+      found = true
+    }
+  }
+  return found
+}
```
ahmedhany82, line 103: This works but if the target word is the first word in the array and the array has 1000 elements, it will still loop on the entire array before returning true. Instead you could directly return true in line 99 so the function ends and return true as soon as the target is found. If the for loop is finished, then this means the target was not found and you can return false.

src/functions-and-arrays.js
```diff
@@ -52,16 +82,25 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
-
+function uniquifyArray(wordsUniqueArray) {
 
+}
```
ahmedhany82, line 87: Check the lab solution for this one and feel free to reach out if you have any questions.

ahmedhany82 (commented):
Hey Nelly, 
Good job with the lab. I added a few comments below. I see you mentioned that you had a problem with the count repetition but it looks fine to me. Perhaps you meant unique array ? Let me know if you have any specific questions. If you have some time, give the bonus a try ;)
Best regards,
Ahmed

PR #2413, opened 2021-06-02
Sinah Scholz-Hehn WDFT-RMT-MAY21

r00kie123:
My goal was to try at least all non-bonus iterations, but after yesterday, that was too much for me :/


TA-Remote:
Hey Sinah! Good job on the first 3 iterations. Try to push a little more to do the other ones fot the next labs. You can do it :)

PR #2415, opened 2021-06-02
Anja WDFT-RMT-MAY21

AnjaRL:


TA-Remote:
Hey Anja! I saw you did all the iterations except the bonus.
The good part 🥇
-Everything is working and you have a very good logic, well done 💯 .
-You are pretty good! I encourage you to always give a try to the bonus!
In order to improve I would just say 👍
-When you are checking if you are receiving empty arrays try to use array.length === 0 to check it instead of just array == 0.
Very nice job!. Keep working hard and practicing.

PR #2421, opened 2021-06-02
BER-WDFT-MAY2021-Joanna

joannaberlin:


src/functions-and-arrays.js
```diff
@@ -26,7 +47,12 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  let sum = 0;
+  for (let i = 0; i < arr.length; i++) {
+    sum += arr[i];
+  } return avg = sum / arr.length;
+}
```
ahmedhany82, line 55: An issue can happen here if the array is empty. In this case, it will have a length of 0 dividing by zero gives Infinity. To avoid that, you could add an if-condition to check the array length and if it is zero, return null. This is why one test case fails for this function.

ahmedhany82 (commented):
Hey Joanna, 
I added some comments to what you have submitted so far. Feel free to continue working on this lab if you have some time and you can reach out if you have any questions anytime.
Best regards,
Ahmed

PR #2426, opened 2021-06-03
Gergö Eperjesi - WDFT Maj2021

GergoEper:
Solved lab

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,35 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 < num2) return num2
+   else 
+   return num1
+  }
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(arr) {
+  let longestWord = arr[0]
+for (let words of arr) {
+  if (words.length > longestWord.length) {
+    longestWord = words
+  }
+}
+return longestWord;
+}
 
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers(arrOfNum) {
+  let sum = arrOfNum[0];
+  for (let num of arrOfNum) {
+    sum += num
+  }
+  return sum;
+}
 
 
 
@@ -26,7 +42,13 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arrOfNum) {
+  let sum = arrOfNum[0];
+  for (let num of arrOfNum) {
+    sum += num
+  }
+  return sum / arrOfNum.length;
+}
 
 
 // Level 2: Array of strings


src/functions-and-arrays.js
```diff
@@ -1,19 +1,35 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 < num2) return num2
+   else 
+   return num1
+  }
```
ahmedhany82, line 6: If statements work without curly braces in case the body is just one line. For an if-else statement, the best practice is to use curly brackets. This can help avoid unexpected errors where the code behaves differently than what you expect.

src/functions-and-arrays.js
```diff
@@ -1,19 +1,35 @@
...
-
+function findLongestWord(arr) {
+  let longestWord = arr[0]
+for (let words of arr) {
+  if (words.length > longestWord.length) {
+    longestWord = words
+  }
+}
+return longestWord;
+}
```
ahmedhany82, line 20: There is one test case that expect you to return null when the array is empty. This you can do by a separate if-statement at the beginning of the function `if (arr.length === 0) return null;`.

src/functions-and-arrays.js
```diff
@@ -1,19 +1,35 @@
...
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(arr) {
+  let longestWord = arr[0]
+for (let words of arr) {
```
ahmedhany82, line 14: Since you initialized `longestWord` with `arr[0]`, you can start the loop from the second array element. You would need a normal for loop in this case, where you start the index with 1.
` for (let i=1; i < arr.length; i++)`

src/functions-and-arrays.js
```diff
@@ -1,19 +1,35 @@
...
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers(arrOfNum) {
+  let sum = arrOfNum[0];
+  for (let num of arrOfNum) {
+    sum += num
+  }
+  return sum;
+}
```
ahmedhany82, line 32: I think here the function will add the first element twice. Since sum is first initialized with the first element, then the for loop starts with the first element.

Since the test cases expect you to return 0 if the array is empty, it is better to initialize the sum with 0, if the array is empty, the for-loop will not iterate and the function will just return 0.

src/functions-and-arrays.js
```diff
@@ -26,7 +42,13 @@ function sum() {}
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arrOfNum) {
+  let sum = arrOfNum[0];
+  for (let num of arrOfNum) {
+    sum += num
+  }
+  return sum / arrOfNum.length;
+}
```
ahmedhany82, line 51: Same argument as above. You can basically use the sum function from above here. The function should also return null if the array is empty.

```
  if (numbers.length === 0) {
    return null;
  }
  const average = sumNumbers(numbers) / numbers.length;
  return average;
}
```

ahmedhany82 (commented):
Hey Gergö,
Good work. Please find my comments below. Feel free to continue working on the remaining iterations and if you have time, give the bonus a try. This lab and Friday's lab are foundational, the more time you spend on them, the easier it gets later on. Feel free to reach out if you have any questions.
Best regards,
Ahmed

PR #2445, opened 2021-06-04
BER-WDFT-MAR2021-JosuePrietoPaz

josppss96:


src/functions-and-arrays.js
```diff
@@ -52,7 +91,23 @@ const wordsUnique = [
...
+  let result = [];
+  
+  for (let word of wordsUnique){
+    if (result.indexOf(word) === -1){
+      result.push(word)
+    }}
+console.log(result)
+return result
+}
+uniquifyArray(wordsUnique)
```
ahmedhany82, line 105: You missed to return null if the array is empty.

src/functions-and-arrays.js
```diff
@@ -52,7 +91,23 @@ const wordsUnique = [
...
+
+    
+      
+  
+
```
ahmedhany82, line 112: I think you missed `doesWordExist` function

ahmedhany82 (commented):
Hey Josue,
Good work. I added some comments below. I would encourage you to make sure you are comfortable with the topics this lab covers. This lab and today's lab are foundational, so the more time you spend on them, the easier it gets later on. Try to check the bonus part if you have some time and feel free to reach out in case you have any questions (even after lab submission).
Best regards,
Ahmed

PR #2480, opened 2021-07-14
[WD PTRM]Cheryl Huang

CherylHuang1988:


IH-WebDev-TA-Remote:
Great Job on this lab Cheryl! 🎉A couple of things to have in mind for the future:
-> JavaScript uses _camelCase_ as a style guide, so when declaring variables, passing arguments/parameters to functions, we should use the camelCase convention! (e.g: `const studentName = "Cheryl"` instead of `const StudentName = "Cheryl")` 
-> Great job reusing the `sum` function on line 141! That's what they are made for! 🥇

That's all for this one, you did great! Keep it up! 😄

PR #2506, opened 2021-08-10
Berlin-WDFT-Rahaf

Farxa:
almost solved


Farxa:
The automated testing is showing that I haven't done anything correct starting Iteration #6: Find elements until the end. Not sure if all I am doing is wrong 😱 or something went wrong with the testing itself.

PR #2526, opened 2021-08-11
Carl Berlin WD functions and arrays

CarlBerthold:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,83 @@
...
 
-function averageWordLength() { }
+function averageWordLength(something) {
+  if (averageWordLength.length === 0){
+    return null;
+  }
+    arrLength = [];
+    for (let i = 0; i < something.length; i++) {
+      arrLength.push(something[i].length);
+    }
```
atkat, line 77: This works and what's nice about this is that you are using the other function. Although I would say that in here it's actually not so necessary because it needs to create a whole new array and that's not so efficient.  Using reduce would be best, but you can also just do something like Jan did:
``` 
let sum = 0;
    sum += words[i].length;
  }
  return sum / words.length;
```

src/functions-and-arrays.js
```diff
@@ -52,14 +97,31 @@ const wordsUnique = [
...
+  return unique;
+}
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(something, starting) {
```
atkat, line 113: Sorry, naming again :). It would be better to name it something like (arr, word), or (haystack, needle) if you'd like, just something explicit.

src/functions-and-arrays.js
```diff
@@ -52,14 +97,31 @@ const wordsUnique = [
...
+  if (something.length === 0){
+    return null;
+  } 
+  for(let findFind of something){
```
atkat, line 117: here you could do `(let element of array)`

atkat (commented):
Hey Carl, 

I have quite a few comments, please have a look below and let me know if you have questions. The naming is something that's common at the beginning, so don't worry but do try to think about it in the future :)

Best,
Katka

PR #2528, opened 2021-08-11
Berlin - WDFT - Ludumo Toni - Attempted lab solved

Ludumo:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,47 @@
...
-
+function findLongestWord(arr) {
+
+  let word = 0;
+  let finder = 0;
+  if (arr.length === 0) {
+    return null
+  } else { 
+  for (let i = 0; i < arr.length; i++) {
+    if (word.length < arr[i].length) {
```
atkat, line 25: `word` is currently a number, so you can't call the length method on it (only on strings or arrays). Instead word should be a string and then you just reassign it a new value each time the condition is true:
`word= arr[i]`

src/functions-and-arrays.js
```diff
@@ -1,19 +1,47 @@
...
+      word = arr[i].length;
+      finder = i;
+      }
+    }
+    return arr[finder]
+  }
+}
```
atkat, line 32: you also do not need the `finder`, look:
```
    if (word.length < arr[i].length) {
      word = arr[i];
      }
    }
    return word
  }
```

Also, you don't need to do an else if, you can save yourself a few lines by just doing an if and then running a loop after that. If the first 'if' is fulfilled, the function will return and stop anyway.

src/functions-and-arrays.js
```diff
@@ -26,13 +54,26 @@ function sum() {}
...
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  if (arr.length === 0) {
+    return null
+  } else {
+    let average = sumNumbers(arr) / arr.length;
+    return average
+  }
+}
```
atkat, line 64: same thing here with the if/else: 
```
 if (arr.length === 0) {
    return null
  } 
return sumNumbers(arr) / arr.length;

```

src/functions-and-arrays.js
```diff
@@ -52,14 +93,41 @@ const wordsUnique = [
...
+      if (word === arr[i]) {
+        exists = true
+        break
+      } else {
+        exists = false
+      }
+    }
+    return exists
+  }
+}
```
atkat, line 130: this works just fine but you could make it  a little more efficient: 
```
function doesWordExist(ar, word) {
   for (let el of ar) {
     if (el === word) {
       return true;
     }
   }
   return false;
}
```

atkat (commented):
Hey Ludumo, 

Please see my comments below :)

Best,
Katka

PR #2575, opened 2021-09-18
WDPT-REMOTE-BRIGIHORVATH

brigihorvath:
I have found that at the mixedAverage it wouldn't accept the answer, only if it is rounded.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -80,7 +80,7 @@ function avg(array) {
 
 const mixedArr = [6, 12, 'miami', 1, 'barca', '200', 'lisboa', 8, false];
 
-console.log(avg(mixedArr));
+//console.log(avg(mixedArr));
 
 // Iteration #5: Unique arrays
 const wordsUnique = [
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,86 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(a, b) {
+  return a > b ? a : b;
```
tomasyaya, line 3: awesome!

src/functions-and-arrays.js
```diff
@@ -1,41 +1,86 @@
...
 
-function findLongestWord() {}
-
-
+function findLongestWord(words) {
+  let longestLength = 0;
+  let longestWord = '';
+  if (words.length === 0) return null;
+  for (let word of words) {
+    if (word.length > longestLength) {
```
tomasyaya, line 14: I think it may be a bit redundant to save both the length and word of the longest word.

If you already have the longest word you can access its length at any time at `longestWord.length`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,86 @@
...
+    sumOfLength += word.length;
+  }
+  return sumOfLength / array.length;
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(array) {
+  if (array.length === 0) return null;
+  return Number((sum(array) / array.length).toPrecision(3));
```
tomasyaya, line 78: very nice

tomasyaya (commented):
Impressive work @brigihorvath!

PR #2576, opened 2021-09-18
WDPT-REMOTE-SEP-Michael-Seitz

Gk1mtd:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,168 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(numA, numB) {
+  let largestNumber
```
tomasyaya, line 3: You could do

`return Math.max(numA, numB)`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,168 @@
...
 
 // Iteration #3.1 Bonus:
-function sum() {}
+function sum(numbers) {
+  if (numbers.length === 0) { //returns 0 if array is empty
+    return 0
+  }
+
+  for (element of numbers) {
+    if (typeof element === "object")
```
tomasyaya, line 65: if you use a different loop just for the throw path then you are going to loop the array twice, you could add it to the other loop

tomasyaya (commented):
Great work with this LAB @Gk1mtd

src/functions-and-arrays.js
Gk1mtd: :D that will come in handy! Thanks.

PR #2578, opened 2021-09-21
WDPT-REMOTE-SEP-2021-Lilli-Reilly

l-reilly:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,63 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
```
tomasyaya, line 3: here we can directly do

return Math.max(num1, num2)

Math is a javascript object with functions, max is one of those functions

here is the documentation

[math max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

src/functions-and-arrays.js
```diff
@@ -1,38 +1,63 @@
...
+
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
-
+function averageNumbers(numbersAvg) {
+  let total = 0
```
tomasyaya, line 45: remember you can re-use the sum function to calculate the total and the do the avg

tomasyaya (commented):
Awesome job here @l-reilly

PR #2579, opened 2021-09-21
WD_PTSET2021_JOANABRITO

jobritomar:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,141 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+
```
tomasyaya, line 3: here we can directly do

`return Math.max(num1, num2)`

Math is a javascript object with functions, max is one of those functions

here is the documentation

[math max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

src/functions-and-arrays.js
```diff
@@ -1,41 +1,141 @@
...
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers(numArr) {
+  let sum = 0
+
+if(numArr.length===0){
+  sum=0
```
tomasyaya, line 40: I think the LAB asked you to return `null` if the condition is met

here you are reassigning the variable sum to the value 0 which is already the value assigned to sum

src/functions-and-arrays.js
```diff
@@ -1,41 +1,141 @@
...
+    
+    switch ( typeof numSum ) {
+      case 'number' :
+        sumB+=numSum
+        break
+      case 'string' :
+        sumB+=numSum.length
+        break
+
+        case 'boolean':
```
tomasyaya, line 67: great use of switch !!

remember javascript will turn a boolean into a number if you try to sum it up so you could do

`sumB += numSum`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,141 @@
...
+
+function averageNumbers(averageArr) {
+
+  let averageTotal = 0
+
+  if (averageArr.length === 0) {
+    return null
+  }
+
+  averageTotal = sumNumbers(averageArr) / averageArr.length
```
tomasyaya, line 97: Great re-use of the function !!

src/functions-and-arrays.js
```diff
@@ -1,41 +1,141 @@
...
+  var stringAv = 0
+
+  if (stringArray.length === 0) {
+    return null
+  }
+
+  for( let stringNumber of stringArray) {
+
+  stringSum += stringNumber.length 
+  stringAv = stringSum / stringArray.length
```
tomasyaya, line 119: the average should be calculated only once when the loop Is done

tomasyaya (commented):
Great work @jobritomar !!!

PR #2580, opened 2021-09-21
Imane Aliker WD PT 14 Sep

phimanim:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,95 @@
...
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(words) {
+  if (words.length === 0) { return null }
+  let longestWord = words.sort(function(a, b) { return b.length - a.length; });
```
tomasyaya, line 18: I don't have a problem and actually encourage anybody to go find information online on how to tackle any specific problem.

The only thing I will strongly suggest is for you to understand what the code is doing.

If what `words.sort` is doing is clear to you, then awesome! but it would be a pity if that was not the case.

Remember that we are not testing this LABS, and I do not pretend they are fully completed.

Please if you need any assistance let me know!

src/functions-and-arrays.js
```diff
@@ -1,41 +1,95 @@
...
 
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(isArray) { 
+  let arrayOfLength = [];
+  if (isArray.length === 0) { return null }
+
+
```
tomasyaya, line 70: I think in this case the first loop is not necessary, right?

Basically, you are looping the array and building one equal to the one you are looping.

Maybe you could directly use the `isArray` variable

src/functions-and-arrays.js
```diff
@@ -1,41 +1,95 @@
...
+  
+   arrayOfLength.push(isArray[i].length);
+   }
+
+  let sum = 0;
+  let total;
+
+  for (let i = 0; i < arrayOfLength.length; i++) {
+    sum += arrayOfLength[i];
+    total = sum / arrayOfLength.length;
```
tomasyaya, line 81: `total = sum / arrayOfLength `

should go outside the for loop, because you should do the division when you have the total amount in sum

src/functions-and-arrays.js
```diff
@@ -52,14 +106,40 @@ const wordsUnique = [
...
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(arr, wrd) {
+  for (let i = 0; i < arr.length; i++) {
+    const found = arr[i];
+    
+    if ( found === wrd) {
+      return "true";
```
tomasyaya, line 125: I think you need to return a boolean, here you are returning the string "true". You want the boolean true

And likewise with the false

tomasyaya (commented):
Excellent job @phimanim !!

PR #2581, opened 2021-09-21
[FT-RMT-092021] DominikaLipka

dominikalipka:
Iteration 1 and 2 finished


dominikalipka:
as in commit, Iteration 3 finished, works in console but doesn't pass the tests, don't understand why.

dominikalipka:
Finished all 7 iterations without bonuses.

PR #2623, opened 2021-10-19
-m "sdsf"

mahrad92:


Farxa:
Hi Mahsa,
your code passed all the tests, good job!

Here are some comments to improve your work for the next times:

1. when you commit to GitHub, please write a good message to this commit. You wrote "sdsf". Later on, you will learn to do as many commits as possible and each commit should tell you what exactly you have just done to the code/ what you have committed. If it's your first commit and you really don't know what to write, you can write "initial commit" for example.

2. your naming for variables is unclear, if you want to be able to know what your code does, what the variables are and want to go back to your code after a while and understand it, add to it or change it, then it is very, very important to use good names or at least write comments to your code. This way you, and in the future, your teammates will be able to work better with your code.

3. I know the tests passed without having to declare “i” with “let”, but please try to make it a habit, that you use this syntax for your loop: 
for ([initialization]; [condition];[final-expression])

4. Another important piece of advice: spacing and indentation should be consistent throughout your code.

PR #2693, opened 2021-11-06
Joao & Diogo PT_WD_Nov_21

diogoascarneiro:
every exercise done, except bonus

### DIFF
--- a/iididi
+++ b/iididi
@@ -1,0 +1,511 @@
+[33mcommit 6e60c305605181f336adff0ca51fe5e00c0029fc[m[33m ([m[1;36mHEAD -> [m[1;32mmaster[m[33m)[m
+Author: Diogo <diogoascarneiro@hotmail.com>
+Date:   Sat Nov 6 15:45:55 2021 +0000
+
+    third and fourth iteration done
+
+[33mcommit d24a4df7d0c910352e46ce28b87392b074f4afde[m
+Author: Diogo <diogoascarneiro@hotmail.com>
+Date:   Sat Nov 6 15:29:10 2021 +0000
+
+    iteration 2 done
+
+[33mcommit 45c84900b38598ac3a7d0e146e0bff9a5a0a4e62[m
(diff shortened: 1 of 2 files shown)


carlarsmendes:
Hi guys! Great job. I hope this lab was useful to clear up some questions you could have this week.
You have a very good clear style of coding, so congrats.
About the bonus sumArray and using the "throw" -> Usually the bonus will ask you some things extra - outside of what was taught :) So well done for getting there. Saying "should throw an error" as test description was a "clue".
I leave you below a solution without the switch and a bit more simplified than yours, which is also correct: 

function sumArray(someArr) {
  if (!someArr) return null;
  let sum = 0;
  for (let el of someArr) {
    // we are adding "typeof" condition to reuse this function
    // to calculate sum of letters in the array of words

    // we can check the type of elements in the array:
    // console.log(`Type of el: ${typeof el}`);

    if (typeof el === 'string') sum += el.length;
    else sum += el;
  }
  return sum;
}

------------------------------------------------------------------
On the last bonus, I send you here another solution so you can check with yours 

  let product = 0;

  const numberOfRows = matrix.length;
  const numberOfColumns = matrix[0].length;

  // console.log(numberOfRows);
  // console.log(numberOfColumns)

  for (let i = 0; i < numberOfRows; i++) {
    let row = matrix[i];
    for (let e = 0; e < numberOfColumns - 3; e++) {
      // 0 - 1 - 2 - 3
      // 1 - 2 - 3 - 4
      let rowProduct = row[e] * row[e + 1] * row[e + 2] * row[e + 3];
      if (rowProduct > product) {
        product = rowProduct;
      }
    }
  }

  for (let i = 0; i < numberOfColumns; i++) {
    for (let e = 0; e < numberOfRows - 3; e++) {
      let columnProduct =
        matrix[e][i] * matrix[e + 1][i] * matrix[e + 2][i] * matrix[e + 3][i];
      if (columnProduct > product) {
        product = columnProduct;
      }
    }
  }

  return product;
};

PR #2694, opened 2021-11-06
WDPT Rach Corentin Solved Lab

CorentinAesch:


carlarsmendes:
My favorite thing is that you created the checkIfEmpty() function which applies what we've been talking about perfectly. Great!! 5*, great job!

**Note:**
All the exercises are great, but I would've preferred to see, for example:
` if (arr.length === 0) return true;`
instead of this:
`if(!arr.length) return true;`

They are both correct, but the second one is a bit more advanced and we're still going through basic conditionals and booleans and such.
Of course it's great if @CorentinAesch , in the end, this way of writing was super clear for both of you :)

PR #2697, opened 2021-11-08
[MADPT1121] Calvin Turner - Failed Matrix

Calvin087:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,123 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
```
Cdelatorre, line 3: cool!, another solution ->  `return num1 > num2 ? num1 : num2`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,123 @@
...
+  } else return num2;
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(words) {
+  if (words.length === 0) return null;
+  if (words.length === 1) return words[0];
```
Cdelatorre, line 13: it's not necessary ;), with your returned result you are covering this solution

src/functions-and-arrays.js
```diff
@@ -1,41 +1,123 @@
...
 
 // Iteration #3.1 Bonus:
-function sum() {}
-
-
+function sum(numbers) {
+  let error = new Error("Unsupported data type sir or ma'am");
+
+  if (numbers.length === 0) return 0;
+  if (numbers.length === 1) return numbers[0];
```
Cdelatorre, line 43: it's not necessary, in fact, this function could fail, if the array has only a single element like ["miami"], in this case you will return "miami" witch is not a valid output

src/functions-and-arrays.js
```diff
@@ -1,41 +1,123 @@
...
+    if (typeof numbers[i] === 'string') {
+      sentence += numbers[i];
+    } else if (typeof numbers[i] === 'number') {
+      count += numbers[i];
+    } else if (numbers[i] === true) {
+      count += 1;
+    } else if (numbers[i] === false) {
+      count += 0;
+    } else {
+      throw new Error("Unsupported data type sir or ma'am");
```
Cdelatorre, line 62: you can add to count variable the word, adding its `.length` ;)

Cdelatorre: so sentence variable could be deleted

src/functions-and-arrays.js
```diff
@@ -1,41 +1,123 @@
...
+
+  return count / numbers.length;
+}
 
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(allWords) {
+  if (allWords.length === 1) return allWords[0].length;
```
Cdelatorre, line 88: it's not necessary, with your `for` loop this case will be covered

src/functions-and-arrays.js
```diff
@@ -1,41 +1,123 @@
...
+  for (let i = 0; i < allWords.length; i++) {
+    count += allWords[i].length;
+  }
+
+  return count / allWords.length;
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(numbers) {
```
Cdelatorre, line 100: same as 3.1 Bonus ;) despite of it's totally correct!

src/functions-and-arrays.js
```diff
@@ -52,16 +134,25 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
-
-
+function uniquifyArray(myArray) {
+  if (myArray.length === 0) return null;
+  let clean = [...new Set(myArray)];
```
Cdelatorre, line 139: 🥇

Cdelatorre (approved):
Perfect Calvin, i wrote down some comments, just for improove your skills, but it's perfect

PR #2698, opened 2021-11-08
solved lab

andregn26:
We had some dificulties in the last two exercises. Mainly in the matter of counting and identifying the repeated word. We also had difficulty completing the bonus exercises

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,60 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  } else if (a < b) {
+    return b;
+  } else if (a === b) {
+    return a;
+  } else {
+    return "Something went wrong"
+  }
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+
+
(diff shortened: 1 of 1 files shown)


carlarsmendes:
On the bonus sum, you would need to add some checks for type, for example: Do something different when this array element is `typeof === "string" `
About the bonuses, don't worry - They were created for more advanced students who maybe need an extra layer of difficulty, but you will get there eventually on module 2. I see all your tests passed. That was your goal! Congrats

The final bonus in particular is quite hard, involves nested arrays which we hadn't went through in depth. Here is a good solution. I send it to you below, but don't worry if you don't fully understand it yet, it will come with time and practice :) 

```
const greatestProduct = matrix => {
  let product = 0;

  const numberOfRows = matrix.length;
  const numberOfColumns = matrix[0].length;

  // console.log(numberOfRows);
  // console.log(numberOfColumns)

  for (let i = 0; i < numberOfRows; i++) {
    let row = matrix[i];
    for (let e = 0; e < numberOfColumns - 3; e++) {
      // 0 - 1 - 2 - 3
      // 1 - 2 - 3 - 4
      let rowProduct = row[e] * row[e + 1] * row[e + 2] * row[e + 3];
      if (rowProduct > product) {
        product = rowProduct;
      }
    }
  }

  for (let i = 0; i < numberOfColumns; i++) {
    for (let e = 0; e < numberOfRows - 3; e++) {
      let columnProduct =
        matrix[e][i] * matrix[e + 1][i] * matrix[e + 2][i] * matrix[e + 3][i];
      if (columnProduct > product) {
        product = columnProduct;
      }
    }
  }

  return product;
};
```

PR #2702, opened 2021-11-09
WDPT NOV21 Michael & Vitor 

micfol:


carlarsmendes:
Hi guys, these look like good, clean solutions :) I like the naming and the solutions you found
One note: (I think this is for the LongestWord function
```
else if (someArr.length === 1 && someArr.includes(word)) {
    return true;
  } else if (someArr.length >= 1 && someArr.includes(word)) {
    return true;
```

If we are checking for === 1 on one condition,
And then we are checking for >=1  and they both have the same outcome, maybe the ===1 is not necessary?
Overall great job on these solutions, let me know if you have specific questions remaining!

carlarsmendes:
So the solution would be:
```
function doesWordExist(someArr, word) {
  if (someArr.length === 0) {
    return null;
  } else if (someArr.length >= 1 && someArr.includes(word)) {
    return true;
  } else {
    return false;
}
```

PR #2758, opened 2022-01-11
[REMOTE WDFT 012022] Denise Sengül

denisesenguel:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,130 @@
...
+    console.log("Please pass a non-empty array as function argument.")
+    return null;
+  }
+
+  let longestWord = arrayOfWords[0];
+  for (let i = 1; i < arrayOfWords.length; i++) {
+    if (arrayOfWords[i].length > longestWord.length) {
+      longestWord = arrayOfWords[i];
+    } else {
+      continue;
```
TA-Remote, line 27: `else {continue}` is not neccessary in this case. You have a `for loop `which will last until you reach end of your array -> until `i < arrayOfWords.length;`. And only then you will exit the loop

src/functions-and-arrays.js
```diff
@@ -1,41 +1,130 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arrayOfNumbers) {
+
+  if (arrayOfNumbers.length > 0) {
+    let average = sumNumbers(arrayOfNumbers) / arrayOfNumbers.length;
+    return average;
```
TA-Remote, line 95: Here, to reduce your code even more, you can directly return the result, without declaring new variable :) 

`return sumNumbers(arrayOfNumbers) / arrayOfNumbers.length;`

src/functions-and-arrays.js
```diff
@@ -52,14 +141,39 @@ const wordsUnique = [
...
+  if (inputArray.length > 0) {
+    
+    let uniqueArray = [];
+  
+    for (i = 0; i < inputArray.length; i++) {
+      if (uniqueArray.includes(inputArray[i])) {
+        continue;
+      } else {
+        uniqueArray.push(inputArray[i]);
+      }
```
TA-Remote, line 158: Here instead of using both `if` and `else`, it is possible to negate a condition by adding exclamation point before the condition. for example:

` if (!uniqueArray.includes(inputArray[i])) {
        uniqueArray.push(inputArray[i]);
   }`

This says, if array does not include the element -> push it into array

TA-Remote (commented):
Some great code here :muscle:

PR #2761, opened 2022-01-11
[Ber FT 012022] Ghalia Al Hasan

Ghalia98:
Tried to solve the unique array iteration with indexOf() and lastIndexOf(), but got lost a bit. Can be it be solved using those two methods?

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -5,7 +5,7 @@ function maxOfTwoNumbers(num1, num2) {
   }
   return num2
 }
-
+//  could be solved with Math.max^
 
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


stephen-gallagher:
Nice work Ghalia! It's clear you have a strong understanding of key JS concepts.

Just a little bit of feedback regarding bonus 3.1 - Javascript will already convert booleans to their numerical value (true = 1, false = 0) so you don't need to add these conditions within your loop. The code you've written is 100% correct, but could be shortened to: 

     if (typeof arr[i] === 'object') {
    } else if (typeof arr[i] === 'string') {
      totalSum += arr[i].length;
    } else {
      totalSum += arr[i];
    }
     return totalSum;

Ghalia98:
> Nice work Ghalia! It's clear you have a strong understanding of key JS concepts.
> 
> Just a little bit of feedback regarding bonus 3.1 - Javascript will already convert booleans to their numerical value (true = 1, false = 0) so you don't need to add these conditions within your loop. The code you've written is 100% correct, but could be shortened to:
> 
> for (let i = 0; i < arr.length; i++) { if (typeof arr[i] === 'object') { throw new Error("Unsupported data type sir or ma'am"); } else if (typeof arr[i] === 'string') { totalSum += arr[i].length; } else { totalSum += arr[i]; } return totalSum;

Thank you for your feedback Stephen!

PR #2763, opened 2022-01-11
final push

yoricktf:
failing three tests, passing it on runJS though

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,66 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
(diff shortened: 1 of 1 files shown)


stephen-gallagher:
Nice Yorick! It's clear you have a very strong understanding of key JS concepts.

I wouldn't worry about the failed test. Your solution is correct but if we look at the expected/received component of the test we can see that the result needs to be to 2 decimal places.

Expected: 5.11
Received: 5.111111111111111

this can be achieved with the toFixed(2) method, where the value inside the brackets determines the number of places to set the decimal to.

yoricktf:
Thanks for the feedback Stephan, when we went over it the next day I
realized that’s what I had done wrong.  In the one example is was testing
with I was getting the expected output so was very confused at that time

PR #2765, opened 2022-01-11
Finished Lab 

jcekelly:
I was only able to complete up until iteration #4 today unfortunately, but I got some very good practice from it. 

I had some trouble with running the tests - my function for iteration 3 was passing the tests, and then began to fail them, but I did not knowingly change the code and the function runs correctly in runJS, so I was unable to find and fix the issue there - though I assume it is a syntax error. Thanks!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,20 +1,39 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a,b) {
+  if (a > b){
+    return a
+  } else {
+    return b
+  }
+}
 
 
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


stephen-gallagher:
Hey Jan, nice work so far! 

 A good way to check why your tests aren't passing is to check the 'expected: received:' component of the lab solution file (the file that you open with live server). For iteration 3 we can see the following on your failed test:

Expected: 4
Received: [4]

This means that the test is expecting a number but is instead receiving a number inside of an array. 

You can actually simplify your code a little bit - there's no need to create an empty array and then push the total into it, you only need to return the sum variable. It would look something like the following:

function sumNumbers(arr) {
  let sum = 0;
    sum += arr[i];
  }
  return sum;
}

PR #2766, opened 2022-01-11
Solved Lab

YoungTaubi:
Hey there,
finished all tasks except the bonuses...


stephen-gallagher:
Nice Jonas! Your submitted solutions all look great and don't need any feedback - would be good to have a try at the bonus questions if you have the time :)

PR #2767, opened 2022-01-11
Marian Gispert WDFT-JAN 2022

mgispert:
Kinda struggled with certain things, but it's all in the comments.
Couldn't even try the matrix bonus, nope.


TA-Remote (commented):
Amazing work, really! I see you are mastering forEach s well :)  I have no remarks unless you want to address something specifically. Keep up :muscle:

PR #2768, opened 2022-01-11
[WDFT-Jan 22] Day 2 Labs solved - Doan Tran

Dorrrn:


src/functions-and-arrays.js
```diff
@@ -52,16 +102,32 @@ const wordsUnique = [
...
 ];
 
-function uniquifyArray() {}
-
+function uniquifyArray(array6) {
+  let newArray = [];
+  for (let i = 0; i < array6.length; i++) {
+    if (array6[i].indexOf === array6[i].indexOf) {
+      newArray = newArray.push;
+    }
```
TA-Remote, line 110: -  I'd advise to go back and read about `indexOf()` a little bit. The syntax of this method looks like this `arrayName.indexOf(searchElement)`.
-  you want to access array, then call the indexOf method and  pass an element you are checking as argument 
- `indexOf()` method return and index of an element, or if it is not existant than it return `-1`

TA-Remote (commented):
Good effort on the lab! Any questions about the comments let me know. We can address it in person! Karina :wave:

src/functions-and-arrays.js
Dorrrn: Thanks for reviewing and detailed feedback! Helps a lot! <3 
Doan

PR #2770, opened 2022-01-11
Solved lab first approach

Jan-000:
This is first try. To be improved.

Questions:

- how to correctly declare an empty array?
- please elaborate about tricking tests with apparently correct solutions (which are in fact wrong)

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,36 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(a,b) {
+  if (a > b) {
+    return a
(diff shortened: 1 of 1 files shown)


stephen-gallagher:
Hey Jan! 

If you want to add a conditional statement to check if an array is empty, I think the best way would be to to check if the length of the array is equal to zero. It would look a bit like the following:

`if(array.length === 0) {
      return null
 }
`

I'm not too sure what you mean about tricking the tests - if you open the _functions-and-arrays-spec.js_ file in the tests folder of the lab you can see what the tests are looking for in your code, I guess there might be a way to write code that fulfils the requirements of the test but doesn't provide the desired result but i'm not too sure with this.

Lemme know if you have any more questions :)

PR #2772, opened 2022-01-11
[BER WD FT Jan 2022]Patrick Doktorowski 

pdokto:
solved lab with bonus


stephen-gallagher:
Nice Patrick! It's clear you have a strong understanding of key JS concepts and kudos for solving the matrix bonus.

PR #2773, opened 2022-01-11
[Remote-FT-012022] Zena Bos

zenabos:
While working on iteration 6, I messed up. I couldn't make the testruns work anymore. Maybe it had to do something with my connection (it failed for a little while)? I undid all my actions, but nothing changed. 

So I'm just making a pull request now, in case I break even more. Learning: do a pull request at an early stage. ;)

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -149,7 +149,13 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes(wordsCount, wordToCount) {}
+function howManyTimes(wordsCount, wordToCount) {
+    let counter = 0;
+    for (let i = 0; i < wordsCount.length; i++)
+    if((wordsCount[i] == wordToCount)) 
+        counter += 1;
+    return counter;    
+  }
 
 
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -52,14 +101,36 @@ const wordsUnique = [
...
-function doesWordExist() {}
+function doesWordExist(wordsFind, wordToFind) {
+  for(let i = 0; 0 < wordsFind.length; i++){
+    if(wordsFind[i] === wordToFind) {
+      return "true";
+    } else {
+      return "false";
+    }
+  }
+}
```
TA-Remote, line 134: really good work, Zena! If you are interested, how to solve the bonus part, you can have a look at the solution.
For that function doesWordExist: instead of a string, you just send the boolean, like this:
```
if(wordsFind[i] === wordToFind) {
      return true;
```
otherwise it is only small things you have to change to have the test completely green. Such as:
```
 if (arrOfWords.length === 0) {
    return null;
  }
```
because the test says:
"should return null if receives an empty array when called"
But I think how you have to deal with the test you have understood very well.

PR #2774, opened 2022-01-11
 WDFT GJ finished most of the labs

gjschoutsen:
Had difficulty with the indexOf method.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,55 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  }else if(num1 < num2){
+    return num2;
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -52,14 +111,27 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(arr) {
+  let noDuplicates = [];
+   
+return noDuplicates;
+}
```
TA-Remote, line 118: Did you read about `indexOf() `and still not understand how to apply it?

- the syntax of this method looks like this arrayName.indexOf(searchElement).
- you want to access array, then call the indexOf method and pass an element you are checking as argument
- indexOf() method returns and index of an element (inside that array), or if it is not existant  -1

TA-Remote (commented):
Great effort! We can go over indexOf() in person if still not clear. Karina :wave:

PR #2775, opened 2022-01-11
WDFT Helena Saborit

Helsinky91:
submitting the lab, lots of things missing, i'm struggling a bit with functions.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,21 +1,49 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+if (num1 > num2) {
+  return num1;
+} else {
+  return num2
+}
+};
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arrWords) {
 
+  if (arrWords.length == 0){
+  return null;
+} else {
+  let longestWord = "";
+  for (let i = 0; i < arrWords.length; i++) {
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,21 +1,49 @@
...
+function sumNumbers(arrNumbers) {
+  let counter = 0;
+  if (arrNumbers.length == 0) {
+    return 0;
+  } else {
+  for (let i=0; i< arrNumbers.length; i++) {
+    return counter += arrNumbers[i];
+  }
+  }
+}
```
TA-Remote, line 45: - Same mistake here - your function does not return anything. 
- Any logic that is inside loop will be repeated each iteration. What you want is to update the variable. And only return it after you finish your full loop cycle

```
for (let i=0; i< arrNumbers.length; i++) {
    counter += arrNumbers[i];
  }
```

- add `return` at the end of else statement

src/functions-and-arrays.js
```diff
@@ -26,13 +54,36 @@ function sum() {}
...
 
-function averageWordLength() { }
+function averageWordLength(word) {
+  if (word.length == 0){
+    return null;
+  } else {
+    let counter = 0;
+    for (let i=0; i<word.length; i++){
+      return (counter += word[i].length) / word.length;
+    }
```
TA-Remote, line 82: - Similar issue here. You want to update counter inside `for loop`, but not return it
- At the moment you are taking each element of array and dividing by the length. However  you'd want to firstly get total sum and only then divide by the length of array

TA-Remote (commented):
It's all about the practice, it's okay to struggle! I have left some comments. Great effort so far, please review the comments and let us know if something is not clear, you want to address specific moment. 

Pay attention to comments about <return>. You are repeating this mistake in several occasions, where the loop returns something instead of the function returning the result

PR #2779, opened 2022-01-11
WD Remote Jan2021 - Ana Pavao

pavaoana:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,43 @@
...
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1<num2) {
+    return num2;
+  }
+  else if (num1>num2) {
+    return num1;
+  }
+  else if (num1=num2) {
+    return num1 || num2;
```
TA-Remote, line 10: - always keep your code as short as possible, meaning leave only neccesary code :) Soo, here you might as well write `return num1;`
- when you are doing comparison, always use `==` or `===`. Because with single` =`  you are asigning a new value to a variable

src/functions-and-arrays.js
```diff
@@ -26,16 +50,38 @@ function sum() {}
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  if (arr.length === 0) return null;
+  let sum = 0;
+  arr.forEach(function(element) {
+    sum = sum + element;
+  } );
+  let average = sum / arr.length;
```
TA-Remote, line 59: Here, can you think of another solution using the previous `function sumNumbers()` ? Instead of calculating sum again ;)

Remember, that `sumNumbers()` returns a sum. And we can use it here

src/functions-and-arrays.js
```diff
@@ -52,14 +98,34 @@ const wordsUnique = [
...
+  if (arr.length === 0) return null;
+  let uniqueArray = [];
+  let filteredArr = arr.filter(function(item, index) {
+  if (arr.indexOf(item) == index) {
+    uniqueArray.push(arr[index])
+  }
+    });
+    return uniqueArray;   
+}
+// I tried to do this in so many different ways (with for loops and for each loops) and it didn't work;
```
TA-Remote, line 111: - have you tried using `indexOf() ` as the task requires? in that case when you loop with `for ` or `forEach` you can be checking if item is already in the array 
- also there is `includes()` method that checks if item is inside an array

- We will learn filter method later this week, but well done - as long as you understand how it works!

TA-Remote (commented):
Overall great work!

PR #2784, opened 2022-01-11
 [FT-BC-WD-JAN 2022] Vinh Nguyen - Solved lab

vinntt:


stephen-gallagher:
Nice Vinh! It's clear to see that you have a very strong understanding of key Javascript concepts, nice work on the matrix bonus! :)

PR #2788, opened 2022-01-12
Solved lab

ola-kiitan:


stephen-gallagher:
Nice work Olayinka! 

Just a little tip, when you write `sum = sum + numbers[i];` this is completely correct, but it can be further shortened to `sum += numbers[i]`

PR #2790, opened 2022-01-12
[WD-FT-Berlin- Jan22] Mehmet Büyükköse

Lavapies:


stephen-gallagher:
Nice Mehmet! It's clear that you have a good understanding of key JS concepts. Would be good to have a try at the bonus questions if you get some free time.

PR #2929, opened 2022-03-29
WDFT MARC 2022 - RUBEN POELEN

coderKron:
did everything that was needed except for bonusses, might do them if im in the mood this evening.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,38 +1,104 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1 = 0, num2 = 0) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num1 < num2) {
+    return num2;
+  } else {
+    return num1;
+  }
+}
 
-
+const highestNumber = maxOfTwoNumbers(12, 26);
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  let word = '';
+  if (arr === undefined) {
+    return null;
+  } else if (arr.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i].length === 0) {
+      return null;
+    } else if (word.length < arr[i].length) {
+      word = arr[i];
+    } else if (word.length === arr[i].length) {
+      continue;
+    }
+  }
+  return word;
+}
 
+const longestWord = findLongestWord([]);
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,38 +1,104 @@
...
+    } else if (word.length === arr[i].length) {
+      continue;
+    }
+  }
+  return word;
+}
 
+const longestWord = findLongestWord([]);
 
+console.log(longestWord);
```
TA-Remote, line 38: After you finish writing code, try to always clean up and remove any lines that are unused ;)

src/functions-and-arrays.js
```diff
@@ -1,38 +1,104 @@
...
+function findLongestWord(arr) {
+  let word = '';
+  if (arr === undefined) {
+    return null;
+  } else if (arr.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i].length === 0) {
+      return null;
```
TA-Remote, line 26: what are you checking here?

src/functions-and-arrays.js
```diff
@@ -52,16 +118,45 @@ const wordsUnique = [
...
+function doesWordExist(arrOfWords, word) {
+  if (arrOfWords.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < arrOfWords.length; i++) {
+    if (arrOfWords[i].indexOf(word) !== -1) {
+      return true;
+    } else {
+      continue;
+    }
```
TA-Remote, line 153: with if statements you do not need to use `continue`, also in previous iterations.
You are inside the `for loop`, so you are already iterating through each element of the array , and only exit the loop if you find exisitng word in this case

src/functions-and-arrays.js
```diff
@@ -52,16 +118,45 @@ const wordsUnique = [
...
+  }
+  return noDuplicates;
+}
 
+const noDuplicates = uniquifyArray(wordsUnique);
+console.log(noDuplicates);
 
 // Iteration #6: Find elements
```
TA-Remote, line 141: Iteration 6 task says, ` Don't use indexOf for this one.`. Can you think of an alternative way to solve this ? ;)

TA-Remote (commented):
Great work Ruben, keep it up!

src/functions-and-arrays.js
coderKron: function doesWordExist(arrOfWords, word) {
  if (arrOfWords.length === 0) {
    return null;
  }
  for (let i = 0; i < arrOfWords.length; i++) {
    if (arrOfWords[i].includes(word) === true) {
      return true;
    }
  }
  return false;
}

coderKron: I read it wrong oops!

PR #2932, opened 2022-03-29
WDFT REMOTE MAR22 Fevi

fevi-ro:
I couldn't solve  Iteration #4 Level 2 with a function.
Iteration #5 was also a bit tricky.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,48 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+    if (a > b) {
+        console.log(a);
+        return a;
+    } else {
+        console.log(b);
+        return b;
+    }
+}
 
+maxOfTwoNumbers(6, 12)
 
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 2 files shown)


src/functions-and-arrays.js
```diff
@@ -26,84 +55,122 @@ function sum() {}
...
 
-function averageNumbers() {}
+function averageNumbers(numbersAvg) {
+    if (!numbersAvg.length) return null;
+    let num = 0;
+    for (let number of numbersAvg) {
+        num += number;
+    }
+    return num / numbersAvg.length;
+};
```
TA-Remote, line 66: Really good work, Fevi!

src/functions-and-arrays.js
```diff
@@ -26,84 +55,122 @@ function sum() {}
...
+console.log(uniqueNames);
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(wordsFind, word) {
+    if (wordsFind.length === 0) return null;
+    for (let el of wordsFind) {
+        if (el === 'disobedience') return true;
```
TA-Remote, line 116: But one more thing: you never know which words you really have in the array. So you can't compare with specific words. It should work for any array of words.

PR #2936, opened 2022-03-29
WD FT Remote March 28 Alexander Nerz

Alex-WD-22:
I had some difficulties finishing it.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,53 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(numb1, numb2) {
+    if (numb1 < numb2) {
+    return numb2;
+  } else if (numb1 > numb2) {
+    return numb1;
+  
+  }
+}
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  let longestWord = '';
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,34 +1,71 @@
...
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(numb1, numb2) {
+  if (numb1 < numb2) {
+    return numb2;
+} else if (numb1 > numb2) {
+    return numb1;
+  
+  } 
+}
```
TA-Remote, line 10: Hi Alex! I won't go into all the failed tests so much now, you can look at the solutions in the Student Portal again to see exactly what was asked for. But still a few notes:
for example in this iteration the test was asking for returning any number if the numbers are equal. You could have solved it like this:
```
function maxOfTwoNumbers(numb1, numb2) {
  if (numb1 < numb2) {
    return numb2;
  } else if (numb1 > numb2) {
    return numb1;
  }
  else {
    return numb1
  }
}

```

src/functions-and-arrays.js
```diff
@@ -1,34 +1,71 @@
...
+  let longestWord = '';
+  for (let i = 0; i < arr.length; i++) {
+    if (arr.length === 0) {
+      return null;
+    } else if (arr[i].length > longestWord.length) {
+      longestWord = arr[i];
+    }
+  }
+  return longestWord;
+}
```
TA-Remote, line 26: here it's just a small thing, the condition to return null needs to be outside of the for loop:
```
  let longestWord = '';
  if (arr.length === 0) {
    return null;
  }
      longestWord = arr[i];
    }
  }
  return longestWord;
}

```

PR #2937, opened 2022-03-29
Valentin Paul [WD MAR22 FT]

Valentin-Paul:
Couldnt get the: result 'null' if array is not defined to work...

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,50 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2) {
+    return number1;
+  } else if (number2 > number1) {
+    return number2;
+  } else if (number1 === number2) {
+    return number1;
+  }
+}
+
+
 
 
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,50 @@
...
+  let longestWord = '';
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i].length > longestWord.length) {
+      longestWord = arr[i];
+    } else if (arr[undefined]) {
+      longestWord = null;
+    }
+  }
+  return longestWord;
+}
```
TA-Remote, line 30: In most iterations you are asked to check if given array is valid - if it is not empty. Because if you receive an empty array there is no need to proceed with the function, to calculate sum , average etc. 

Here you have else statement to check that, but it will not work, because the code will stop on  line 22. Your for loop condition says to iterate while` i` is less than `array length`. And your array length is 0 already if it is empty ;) , so we will never go inside the loop!

One way to do it is to check if given array length is zero, and in that case you want to return `null` (which will 'exit' the function)

```
if (arr.length < 0) {
    return null
}
```

TA-Remote (commented):
Great work on the lab! Try to find alternative way to check if given array is empty or not. I made a suggestion :)

PR #2940, opened 2022-03-29
Solved lab

infjay:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,60 @@
...
+if (words.length == 0){
+  return null;
+} else if ( words.length == 1) {
+  return words[0];
+} else {
+  let longestword = "";
+  words.forEach(word => {
+   if (word.length > longestword.length) {
+   longestword = word;
+   return longestword;
```
TA-Remote, line 28: you can delete this return in the if condition

TA-Remote:
Great, Jawid! Only small things that could be changed, really good job!

PR #2941, opened 2022-03-29
WDFT 22 MARCH Diana 

didicozaur:


TA-Remote:
Hi Diana! All these green fields in the tests, love it haha! But still some things to improve, which I will write in the following comments

src/functions-and-arrays.js
```diff
@@ -1,38 +1,72 @@
...
+  });
+  return result / numbersAvg.length;
+}
 
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(wordsArr) {
+  if (words.length === 0) {
```
TA-Remote, line 61: here it was just a typo, instead of words.length, you need to use wordsArr.length because you called the parameter like that. So, like this:
```
if (wordsArr.length === 0) {
    return null;
  }
```

src/functions-and-arrays.js
```diff
@@ -52,16 +86,35 @@ const wordsUnique = [
...
 
-function uniquifyArray() {}
-
+function uniquifyArray(wordsUnique) {
+  if (wordsUnique.length === 0) {
+    return null;
+  }
 
+  wordsUnique.forEach((word) => {});
+}
```
TA-Remote, line 95: one possible solution would be like this:
```
function uniquifyArray(wordsUnique) {
  if (wordsUnique.length === 0) {
    return null;
  }

  let unique = []
  wordsUnique.forEach((word) => {
    if (unique.indexOf(word) < 0){
      unique.push(word)
    }
  });
  return unique
}
```

TA-Remote: If you have any questions about this task or the bonus tasks, please feel free to ask me (Joanne) or Marian and Karina

PR #2942, opened 2022-03-29
[WDFT032022] Ivo Silva

Ivo-S-Silva:
Completed all the questions, including bonus. Left some comments on code especially on the last questions, since I had to google for support and had to adapt code I found in C ++ to work on Javascript.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,86 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2) {
+    return number1;
+  } else if (number2 > number1) {
+    return number2;
+  } else if (number1 === number2) {
+    return number1;
+  }
+}
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arrayWords) {
+  const numOfLetters = [];
+  let indexOfMax;
+  let maxNumOfLetters;
+
+  if (arrayWords.length === 0) {
+    return null;
+  } else if (arrayWords.length === 1) {
+    return arrayWords[0];
+  } else {
+    // Created a new array with the word sizes in the same order of the original words array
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,141 @@
...
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2) {
+    return number1;
+  } else if (number2 > number1) {
+    return number2;
+  } else if (number1 === number2) {
+    return number1;
+  }
```
TA-Remote, line 9: Since you clearly understand your own code, some tips for you to refactor ;)

since we only check the length, below code would do the same:
```
  if (number1 >= number2) {
    return number1;
  } else if (number2 > number1) {
    return number2;
  }
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,141 @@
...
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arrayWords) {
+  const numOfLetters = [];
+  let indexOfMax;
+  let maxNumOfLetters;
+
+  if (arrayWords.length === 0) {
+    return null;
```
TA-Remote, line 22: just a recommendation, it would be even easier to read your code if you check array's validity first. And only then start another if statement:

```
if (arrayWords.length === 0) return null;

if (arrayWords.length === 1) {
  // remaining logic here
}
```

If you write it like this, you very clearly separate the logic. If array s not valid i just exit the function. If valid - i carry on. Also if you noticed, if my return is only one line I do not need curly braces!

And even more advanced syntax could be this:
`if (!arrayWords.length) return null;`

Here, if `arrayWords.length = 0` it will return false (in js zero is falsy), so we negate it with explamation point to get desired result.

TA-Remote (commented):
Amazing work Ivo, Keep it up! 
I left some random notes to improve please tell me if not clear :D 

Well done on figuring out the solutions, it is every day thing to google things so great for being able to apply those. In some places code can be refactored, i advise looking at the solution we uploaded on student portal to compare alternative approaches - and more simple I have to say ;) 

Also nice one on using forEach loop ! Some advanced code here

K.

PR #2943, opened 2022-03-29
WDFTMARCH2022+Sepideh.att

Sepidehatt:


src/functions-and-arrays.js
```diff
@@ -52,14 +163,44 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(wordsUnique) {
+if (!wordsUnique.length) {
+   return null;
+}else {
+  UniqueArr = [];
```
TA-Remote, line 170: As mentioned before always declare variable with const or let. Also use start your variable name with lowerCase. Not an error but a convention :)

TA-Remote (commented):
Keep it up Sepid! Amazing effort on the assignment, i left some notes for you, things to improve. Your code could be a bit simplier/shorter but that comes with more practice ;)

We will upload solution on student portal, so have a look at other approaches :)

PR #2946, opened 2022-03-30
WDFT Remote Lab Function & Array March29 - Ha Thu TRAN

HaThu6989:
Hi, 

Except Bonus - Iteration #8: Product of adjacent numbers, i can't do that.
For the rest, I tried my best but i don't know why there are some functions that worked fine but when I used Testing with Jest, it didn't work.

Best regards,

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,94 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hi Thu! It's already late, so I will not go into detail today.
If you understood everything you coded, wow! Great work. Then you don't have to worry.

The only thing I noticed: sometimes it could be a bit less code, but you can see the solutions in the Student Portal and compare!

PR #2949, opened 2022-03-30
4 labs tried to solve :/

sina-kaabi:
hard work to complete all 8!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,38 +1,127 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+
+
+function maxOfTwoNumbers(harrySize, bobSize) {
+  if (harrySize < bobSize) {
+    return bobSize;
+  } 
+  
+  else if (bobSize < harrySize) {
+    return harrySize;
+   }
+
+   else  {
+     return harrySize;
+   }
+  
+};
+
+let harrySize = 7;
+let bobSize = 5;
+
+console.log(maxOfTwoNumbers(harrySize, bobSize));
+
 
 
 
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+//const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+/*findLongestWord("mystery brother aviator crocodile pearl orchard crackpot"); */
+
+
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,38 +1,127 @@
...
+   else  {
+     return harrySize;
+   }
+  
+};
+
+let harrySize = 7;
+let bobSize = 5;
+
+console.log(maxOfTwoNumbers(harrySize, bobSize));
```
TA-Remote, line 22: when you finish writing your code, always try to cleanup and delete anything that is unused ;)

src/functions-and-arrays.js
```diff
@@ -1,38 +1,127 @@
...
+ 
+  for(var i = 0; i < myArray.length; i++){
+    if(myArray[i].length > longestWord){
+	longestWord = myArray[i].length;
+     }
+  }
+  return longestWord;
+}
+
+console.log(longestStringForLoop(myArray));
```
TA-Remote, line 57: Here you have changed the name of the function - so the tests cannot find and check it! This week more labs will have automated tests, so be sure not to touch the function ma,e only write code inside the function scope :)

Also, your longestWord has to be a string, the actual word. Currently, on line 51, your saving (and later returning) the length of that string.

src/functions-and-arrays.js
```diff
@@ -1,38 +1,127 @@
...
 
-function sumNumbers() {}
+let sum = 0;
+
+  for (let i = 0; i < numbers.length; i++) {
+    sum += numbers[i];
+  }
+  
+
+console.log(sum);
```
TA-Remote, line 75: In iteration 3, your function disappeared... you have some logic to calculate sum , but no function itself..

And once you have your logic inside the function, do not forget to `return` the result ;)

src/functions-and-arrays.js
```diff
@@ -1,38 +1,127 @@
...
+var sum = 0;
+for (var number of numbersAvg) {
+    sum += number;
+}
+
+
+
+average = sum / numbersAvg.length;
+
+console.log(average);
```
TA-Remote, line 100: in iteration4:
- you have deleted the function again.. just left the logic. Please keep all the code that is given to you in exercise!
- try to avoid using` var`, in modern javascript you will rarely see it. Also it has a bit different scope than let so you might run into issues with more complex code
- in js we must declare every variable that we use (like you did with `sum`). your variable `average` was never declared..

TA-Remote (commented):
I made some notes for you here. My recommendation is to try and understand the comments and see if you can improve the code. In next labs be sure to ask the class or join Gabriel's session so he can help you with concepts that you do not understand. Keep working! :)

PR #2967, opened 2022-04-12
[WDFT042022] Dino Marchiori

dinaco:


src/functions-and-arrays.js
```diff
@@ -1,42 +1,89 @@
...
+  return a > b ? a : b;
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
-
+function findLongestWord(words) {
```
webLisbonTA, line 9: Great solution! Two comments here:

- A good practice is to do your parameter checking before the array loop, if the array is empty you don't the program even to thing about looping through it

- Also, you don't need to split('') the element -- you are already inside a loop and you have access to each element individually.

webLisbonTA (commented):
Well done, Dino!
Overall, very good implementation of ternary operator and array methods!

PR #2970, opened 2022-04-12
[FTLIS042022] João Gonçalves

joaopdg:
All done. Bonus 8 found the solution on google, don't really know how to do it


webLisbonTA (commented):
Great job on finishing everything, congrats! A quick tip: try and space and organize your code a little bit better so it becomes more readable.

PR #2971, opened 2022-04-12
[FTLIS042022] Duarte Alves 

drttyy:


webLisbonTA (commented):
Well Done, Duarte!

A hint for Bonus - Iteration #4.1
You could reuse your mixedArray sum function here to calculate the avg of mixed array

PR #2976, opened 2022-04-12
[FTLIS042022] Mariana Oliveira

miloliveira:
All done, except for the Bonus 8 exercise.


webLisbonTA (commented):
Great job on doing every mandatory iteration and most of the bonus!

PR #2980, opened 2022-04-12
[FTLIS042022] Catarina Salvado

catarinapsalvado:


src/functions-and-arrays.js
```diff
@@ -78,8 +155,17 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
+function howManyTimes(arr, word) {
```
webLisbonTA, line 158: Take a further look on this function, it is not passing the tests yet.
The logic is good, but there are 2 things you need to change to make it work

- if you do `sum += sum[i]`your are adding the string element on the sum variable, what you want instead is to `sum += 1` every time a word is found.
- The function return is inside the for loop, it should be outside at the end of your function

webLisbonTA (commented):
Well Done, Catarina!

A hint for iterations 3.1 and 4.1 - Bonus with the mixed array: --> try to implement some conditions to check the type of the element on the array loop, you can do that either with na if statement or with a switch -- and sum the value in a different way according with the type. Example:

- If type is string - sum the length of the string
- If type is number - sum the value
- If type is boolean - sum 1 if it is true and 0 if it is false

Also, remember that you can later reuse the sum mixed array function on inside your avg function.

PR #2981, opened 2022-04-12
[FTLIS042022] Leticia M.

Pip0kinha:
Hey =) I managed to solve exercises from 1-7 but I didn't manage to eliminate the error messages about empty array and unsupported data type of Bonus exercises 3.1 and 4.1 although i believe I did include those conditions in my code. Would love to know what I got wrong on those two exercises! 
Did not have enough time to try exercise 8.


webLisbonTA (commented):
Great job on doing most of the bonus and all the mandatory iterations! A quick tip: research the term "throw error" for a better understanding of the bonus :)

PR #2982, opened 2022-04-12
[FTLIS042022] anaPinho

AnaGonPinho:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,48 @@
...
+
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
 
+function findLongestWord(arr) {
```
webLisbonTA, line 18: The logic here is perfect, but it is not passing the tests, because your parameter is called arr, so you need to do all the code related to that, for in your if statement if needs to compare 
```
      longestWord = arr[i]
        }
```

instead of 

```
      longestWord = words[i]
```

src/functions-and-arrays.js
```diff
@@ -24,15 +53,38 @@ function sum() {}
...
+console.log(averageNumbers(numbersAvg));
+
 
 
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(arr) { 
+let sum = 0;
```
webLisbonTA, line 77: Same here, all your code inside the function should be written in relation with the function parameter, like this:
```
function averageWordLength(arr) { 
let sum = 0;
if (!arr.length) return null;

for (let i=0; i<arr.length; i++) {
  sum += arr[i].length;
 }
return sum / arr.length;
}
```

webLisbonTA (commented):
Hi Ana,

Well done! If you want to review function parameters and argument, please talk with one of the TAs

PR #2984, opened 2022-04-12
[FTLIS042022] Alexandre Alves

aletrad:


src/functions-and-arrays.js
```diff
@@ -52,14 +93,44 @@ const wordsUnique = [
...
+
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(findElem, words) {
+  let counter = 0;
```
webLisbonTA, line 117: Good logic! If you want to make it simpler, you could skip the counter and return true directly as soon as it find a elements that matches the work

webLisbonTA (commented):
Great work Alexandre!

A hint for iterations 3.1 and 4.1 - Bonus with the mixed array: -->  try to implement some conditions to check the type of the element on the array loop, you can do that either with na  if statement or with a switch -- and sum the value in a different way according with the type.  Example: 

- If type is string - sum the length of the string
- If type is number - sum the value
- If type is boolean - sum 1 if it is true and 0 if it is false

Also, remember that uou can later reuse the sum mixed array function on inside your avg function.

PR #2985, opened 2022-04-13
[wdft042022] Mariana Pontes

mariannasp8:


webLisbonTA (commented):
Great job on finishing all of the mandatory iterations and some bonus!

src/functions-and-arrays.js
```diff
@@ -1,41 +1,165 @@
...
+  let sum = 0;
+  for (let el of arr) {
+    if (typeof el === 'string') {
+      sum += el;
+    } else {
+      sum += el;
+    }
+  }
+  return sum;
+}
```
webLisbonTA, line 56: You are missing one test on this exercise. Try and research the "Throw Error" phrase for a better understanding of how it works

src/functions-and-arrays.js
```diff
@@ -1,41 +1,165 @@
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
+function averageNumbers(arr) {
+  if (arr.length === 0) {
+    return 0;
+  }
+  return sumNumbers(arr) / arr.length;
+}
```
webLisbonTA, line 130: Missing a condition that returns null if the parameter is an empty array

PR #2987, opened 2022-04-13
[WDLIS-FT-APR2022] João Antunes

joaocbantunes:
All the exercises are done, but getting two errors on the bonus and the bonus 8 i have no idea how to solve (found the solution on internet)


webLisbonTA (commented):
Great job on doing every mandatory iteration and also the bonus!

PR #2994, opened 2022-05-04
[REMOTO - PT - 052022] María Vence

marialalin:


src/functions-and-arrays.js
```diff
@@ -52,14 +121,46 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(c) {
```
EdgarHiSo, line 124: `function uniquifyArray(arr) {
  let newArr = []
  if (arr.length === 0) {
    return null
  } else {
      if (newArr.indexOf(arr[i] === -1)) {
        newArr.push(arr[i])
      }
    }
    return newArr
  }
}`

EdgarHiSo, line 124: `function uniquifyArray(arr) {
  let newArr = []
  if (arr.length === 0) {
    return null
  } else {
      if (newArr.indexOf(arr[i] === -1)) {
        newArr.push(arr[i])
      }
    }
    return newArr
  }
}`

PR #2995, opened 2022-05-04
[Remote-PT-APR2022] Pablo Alcon

PabloAlch:
Iterations completed without js test


EdgarHiSo (commented):
Muy buen lab Pablo!! A seguir así!!

PR #3005, opened 2022-05-07
chore: illustration

abernier:
adding an illustration:
image


ross-u (changes requested):
@abernier Nice illustration. After checking it seems that the image originally appears on the fireship.io.
We should go with another illustration or create a diagram of our own.

PR #3015, opened 2022-05-21
Solved lab

KiraTeresa:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,76 @@
...
-
-
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
+function averageNumbers(numbersArray) {
+  let sum = 0;
```
filipe-freire, line 45: Instead of re-implementing the sum logic, you could call the sum function you declared above and use it here! Isn't reusing functions wonderful? 😄

filipe-freire (commented):
Great job Kira! 🎉 Not much for me to add here, you did very well 😄 How about attempting some of the bonuses? 👀

KiraTeresa:
> Great job Kira! 🎉 Not much for me to add here, you did very well 😄 How about attempting some of the bonuses? 👀

Thank you for your input Filipe. The bonuses are also done now :)

filipe-freire:
Awesome work! 😁🏆

PR #3016, opened 2022-05-21
[WDPT-052022] Tiffany (Yun-Hsuan Chang)

tiffanychang-fr:


src/functions-and-arrays.js
```diff
@@ -78,9 +146,13 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
-
-
+function howManyTimes(wordsCount, search) {
+  let counter = 0;
+  for (let word of wordsCount) {
+    word === search && counter++;
```
filipe-freire, line 152: Nicely done! 🥇

src/functions-and-arrays.js
```diff
@@ -1,41 +1,90 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
+  return num1 > num2 ? num1 : num2;
```
filipe-freire, line 3: Great work using the ternary operator already! 👏🏻 👏🏻 

Another way to solve this one would be to use the `Math.max()` method. You can read more about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

filipe-freire (commented):
Amazing job Tiffany! Not much for me to say here, review wise! You did an excellent job. Congrats and keep it up! 🥇

PR #3018, opened 2022-05-23
Solved lab

daianeReina:
Done! All the iterations.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,41 +1,142 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(input) {
+  if (input.length === 0) {
+    return null;
+  }
+  let longestWord = '';
+  for (let i = 0; i < input.length; i++) {
+    if (longestWord.length < input[i].length) {
+      longestWord = input[i];
+    }
+  }
+  return longestWord;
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,142 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
```
filipe-freire, line 3: Good job here! Another way to solve this one would be to use the `Math.max()` method. You can read more about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

src/functions-and-arrays.js
```diff
@@ -1,41 +1,142 @@
...
+  }
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(input) {
+  if (input.length === 0) {
```
filipe-freire, line 14: Could also have written `if (!input.length) {....}`. So many ways to achieve the same, am I right? 😅

src/functions-and-arrays.js
```diff
@@ -1,41 +1,142 @@
...
+  if (element.length === 0) {
+    return null;
+  }
+  let sumElements = 0;
 
+  for (i = 0; i < element.length; i++) {
+    sumElements += element[i];
+  }
+
+  let averageElements = sumElements / element.length;
```
filipe-freire, line 81: You could have reused the `sum()` function written above (line 45) here! `averageElements = sum(element) / element.length` 

Also, since the value of `averageElements` won't be reassigned further in the code, best to declare it with a `const` instead of a `let`. 💯

filipe-freire (commented):
Very nicely done Daiane! 😄 Congratulations on another lab fiished! 🎉  Keep up the good work & let me know if you have any doubts you want to clarify. ✌🏻

PR #3034, opened 2022-05-27
[WDPT-Remote-05-2022] PELAYO PEREZ DE AYALA 

PelayoPAR:
Solved lab

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -178,7 +178,7 @@ function uniquifyArray(someArray) {
   }
   return uniqueArray;
 }
-
+ 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -14,7 +14,7 @@ function findLongestWord(arrWords) {
   if (arrWords.length === 0) {
     return null;
   } else if (arrWords.length !== 0) {
-    var longestWordYet = ' ';
+    let longestWordYet = ' ';
     for (i = 0; i < arrWords.length; i++) {
       if (arrWords[i].length > longestWordYet.length) {
         longestWordYet = arrWords[i];
@@ -95,8 +95,6 @@ function averageWordLength(words) {
 
 // Bonus - Iteration #4.1
 
-
-
 function avg(mixedArray) {
   if (mixedArray.length === 0) {
     return null;
@@ -114,7 +112,7 @@ function avg(mixedArray) {
     }
   });
   result /= mixedArray.length;
-  return Math.round(result*100)/100
+  return Math.round(result * 100) / 100;
 }
 
 // Iteration #5: Unique arrays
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,151 @@
...
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arrWords) {
+  if (arrWords.length === 0) {
+    return null;
+  } else if (arrWords.length !== 0) {
```
filipe-freire, line 15: This check here is redundant, since you are already checking if the length of the array is equal to 0 on line 13. This if block will be read in case it is not, automatically. 😄

src/functions-and-arrays.js
```diff
@@ -1,41 +1,151 @@
...
+  return sum;
+}
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
+function averageNumbers(nums) {
```
filipe-freire, line 98: You could have reused the sum function (which you built on line 50) here! Isn't reusing functions amazing? 😁

src/functions-and-arrays.js
```diff
@@ -78,9 +211,17 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
+function howManyTimes(wordsList, wordToCount) {
+  let count = 0;
 
+  let position = wordsList.indexOf(wordToCount);
 
+  while (position !== -1) {
```
filipe-freire, line 219: uhh using a while loop! Daring! 😄 Well done!

src/functions-and-arrays.js
```diff
@@ -52,16 +162,39 @@ const wordsUnique = [
...
+  let uniqueArray = [];
+
+  if (someArray.length === 0) {
+    return null;
+  }
+  
+  for (let i = 0; i < someArray.length; i++) {
+    if (uniqueArray.includes(someArray[i]) == false) {
+      uniqueArray.push(someArray[i]);
+    } else if (uniqueArray.includes(someArray[i]) == true) {
```
filipe-freire, line 175: This check on line 175 is redundant, since it is the exact opposite of line 173. In these cases, best to use a simple `if else` instead! ✌🏻

filipe-freire (commented):
Nice work finishing the lab Pelayo! 🥇 One of many to come! If you have time, give it a go with just array methods now. 😄 Congrats!

PelayoPAR:
Olá Filipe!

Duly noted and changed the redundancies. I suspected they were so but, I
guess I was afraid of getting more "TEST FAILED" notifications and left
them as they were heheh!
Indeed, I could have reused the sum function, however, at that point I was
not in my brightest state of mind 😅
About the while loop, I wanted to add some spice to the code 😋, to be
honest, I was kinda surprised it worked, and I actually took inspiration
from one of the thousands of website tabs I had opened at that time 😂.

Muito obrigado! Your feedback is much appreciated!
Cumprimentos,

Pelayo Pérez
Ironhacker 👾💪

PelayoPAR:
One last thing, should I commit the version with your corrections or leave
it as it is?

filipe-freire:
Sounds good Pelayo! You can commit & push the corrections if you wish, no problem there 😁

PR #3035, opened 2022-05-27
[REM-PT-052022] Nuno Morais

NunoD87:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,89 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(n1, n2) {
+  if (n1 > n2) return n1;
```
filipe-freire, line 3: Nice one Nuno! 🏆

Another way to solve this one would be to use the Math.max() method. You can read more about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

Or maybe even attempt solving it with the all mighty [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)! 😏

src/functions-and-arrays.js
```diff
@@ -1,41 +1,89 @@
...
+}
 
+// Iteration #3.1 Bonus:
+function sum(arr) {
+  let sum = 0;
+
+  if (!arr.length) return 0;
+
+  for (let i = 0; i < arr.length; i++) {
+    switch (typeof arr[i]) {
```
filipe-freire, line 47: Love the use case of a switch statement here! So clean 👌🏻

src/functions-and-arrays.js
```diff
@@ -1,41 +1,89 @@
...
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
+function averageNumbers(arr) {
+  if (!arr.length) return null;
+  return sumNumbers(arr) / arr.length;
```
filipe-freire, line 71: Reusing functions sure is the way to go! Lovely work 🥇

PR #3039, opened 2022-05-31
brain go brrr

ivvva:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,113 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
```
mloelkes, line 5: You could replace "else if"- and "else"-statements with "return num2"

src/functions-and-arrays.js
```diff
@@ -1,41 +1,113 @@
...
+}
 
 // Iteration #3.1 Bonus:
-function sum() {}
-
-
+const mixedArr = [6, 12, 'miami', 1, true, 'barca', '200', 'lisboa', 8, 10];
+function sum(mixedArr) {
+  let sum = 0;
+  mixedArr.forEach((number) => {
```
mloelkes, line 46: Not every element in the array is a number, so there might be a better name

src/functions-and-arrays.js
```diff
@@ -1,41 +1,113 @@
...
+    if (typeof number != 'string' && typeof number != 'boolean' && typeof number != 'number'){
+      throw new Error("Unsupported data type sir or ma'am")
+    } else if ((mixedArr.length === 0)) {
+      return 0;
+    } else if (typeof number == 'string') {
```
mloelkes, line 51: After all the null- and error-checks I would start with a new if-statement instead of else-if, that makes the code more readable

src/functions-and-arrays.js
```diff
@@ -1,41 +1,113 @@
...
+      number = number.length;
+    } else if (number == 'true') {
```
mloelkes, line 53: You could remove the "true"- and "false"-conditions, because "sum += number" is fine for numbers and booleans

src/functions-and-arrays.js
```diff
@@ -124,7 +217,7 @@ if (typeof module !== 'undefined') {
     avg,
     uniquifyArray,
     doesWordExist,
-    howManyTimes,
-    greatestProduct
+    howManyTimes
+    // greatestProduct
   };
 }
```
mloelkes, line 223: Good work! 👍

PR #3040, opened 2022-05-31
MAY-BERLIN-JOAO-LAB-FUNCTIONS-31.5.22

FullerJoao:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,45 @@
...
+  if (num1 > num2) return num1;
+  else return num2;
+}
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(str) {
```
mloelkes, line 11: This is an array, so a better name would be something like "strings"

src/functions-and-arrays.js
```diff
@@ -52,14 +92,36 @@ const wordsUnique = [
...
+  let uniqueArr = [];
+     for (let i = 0; i < words.length; i++) {
+      let word = words[i];
+
+       if (uniqueArr.indexOf(word) < 0) {
+       uniqueArr.push(word);
+     }
+   }
+
+  for (let word of words) {
```
mloelkes, line 106: The second for-loop is not necessary here

src/functions-and-arrays.js
```diff
@@ -78,7 +140,13 @@ const wordsCount = [
...
 ];
 
-function howManyTimes() {}
+ function howManyTimes(arr, word) {
+  let count = 0;
+  for (let el of arr) {
+    if (el === word) count++;
+  }
+  return count;
+ }
```
mloelkes, line 152: Good work! 👍

PR #3041, opened 2022-05-31
WD-Ber-JS/functsandArrayLab-AlexSmith

alexs-2020:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,120 @@
...
-
+console.log(findLongestWord(words))
+function findLongestWord(arr) {
+  let length = 0;
+  let lWord = ''
+  if( arr.length == 0){
+    return null;
+  }
+  for(let i = 0; i < arr.length; i++){
+    if(length < arr[i].length){
```
mloelkes, line 22: You could also use lWord.length here instead of length

src/functions-and-arrays.js
```diff
@@ -52,14 +131,41 @@ const wordsUnique = [
...
+  let uniqueArr = [];
+  if (arr.length == 0) {
+    return null
+  }
+  for (let i = 0; i < arr.length; i++) {
+    if(uniqueArr.indexOf(arr[i]) == -1){
+       uniqueArr.push(arr[i])
+    }
+    else {
+    continue
```
mloelkes, line 144: The "else"-condition is not necessary here, because you anyway continue with the next value

src/functions-and-arrays.js
```diff
@@ -78,7 +184,30 @@ const wordsCount = [
...
+  for (let i = 0; i < arr.length; i++) {
+   
+    if(uArr.indexOf(arr[i]) == -1){
+      console.log(uArr.indexOf(arr[i]))
+       uArr.push(arr[i])
+       if (arr[i] == search) {
+        counter = counter + 1     
+ }
+    }
+    else if (arr[i] == search) {
```
mloelkes, line 202: You only need your "else if"-logic here, the "if" and "else" is not necessary

PR #3042, opened 2022-05-31
AlexanderFarmakis-Berlin

alex-fmks:


src/functions-and-arrays.js
```diff
@@ -26,7 +48,13 @@ function sum() {}
...
+function averageNumbers(numbers) {
+  if(!numbers.length) {
+    return 0;
+  }
+  let sum = sumNumbers(numbers)
+  return sum / numbers.length
+}
 
 
 // Level 2: Array of strings
```
mloelkes, line 60: Good work, maybe you want to add the missing methods! 👍

PR #3045, opened 2022-05-31
60% Solved lab Wen-Berlin

WHL99:


src/functions-and-arrays.js
```diff
@@ -52,14 +124,44 @@ const wordsUnique = [
...
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(wordsUnique) {
+  let newArr = []
+  if (wordsUnique.length === 0) {
+    return null
+  }
+  for (let i = 0; i < wordsUnique.length; i++) {
+    if (wordsUnique.indexOf(wordsUnique[i], i + 1) == -1) {
```
mloelkes, line 133: You may want to check this logic again. You have to check the indexOf from the newArr

src/functions-and-arrays.js
```diff
@@ -52,14 +124,44 @@ const wordsUnique = [
...
+  }return newArr
+}
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(theWordNeededSearch) {
```
mloelkes, line 144: This function should get two arguments, the array and the word to search for

PR #3046, opened 2022-05-31
AntonioSaleme-Berlin

thusspokedata:


src/functions-and-arrays.js
```diff
@@ -1,42 +1,147 @@
...
-
-
+// should return the sum with one number array
+// should return zero if receives an empty array when called
+
+function sumNumbers(arr) {
+  let total = 0;
+  if (arr.length == 0) {
+    return 0;
+  } else if (arr.length == 1) {
```
mloelkes, line 44: I would start with a new if-condition to separate the different logical steps

PR #3050, opened 2022-06-01
Berlin Azadeh Galidari

devgirlpro:


src/functions-and-arrays.js
```diff
@@ -52,7 +148,36 @@ const wordsUnique = [
...
+
+  if(mixedWordsArrLength === 0) {
+    return null;
+  }
+
+
+
+  mixedWordsArr.forEach( (singleWord)=> {
+    console.log("singleWord ==>", singleWord)
+    for(let i = 0; i < mixedWordsArrLength; i++){
```
mloelkes, line 164: You don't need two loops here, you can just use this for-loop

src/functions-and-arrays.js
```diff
@@ -52,7 +148,36 @@ const wordsUnique = [
...
+      if (singleWord !== mixedWordsArr[i]){
```
mloelkes, line 165: Please check this logic again, it should just check if the element is already in the new array

src/functions-and-arrays.js
```diff
@@ -111,6 +250,12 @@ function greatestProduct() {}
...
 
+
+
+
+
+
+
 // The following is required to make unit tests work.
 /* Environment setup. Do not modify the below code. */
 if (typeof module !== 'undefined') {
```
mloelkes, line 261: Good work! 👍

PR #3061, opened 2022-06-02
Jule Schlag final version

zikaden:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(anothermixArr) {
+  if (anothermixArr.length === 0) {
+    return null;
+  }
+  let sum = 0;
+  for (let i = 0; i < anothermixArr.length; i++)
```
mloelkes, line 105: You could also use your sum-function here instead of the loop

src/functions-and-arrays.js
```diff
@@ -52,14 +129,36 @@ const wordsUnique = [
...
 // Iteration #6: Find elements
-const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
-
-function doesWordExist() {}
+const wordsFind = ['machine', 'subset', 'trouble', 'starting', , 'eating', 'truth', 'disobedience'];
+
+function doesWordExist(compareArr, lookUpWord) {
+  let match = [];
+  let i = 0;
+  while (i < compareArr.length) {
```
mloelkes, line 150: Cool solution! 😎

src/functions-and-arrays.js
```diff
@@ -106,7 +214,7 @@ const matrix = [
   [1, 70, 54, 71, 83, 51, 54, 69, 16, 92, 33, 48, 61, 43, 52, 1, 89, 19, 67, 48]
 ];
 
-function greatestProduct() {}
+function greatestProduct() { }
```
mloelkes, line 220: Good work! 👍

PR #3072, opened 2022-06-03
RUIZ, FRAN.Berlin

FranEjarque:


src/functions-and-arrays.js
```diff
@@ -1,24 +1,70 @@
...
+  }
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(Array) {
```
mloelkes, line 17: The argument "array" should be lowerCase

src/functions-and-arrays.js
```diff
@@ -1,24 +1,70 @@
...
+  let longestWord = Array[0];
+
+  if (Array === undefined || Array.length == 0 ) {
+    return null
+  } if (Array.length === 1) {
+    return Array[0]
+  } for ( let i = 0; i < Array.length; i++) {
+    if ( Array[i].length > longestWord.length) {
+      longestWord = Array[i]
+    } if ( Array[i].length === longestWord.length) {
```
mloelkes, line 27: You don't need the second if-statement in the for-loop

src/functions-and-arrays.js
```diff
@@ -1,24 +1,70 @@
...
+  let total = 0;
+
+  if (arr.length == 0) {
+    return 0
+  } else if (arr.length == 1) {
+    return arr[0]
+  } 
+  return total 
+};
+
```
mloelkes, line 70: Well done! 👍 If you find the time, try to add the missing functions as well.

PR #3086, opened 2022-06-16
lab done with bonuses-Palash 

palashchadha:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,90 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(a, b) {
+  if (a >= b) {
+    return a;
+  }
+  return b;
```
TA-Remote, line 6: Little challenge for you here :)

```
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,90 @@
...
 
-function averageNumbers() {}
-
+function averageNumbers(arr) {
+  if (arr.length === 0) {
+    return null;
+  } else {
+    return sumNumbers(arr) / arr.length;
+  }
+}
```
TA-Remote, line 67: Great you are reusing the function for sum!

src/functions-and-arrays.js
```diff
@@ -52,16 +101,34 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
-
-
+function uniquifyArray(arr) {
+  if (arr.length === 0) {
+    return null;
+  }
```
TA-Remote, line 107: For shorter syntax, if you return only one line of code, you can write it like this

```
```

TA-Remote (commented):
Awesome job Palash!

PR #3089, opened 2022-06-16
[RMFT June '22] Lisa Hilterhaus: solved lab

cortinaHs:


TA-Remote:
reusing function and a DRY code! really Great work!

PR #3090, opened 2022-06-16
Alex's pull request for the lab-javascript-functions-and-arrays

Virgule09:
I almost completed all the steps, but had no time to dig deeper and understand why it doesn't work. I feel that with more time I could find what doesn't work.
With the solution I will practice more during the week ends on the optional lab.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -172,8 +172,8 @@ function howManyTimes(arrayArgument, wordArgument) {
       return 0;
     };
   }
-  
 }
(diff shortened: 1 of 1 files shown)


TA-Remote (commented):
- Great job overall! 
- Pay attention where you're declaring your variables, mind the scope (this can get tricky and hard!)
- Pay attention to the indentation of your code - It'll be easier to read, understand your code and indentify scopes :)

- Keep practicing and this points will get easier with time.
- Your code is easy to understand which is great!!

PR #3092, opened 2022-06-16
Paul Tiry [WDFT JUN2022]

ptiry02:
I was able to do the bonus iteration 8 💪  but my brain exploded while I was trying the 8.1 🤯.
🤣🤣

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,41 +1,95 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,95 @@
...
 
-function averageNumbers() {}
-
+function averageNumbers(numbersList) {
+  let sum = 0;
+  if (!numbersList[0]) return null;
+  numbersList.forEach(function (n) {
+    sum += n;
+  });
+  return sum / numbersList.length;
```
TA-Remote, line 65: Another approach would be to call `sum() `function that you already defined ;)

TA-Remote (commented):
Great work!

PR #3093, opened 2022-06-16
WD RM June 2022 Matthieu DUPERREY

MatthieuDUPERREY:


src/functions-and-arrays.js
```diff
@@ -1,32 +1,61 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(numbersAvg) {
+  if (numbersAvg === 0) {
+    return null
+  } else {
+    for (let sum = 0, i = 0; i < numbersAvg.length; i++) {
```
TA-Remote, line 52: and here , you need to declare sum outside of for loop and also if..else statement,if you want to return it directly in the function, because of scope ;)
  let sum = 0
  if (numbersAvg.length === 0) {
    return null
  } else {
    
    for ( let i = 0; i < numbersAvg.length; i++) {
      sum += numbersAvg[i];
    }
    
  }

  return sum / numbersAvg.length;
}

PR #3094, opened 2022-06-16
Stephanie Tovar - Solved lab

Stephanie-Tovar:
I did not have the the to do the bonuses :(

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,20 +1,40 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  return Math.max(num1, num2);
+}
+maxOfTwoNumbers(43, 24)
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(words) {
+  if (words.length === 0){
+    return null;
+  }
+    let longestWord = "";
+    for (let i = 0; i < words.length; i++){
+      if (longestWord.length < words[i].length){
+        longestWord = words[i];
+      }
+    }
+  return longestWord;
+  }
+  
+findLongestWord(words)
 
 
 // Iteration #3: Calculate the sum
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,20 +1,40 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  return Math.max(num1, num2);
+}
+maxOfTwoNumbers(43, 24)
```
TA-Remote, line 5: After you finish the exercise, project etc - try to delete any unused lines of code. We only had to define the functions, no need to call them ;)

src/functions-and-arrays.js
```diff
@@ -1,20 +1,40 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  return Math.max(num1, num2);
```
TA-Remote, line 3: Cool you're finding new methods to use!

src/functions-and-arrays.js
```diff
@@ -1,20 +1,40 @@
...
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(words) {
+  if (words.length === 0){
+    return null;
+  }
```
TA-Remote, line 14: For shorter syntax, if you return only on line after if statement, you could write it like this:
```
```

src/functions-and-arrays.js
```diff
@@ -26,13 +46,27 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
+function averageNumbers(numbersAvg) {
+  if (numbersAvg.length === 0){
+    return null;
+  }
+let average = sumNumbers(numbersAvg) / numbersAvg.length;
```
TA-Remote, line 53: You reused the previous function! :ok_hand:

TA-Remote (commented):
Nice job Stephanie, keep up great work!

Awesome that you're using array methods already, like filter() method :muscle:

PR #3095, opened 2022-06-16
Lara Rockenstein WD FT June 22

LMRBLN:


TA-Remote:
Great Job!

src/functions-and-arrays.js
```diff
@@ -78,7 +228,21 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
+function howManyTimes(testWord, arr) {
```
TA-Remote, line 231: you need to pass array as first argument and test word as second argument.
 function howManyTimes( arr, testWord)

PR #3098, opened 2022-06-16
Tobias Conrad - WDFTJUN22

maktooob:
couldn´t wrap my head around the last one sadly

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,60 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1
+  }
+  else return num2
+}
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(words) {
+  let longestWord = words[0];
+  let result;
+  if (words.length === 0) {
+    return null
+  }
+  for (let i = 0; i< words.length; i++) {
+   if (words[i].length > longestWord.length) {
+          longestWord = words[i];     
+   }
+  }
+  return longestWord;
+}
 
 
 // Iteration #3: Calculate the sum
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1
+  }
+  else return num2
```
TA-Remote, line 6: Little challenge for you here :)

```
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(words) {
+  let longestWord = words[0];
+  let result;
+  if (words.length === 0) {
+    return null
+  }
```
TA-Remote, line 18: For shorter syntax, if you return only on line after if statement, you could write it like this:
```
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
+     averageArr += 1
+   }
+   else if  (typeof wordsArr[i] === "string") {
+     averageArr += wordsArr[i].length 
+   }
+   else if (typeof wordsArr[i] === "number") {
+     averageArr += wordsArr[i]
+   }
+ 
+}
```
TA-Remote, line 55: Here is the syntax to throw an error. However you might also want to check for the case if array element is boolean :)

```
else {
    }
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
-function averageNumbers() {}
+function averageNumbers(numbersAvg) {
+  let average = 0;
+  if (numbersAvg.length === 0){ 
+    return null
+  }
+  for (let i = 0; i < numbersAvg.length; i++){
+    average += numbersAvg[i] 
+  }
+return average / numbersAvg.length
```
TA-Remote, line 73: Another approach would be simply to call `sum() `function you described above

TA-Remote (commented):
Well done!

PR #3099, opened 2022-06-16
solves problems

zubairmalek:
solved problems!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,49 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a >= b) {
+    return a 
+  } else {
+    return b
+  }
+}
 
 
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,49 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a >= b) {
+    return a 
+  } else {
+    return b
+  }
```
TA-Remote, line 7: Little challenge for you here :) 

```
```

src/functions-and-arrays.js
```diff
@@ -1,19 +1,49 @@
...
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+const words = [];
+
+function findLongestWord(words) {
+  // console.log(words)
+  let longestWord=''
+  if (words.length === 0) {
+   
+    return null;
+  }
```
TA-Remote, line 21: For shorter syntax,  if you return only on line after if statement, you could write it like this:

```
```

PR #3103, opened 2022-06-17
Nic Solitom - Solved Lab (will do Bonus tasks during weekend)

nicsolitom:


src/functions-and-arrays.js
```diff
@@ -52,16 +114,36 @@ const wordsUnique = [
...
+  } else {
+  let checkedWords = [];
+  wordsArr.forEach(function (word) {
+    if (!checkedWords.includes(word)) {
+      checkedWords.push(word)
+    }
+  })
+  return checkedWords;
+  }
+}
```
TA-Remote, line 129: As alternative, try to solve this one with `indexOf() `:)

src/functions-and-arrays.js
```diff
@@ -78,8 +160,19 @@ const wordsCount = [
...
+  } else {
+    let count = 0;
+    wordsArr.forEach(function (word) {
+      if (word === wordToFind) {
+        count++;
+      }
+    })
+    return count;
+  }
+}
```
TA-Remote, line 175: Only a suggestion for alternative layout of the code -> to make it easier to read.

- First thing, you can rule out all the edge case.
Plus, if you are returning only one line -> you can use below syntax

```
if (wordsArr.length === 0) return 0;
```

- And after do remaining logic 
```
 let count = 0;
    wordsArr.forEach(function (word) {
      if (word === wordToFind) {
        count++;
      }
    })
    return count;
    
```

Again, just a suggestion, but you write as you prefer ;)

TA-Remote (commented):
Great !

PR #3104, opened 2022-06-17
Jack Le

jackletran:
Some functions work but don't pass the tests, will try to fix today or on the weekend

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -40,7 +40,7 @@ function sumNumbers(numbers) {
 
 // Iteration #3.1 Bonus:
 
-const mixedArr = [6, 12, 'miami', 1, true, 'barca', '200', 'lisboa', 8, 10];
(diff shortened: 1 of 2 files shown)


TA-Remote:
Very good. 
some of your test not passed, because you declared variables outside function, also you called function that brokes the automated test. ;)

src/functions-and-arrays.js
```diff
@@ -52,16 +148,71 @@ const wordsUnique = [
...
 
-function doesWordExist() {}
+let keyword = 'truth';
+let t = 0;
+let itsTrue = 0;
+let itsFalse = 0;
+
+function doesWordExist(wordsFind, keyword) {
+  while (itsTrue === 0) {
+    if (wordsFind[t].includes(keyword)) {
```
TA-Remote, line 175: Here, you need to check if keyword is inside wordsFind array, so you can write :
if(wordsFind[t] == keyword){
 return true;
  }
}

jackletran: Hey, thanks for the feedback, I fixed all the errors for the mandatory iterations now!

PR #3105, opened 2022-06-17
Khalid 

Khalid9180:
worked Zubair and Oscar on iteration 2,3 and 4
I had issue with iteration elements

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,44 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a,b) {
+ if(a>=b) {
+  return a
+ } else if (b>a) {
+  return b
+ }
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+function findLongestWord(words) {
+  let longestWord = '';
+  if (words.length === 0 ) {
+    return null }
+  for (i=0; i<words.length; i++) {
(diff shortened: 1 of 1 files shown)


TA-Remote:
Good Job Khalid!

src/functions-and-arrays.js
```diff
@@ -51,15 +90,28 @@ const wordsUnique = [
...
+    return wordsUnique.filter((item, index) => wordsUnique.indexOf(item) === index)
+  }
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(wordsFind, word5) {
```
TA-Remote, line 104: here you need to write for loop,inside if..else statement  :
 function doesWordExist(wordsFind, word5) {
  if (wordsFind.length === 0) {
    return null
  }
  else {
    for (i = 0; i < wordsFind.length; i++) {
      if (wordsFind[i] == word5) {
        return true
      } 
    }
    return false
  }
}

src/functions-and-arrays.js
```diff
@@ -78,7 +130,16 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
+function howManyTimes(wordsCount, word9 ) {
+  if (wordsCount.length === 0 ) {
+    return 0 }
+     x =0
```
TA-Remote, line 136: when you want to declare a varible , need to use let/const word 
also it is better you choose a meaningful name for your variable ;) in this case, you can write counter instead of x 
let counter = 0

PR #3134, opened 2022-07-08
[FTLISBON072022] Andrea Arestigue

aarestigue:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,85 @@
...
+  
+
 
 
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers() {
```
webLisbonTA, line 57: On this function you should add a parameter, meaning that it should look like this:
```
function sumNumbers(arr) {
  let sum = 0;
  for (i=0; i < arr.length; i++ ) {
      sum += arr[i]
  }
  console.log (sum);
  return sum;
}
```
Besides that, the whole logic was perfect

src/functions-and-arrays.js
```diff
@@ -52,7 +118,24 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
+
+function uniquifyArray() {
```
webLisbonTA, line 122: Here you had the same issue as before, you were not adding a parameter to your function

webLisbonTA (commented):
Good job on the lab!

PR #3172, opened 2022-07-23
BER PT Patrick

Patrick-Groot:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,98 @@
...
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  let longestWord = []
+  if (arr.length === 0) {
+    return null}
+  else {
```
trebeljahr, line 15: you can drop the else and one level of indentation

trebeljahr: a better version of findLongestWord would look something like this: 
```javascript
if (!arr.length) return null;
let longestWordSoFar = arr[0];
for (let word of arr) {
  if (word.length > longestWordSoFar.length) {
    longestWordSoFar = word;
  }
}
return longestWordSoFar;
 ```

File: src/functions-and-arrays.js [hunk 2]
```diff
@@ -1,41 +1,98 @@
...
 // Iteration #3.1 Bonus:
-function sum() {}
-
+function sum(arr) {
+  let result = 0;
+  for (let i = 0; i < arr.length; i++){
+    if (typeof arr[i] === 'string') {
+      result += arr[i].length;
+    } 
+    else if (typeof arr[i] === 'object' || typeof arr[i] === 'array'){
```
trebeljahr, line 59: change the order around - the last statement in the if/else chain should check errors, this way if you forgot to handle a datatype it will throw an error.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,98 @@
...
 
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  if (arr.length === 0) {return null}
+  else if (arr.length > 0) {return sumNumbers(arr) / arr.length;
```
trebeljahr, line 76: ternary would make this code even better.
`return !arr.length ? null : sumNumbers(arr) / arr.length`

src/functions-and-arrays.js
```diff
@@ -52,15 +109,32 @@ const wordsUnique = [
...
+function uniquifyArray(arr) {
+  let arr2 = [];
+  if (arr.length === 0) {return null}
+  else {
+    for (i = 0; i < arr.length; i++) {
+    if (arr2.length === 0) {
+      arr2.push(arr[i])
+      }
+    else if (arr2.includes(arr[i]) === false) {arr2.push(arr[i])}
+    else continue
```
trebeljahr, line 121: the loop will continue on its own. It's not necessary to add another else block that says continue ^^

src/functions-and-arrays.js
```diff
@@ -106,9 +187,42 @@ const matrix = [
...
 ];
 
-function greatestProduct() {}
-
-
+function greatestProduct(matrix) {
+  let resultHorizontal = 0;
+  let resultVertical = 0;
+  let highestResult = 0;
+  for (i = 0; i < matrix.length; i++){
```
trebeljahr, line 194: very clean good job.

src/functions-and-arrays.js
```diff
@@ -106,9 +187,42 @@ const matrix = [
...
+};
+
+
+function greatestProductOfDiagonals(matrix) {
+  let resultTowardsBottomRight = 0;
+  let resultTowardsBottomLeft = 0;
+  let highestResult = 0;
+  for (i = 0; i < matrix.length; i++){
+    for (j = 0; j < matrix.length - 3; j++){
+      resultTowardsBottomRight = matrix[j][i] * matrix[j+1][i+1] * matrix[j+2][i+2] * matrix[j+3][i+3];
```
trebeljahr, line 215: this should break when you reach i at the end of matrix.length.

trebeljahr: because i + 3 when i === matrix.length - 1 shouldn't exist in the array. so will be undefined, so would give you NaN.

trebeljahr (commented):
really good work. Impressive thinking on the bonus iterations especially. :+1:

PR #3179, opened 2022-07-25
JS lab

avarap:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,162 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers( a, b) {
+  if (a>b){
```
trebeljahr, line 3: Math.max would be your friend here :)

src/functions-and-arrays.js
```diff
@@ -1,41 +1,162 @@
...
+  }
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(wordArray) {
```
trebeljahr, line 19: this can be simplified. 
```javascript
if (!arr.length) return null;
let longestWordSoFar = arr[0];
for (let word of arr) {
  if (word.length > longestWordSoFar.length) {
    longestWordSoFar = word;
  }
}
return longestWordSoFar;
  ```

File: src/functions-and-arrays.js [hunk 3]
```diff
@@ -1,41 +1,162 @@
...
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers(arr) {
+  if (arr.length === 0){
+    return 0;
+  }
+  if (arr.length === 1){
```
trebeljahr, line 47: same idea here, no need to check the arr.length === 1 case separately, because you would automatically add that to your sum. 0 + n = n; so your loop would only do one iteration with only one element and the "sum" of an array of 1 number only is well... that number.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,162 @@
...
+  let sum=0;
+  for(var i=0; i<= arr.length-1; i++){
+    let arrObjType=typeof(arr[i]);
+    if (arrObjType==="number"){
+      sum+=arr[i];
+    }
+    else if (arrObjType==="string"){
+      sum+=arr[i].length;
+    }
+    else if(arrObjType==="boolean"){
```
trebeljahr, line 76: funnily enough + on a boolean would convert the boolean into a number per default, so you could drop those extra cases entirely. A valid solution would be just handling the string case different and the other two can use the regular + operator and utilize the fact that JS does type coercions on booleans :)

src/functions-and-arrays.js
```diff
@@ -1,41 +1,162 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  if (arr.length === 0){
+    return null;
+  }
+  if (arr.length === 1){
```
trebeljahr, line 101: you have a sum function above. use it ;)

src/functions-and-arrays.js
```diff
@@ -1,41 +1,162 @@
...
 
-function averageWordLength() { }
+function averageWordLength(arr) {
+  if (arr.length === 0){
+    return null;
+  }
+  if (arr.length === 1){
+    return arr[0].length;
+  }
+  let sum=0;
```
trebeljahr, line 122: same idea here, if you have a function for something, use it.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,162 @@
...
+  for(var i=0; i<= arr.length-1; i++){
+    sum+=arr[i].length;
+  }
+  return sum/arr.length;
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(arr) {
```
trebeljahr, line 130: and again ^^ you could re-use the summation logic you already wrote by calling that function.

src/functions-and-arrays.js
```diff
@@ -52,14 +173,34 @@ const wordsUnique = [
...
+function uniquifyArray(arr) {
+  if (arr.length === 0){
+    return null;
+  }
+  if (arr.length === 1){
+    return arr[0];
+  }
+  let newArr=[];
+  for(let i=0; i<=arr.length-1;i++){
+    if(newArr.indexOf(arr[i]) === -1){
```
trebeljahr, line 185: that's a clever way to do it. there is .includes though ;)

src/functions-and-arrays.js
```diff
@@ -52,14 +173,34 @@ const wordsUnique = [
...
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(arr, word) {
+  if (arr.length === 0){
+    return null;
+  }
+
+  return (arr.indexOf(word) > -1);
```
trebeljahr, line 202: same here. .includes is a thing

src/functions-and-arrays.js
```diff
@@ -106,9 +256,25 @@ const matrix = [
...
+function greatestProduct(arr) {
+  if (arr.length === 0){
+    return null;
+  }
+  if (arr.length === 1){
+    return arr[0];
+  }
+  let sum=0,x=arr.length,y=0;
+  for(let i=0;i<=arr.length-1;i++){
+    let arr2=arr[i];
```
trebeljahr, line 268: I would name this row.

src/functions-and-arrays.js
```diff
@@ -106,9 +256,25 @@ const matrix = [
...
+    y+=arr2.length;
+    for(let j=0;j<=arr2.length-1;j++){
+      sum+=arr2[j];
+    }
+  }
+  if (sum===y)return 1;
```
trebeljahr, line 274: this is passing the test cases but not really what the exercise asks for haha 

when you loop over the row, one thing you want to do is to loop only until row.length - 3; this way you can use the index j, j+1, j+2 and j+3 to get the 4 elements in the row without running over at the end with the j+3. then you multiply these values together to get one of the possible products for this row, push this product into a list of products and then find the maximum value of that list in the end... 

```javascript 
const allProducts = []
for (let row of arr) {
  for (let j = 0; j < row.length - 3; j++){
    allProducts.push(row[j] * row[j+1] * row[j+2] * row[j+3])
  }
}
return Math.max(...allProducts)
```

trebeljahr: the code I wrote there only covers the row case, but I leave it as an exercise to you to translate it to cover the column case as well :)

trebeljahr (commented):
really good work. some things can be refactored and I left a comment trying to explain how you could tackle the bonus :)

PR #3195, opened 2022-07-30
Kinda done :-D

herrFernandez:


src/functions-and-arrays.js
```diff
@@ -1,21 +1,43 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
 
+function maxOfTwoNumbers(firstNumber,secondNumber) {
+
+  if (firstNumber > secondNumber) {
```
trebeljahr, line 5: Math.max is your friend.

src/functions-and-arrays.js
```diff
@@ -1,21 +1,43 @@
...
+    return secondNumber;
+  }
+}
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord() {
```
trebeljahr, line 16: in pseudocode a solution would look like this. 
init the longestWord to ""
loop over each word
if currentWord is bigger than longestWord 
update longestWord to currentWord
after loop return longestWord

src/functions-and-arrays.js
```diff
@@ -26,16 +48,69 @@ function sum() {}
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(array) {
+  
+  if (array.length === 0) {
+    return null;
+  } else {
+    const average = sumNumbers(array) / array.length;
+  return average;
```
trebeljahr, line 57: indentation is important - try to match indentation - or even better, set up prettier and let it do the heavy lifting for you.

src/functions-and-arrays.js
```diff
@@ -26,16 +48,69 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(array) {
+  
+  if (array.length === 0) {
+    return null;
+  } else {
+    const average = sumNumbers(array) / array.length;
```
trebeljahr, line 56: I like that you re-use the function you already wrote, so far nobody else had this as part of their solution and I marked it in every single feedback. Good job :)

src/functions-and-arrays.js
```diff
@@ -26,16 +48,69 @@ function sum() {}
...
+  if (array.length === 0) {
+    return null;
+  } else { 
+    let sum = 0;
+    let average = 0;
+    for(let i = 0; i < arr.length; i++) {
+    let number = arr[i].length;
+    sum += (number.length);
+    }
+    average = sum / arr.length;
```
trebeljahr, line 78: you can use your generic sum function here ^^

src/functions-and-arrays.js
```diff
@@ -26,16 +48,69 @@ function sum() {}
...
+  }
+ 
+}
+
+
+function avg(arrayOfWords) {
+  let sum = 0;
+
+  arrayOfWords.forEach(item => {
+    sum += item;
```
trebeljahr, line 104: this wouldn't work. it just adds the current word to sum. not the length. also you have this function avg. defined twice. maybe rename or delete/comment out one of them.

trebeljahr (commented):
overall good work, but like you said yourself you should go over it once more and try to finish more of it/fix some stuff when you find the time.

PR #3197, opened 2022-08-09
Ber-WDFT-Kenneth Elumba-first iteration

kenpun:


src/functions-and-arrays.js
```diff
@@ -1,20 +1,46 @@
...
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers() {
+  if (a > b) {
+    return a 
+   } else if (a < b) {
+    return b
+   } else if (a === b) {
+    return a || b
```
mloelkes, line 8: This will always return `a`, because `a` is always truthy. You wouldn't need the last else-if-statement, because if both numbers are equal, you just can return `b`.

src/functions-and-arrays.js
```diff
@@ -52,16 +99,46 @@ const wordsUnique = [
...
-
+function doesWordExist(arr, searchWord) {
+  if (arr.length === 0) {
+    return null;
+  } else {
+    let match = false;
+
+    for (let i = 0; (i < arr.length) & !match; i++) {
+      if (arr[i] === searchWord) {
+        match = true;
```
mloelkes, line 133: You could also move the return-statement into the loop (in line 133), then you don't need the `match`-variable and the `break`.

src/functions-and-arrays.js
```diff
@@ -78,8 +155,14 @@ const wordsCount = [
...
+  let wordCount = 0
+  for (const str of strArray){
+    if (str === strToCount)
+      wordCount++;
+  }
+  return (wordCount)
+}
 
 
 // Iteration #8: Bonus
```
mloelkes, line 168: Good solutions! 🏝

PR #3198, opened 2022-08-09
[BER-WBFT-AUG] Pavel Popov

veykos:
Solved the lab + bonus exercises.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,41 +1,110 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a,b) {
+  if (a > b) {
+    return a;
+  } else {
+    return b;
+  }
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -106,9 +209,65 @@ const matrix = [
...
+
 
 
+function greatestProductOfDiagonals(matrix) {
+  let biggest_product_left = 0;
+  let biggest_product_right = 0;
+
+  for (let i = 0; i <= matrix.length - 4; i++) {
+    for (let z = 0; z <= matrix.length -4; z++) {
+      let current_sum = matrix[i][z] * matrix[i+1][z+1] * matrix[i+2][z+2] * matrix[i+3][z+3];
```
mloelkes, line 249: If you increment both `i` and `z`, you will "move" diagonal through the array. You would need to only increment `i` or `z` to "move" horizontally or vertically through the array.

src/functions-and-arrays.js
```diff
@@ -106,9 +209,65 @@ const matrix = [
...
+
+  if (biggest_product_left > biggest_product_right) {
+    return biggest_product_left;
+  } else {
+    return biggest_product_right;
+  }
+}
 
 
 // The following is required to make unit tests work.
```
mloelkes, line 273: Very good solutions! 👍

PR #3199, opened 2022-08-09
BER FTWD AUG2022 Jean Lab function and arrays

jpta24:


src/functions-and-arrays.js
```diff
@@ -78,7 +217,19 @@ const wordsCount = [
...
 
-function howManyTimes() {}
+function howManyTimes(wordsCount,string) {
+  if(wordsCount.length === 0){
+    return 0
+  }
+  let numberTimes = 0
+  for(let word of wordsCount){
+    if(word === string){
+      numberTimes = numberTimes + 1
```
mloelkes, line 227: You could replace this with `numberTimes++`

src/functions-and-arrays.js
```diff
@@ -78,7 +217,19 @@ const wordsCount = [
...
+    }
+  }
+  return numberTimes
+  
+}
```
mloelkes, line 235: Very good work! 👍

PR #3200, opened 2022-08-09
BER_JohannaKuhlee_Solved lab

jk030:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,49 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a, b) {
+  if (a>b){
+    return a
+  }
+  else if (b>a) {
+    return b
+  }   
+  else if (a=b) {
```
mloelkes, line 9: Instead of your else if-statements, you could just write `return b`, because if a ist not the highest number, b is the highest number. If both are equal, it doesn't matter which you return.

src/functions-and-arrays.js
```diff
@@ -52,15 +110,40 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
+function uniquifyArray (arrayWithDublicatesEntries) {
+
+  if (arrayWithDublicatesEntries.length === 0) 
+    return null;
+
+    let arrayWithDublicates = [];
```
mloelkes, line 118: A better name would be `arrayWithoutDuplicates`

src/functions-and-arrays.js
```diff
@@ -52,15 +110,40 @@ const wordsUnique = [
...
+function uniquifyArray (arrayWithDublicatesEntries) {
+
+  if (arrayWithDublicatesEntries.length === 0) 
+    return null;
+
+    let arrayWithDublicates = []; 
+
+  for (let entry of arrayWithDublicatesEntries) { 
+  if (!arrayWithDublicatesEntries.includes(entry))
+     {arrayWithDublicatesEntries.push(entry)}
```
mloelkes, line 122: Instead of `arrayWithDublicatesEntries`, you should work here with the `arrayWithoutDuplicates`

src/functions-and-arrays.js
```diff
@@ -128,3 +210,5 @@ if (typeof module !== 'undefined') {
     greatestProduct
   };
 }
+
+
```
mloelkes, line 214: Good solutions! 🏝

PR #3201, opened 2022-08-09
Ber-FTWD-Joanna Zielinska

j-zielinska:


src/functions-and-arrays.js
```diff
@@ -106,7 +197,34 @@ const matrix = [
...
+            maxColums = sumC
+          }   
+     }
+    }
+    if  (maxRow > maxColums ) {
+      return maxRow
+
+    }
+    return maxColums
+}
```
mloelkes, line 227: Very good solutions! ⭐

PR #3204, opened 2022-08-10
BER WDFT - W1D2  Camila Yokoo - LAB Functions and Arrays

kyotosng:


src/functions-and-arrays.js
```diff
@@ -52,16 +110,32 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
-
-
+function uniquifyArray(wordsUnique) {
```
mloelkes, line 113: I wouldn't call this variable wordsUnique, because it is the array which contains also duplicate entries.
Better would be something like "wordsWithDuplicateEntries"

src/functions-and-arrays.js
```diff
@@ -78,7 +152,23 @@ const wordsCount = [
...
+function howManyTimes(wordsCount, searchedWord) {
+  let timesWordAppears = 0;
+
+  if (wordsCount.length === 0) {
+    return 0;
+  } 
+  
+  for (let word of wordsCount) // word would be the equivalent of wordsCount[i]
+    if (word === searchedWord) {
+      timesWordAppears += 1;
```
mloelkes, line 164: You could also use timesWordAppears++ here

src/functions-and-arrays.js
```diff
@@ -78,7 +152,23 @@ const wordsCount = [
...
+    }
+    return timesWordAppears;
+  
+
+    // for/in loop should be used for objects
+    // for/of loop should be used for arrays
+}
```
mloelkes, line 173: Good work, keep it up! 🏝

PR #3205, opened 2022-08-10
Berlin-WDFT-Erika

helloannaerika88:
Solved lab

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,70 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(a,b) {
+  if (a > b) return a;
+  return b
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -106,7 +208,35 @@ const matrix = [
   [1, 70, 54, 71, 83, 51, 54, 69, 16, 92, 33, 48, 61, 43, 52, 1, 89, 19, 67, 48]
 ];
 
-function greatestProduct() {}
+function greatestProduct(matrix) {
+  let rowNumber = 0;
+  let columnNumber = 0;
+  let greatestResult = 0;
+  
+  for (let i = matrix [0][0]; i < 17; i++) {
```
mloelkes, line 216: To iterate over all elements in the two-dimensional-array, you would need both loops nested inside of each other.

src/functions-and-arrays.js
```diff
@@ -106,7 +208,35 @@ const matrix = [
...
+  for (let maxColumnNumber of columnNumber) {
+    if (maxColumnNumber > greatestResult) {
+      greatestResult = maxColumnNumber;
+    }
+
+  
+  }
+
+  return greatestResult;
+}
```
mloelkes, line 241: Good work! 👍

PR #3226, opened 2022-08-12
BER-WDF-082022-Miriam

miriam-plt:


src/functions-and-arrays.js
```diff
@@ -1,43 +1,109 @@
-// Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+// Iteration #1: Find the maximum OK
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a
+  } else if (a < b) {
+    return b
+  } else if (a === b) {
```
mloelkes, line 7: You don't need the `else if` statement, because if `a` and `b` are the same, you can just return `b`.

src/functions-and-arrays.js
```diff
@@ -1,43 +1,109 @@
...
+// Level 1: Array of numbers OK
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  let sum = 0
+  if (arr[0] === undefined){
+    return null
+  }
+  for (let i = 0; i < arr.length; i++){ //(use the function) return sumNumbers(numbers) / numbers.length;
```
mloelkes, line 84: You could also use your own sum-function here

src/functions-and-arrays.js
```diff
@@ -52,15 +118,45 @@ const wordsUnique = [
...
 
-// Iteration #6: Find elements
+
+
+// Iteration #6: Find elements OK
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(words, wordToSearch) {
+  if (words[0] === undefined){
```
mloelkes, line 148: Good idea! You could also check for `words.length === 0` here

src/functions-and-arrays.js
```diff
@@ -78,7 +174,21 @@ const wordsCount = [
...
+    }
+  }
+  return sum
+}
+
+// function with two parameters
+// new variable = 0
+// for of loop
+// if 
+// count++
```
mloelkes, line 191: Good work! 🌞

PR #3227, opened 2022-08-12
Berlin Paul Rombach 

paulifauli:


src/functions-and-arrays.js
```diff
@@ -78,7 +141,21 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
+
+function howManyTimes (a,b){
```
mloelkes, line 145: `a` and `b` are not good names for the arguments, because they are too generic. Something like `words` and `wordToSearchFor` would be better

src/functions-and-arrays.js
```diff
@@ -78,7 +141,21 @@ const wordsCount = [
...
+let count = 0;
+for (i=0;i<a.length;i++) {
+if (a[i] !== b){
+continue} else {
+count += 1
+}
+}
+return count
+}
+
```
mloelkes, line 160: Good solutions! 👍

PR #3231, opened 2022-08-14
almost solved lab

maurotbarone:


src/functions-and-arrays.js
```diff
@@ -78,8 +173,13 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
-
+function howManyTimes(words,word) {
+  if (words.length===0) {return 0}
+  filterdArray= words.filter(function(words){
```
mloelkes, line 178: The argument here should be `word`

src/note.js
```diff
@@ -0,0 +1,20 @@
+const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
+
+function sum(nbrs) {
+  let total=0;
+  if ((nbrs!== 'number')||(nbrs!== 'string')||(nbrs!== 'boolean')){return 'error'}
```
mloelkes, line 5: If the data type is not correct, you should throw an exception.
You can do that by using `throw`:
`throw "Unsupported data type sir or ma'am";`

src/note2.js
```diff
@@ -0,0 +1,30 @@
...
+    'sharp',
+    'playground',
+    'poison',
+    'communion',
+    'simple',
+    'bring'
+  ];
+  
+function uniquifyArray(vocabulary) {
+    newArray=vocabulary;
```
mloelkes, line 16: To create a new array, you need to use
`const newArray = []` or
`const newArray = new Array(<amount of elements>)`

src/tempCodeRunnerFile.js
```diff
@@ -0,0 +1 @@
+wordsArr
```
mloelkes, line 1: Gooood work! ✨✨

PR #3454, opened 2022-11-24
[FTRMT112022] Jonnathan Pineda

Jogopin:
almost everything done


Jogopin:
as i told you before felipe, today I dont have to much time for the lab, so i was doing it during the breaks. :)

ill try to finish it when im back

PR #3457, opened 2022-11-24
WDRMTFTNOV2002 - Dave Mullally

David-Mullally:


Dengo777:
some of my solutions worked when I tested  them in CodePen but not in the lab
   -Bonus: Calculate the average of a mixed elements array Return the average on a mixed array
   -last two steps of Bonus Quest 
not sure why..

PR #3498, opened 2023-01-12
[FT-WD-Berlin-W1D4-lab] Indrajeet Patil

IndrajeetPatil:


src/functions-and-arrays.js
```diff
@@ -1,86 +1,200 @@
...
+  }
 
-// Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+  if (words.length === 1) {
+    return words[0];
+  }
 
-function findLongestWord() {}
+  let wordLengths = words.map((word) => {
```
mloelkes, line 22: I would be more efficient to store the word in a string to avoid having multiple loops

src/functions-and-arrays.js
```diff
@@ -1,86 +1,200 @@
...
+
+  let nonAllowedObjects = uniqueObjects.filter((item) => {
+    return item !== "number" && item !== "string" && item !== "boolean";
+  });
+
+  if (nonAllowedObjects.length > 0) {
+    throw new Error("Unsupported data type sir or ma'am");
+  }
+
+  let sumNum = mixedArray
```
mloelkes, line 66: Also here it would be more efficient to have one loop with multiple conditions than multiple loops

src/functions-and-arrays.js
```diff
@@ -1,86 +1,200 @@
...
+  if (wordsArray.length === 0) {
+    return null;
+  }
 
-function averageWordLength() { }
+  let wordLengths = wordsArray.map((word) => {
+    return word.length;
+  });
+
+  return averageNumbers(wordLengths);
```
mloelkes, line 124: Clever solution! 👍

src/functions-and-arrays.js
```diff
@@ -125,6 +260,6 @@ if (typeof module !== 'undefined') {
     uniquifyArray,
     doesWordExist,
     howManyTimes,
-    greatestProduct
+    greatestProduct,
   };
 }
```
mloelkes, line 265: Very good solutions! Keep up the good work! 🙂

IndrajeetPatil: Thanks a lot for the comments and the praise! 😄  I will update the solutions accordingly.

PR #3499, opened 2023-01-12
[BER-WDFT-012023] Neil Pilarca

Neilly28:
Pull request for LAB - JS Functions & Arrays

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -71,15 +71,12 @@ function sum(mixedArr) {
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
 function averageNumbers(numbersAvg) {
-  let total = 0;
+  // let total = 0;
   let avg = 0;
   if (numbersAvg.length == 0) {
     return null;
   }
-  for (let i = 0; i < numbersAvg.length; i++) {
-    total += numbersAvg[i];
-    avg = total / numbersAvg.length;
-  }
+  avg = sumNumbers(numbersAvg) / numbersAvg.length;
   return avg;
 }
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,119 +1,294 @@
...
-
+function sum(mixedArr) {
+  let total = 0;
+  // if (typeof mixedArr !== "object") {
+  //   return null;
+  // }
+  for (let i = 0; i < mixedArr.length; i++) {
+    if (typeof mixedArr[i] === "string") {
+      total += mixedArr[i].length;
+    } else if (typeof mixedArr[i] === "number") {
```
mloelkes, line 55: You could  also treat true and false like numbers:
`else if (typeof mixedArr[i] === "number" || typeof mixedArr[i] === "boolean")`

src/functions-and-arrays.js
```diff
@@ -1,119 +1,294 @@
...
-function averageWordLength() { }
+function averageWordLength(wordsArr) {
+  let total = 0;
+  let avg = 0;
+  if (wordsArr.length == 0) {
+    return null;
+  }
+  for (let i = 0; i < wordsArr.length; i++) {
+    total += wordsArr[i].length;
+    avg = total / wordsArr.length;
```
mloelkes, line 105: You should calculate the average after the loop is finished calculating the total.

src/functions-and-arrays.js
```diff
@@ -1,119 +1,294 @@
...
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(mixedArr) {
+  let total = 0;
+  let avg = 0;
+  if (mixedArr.length == 0) {
+    return null;
+  }
+  for (let i = 0; i < mixedArr.length; i++) {
```
mloelkes, line 117: You could use your function sum() here.

src/functions-and-arrays.js
```diff
@@ -125,6 +300,6 @@ if (typeof module !== 'undefined') {
     uniquifyArray,
     doesWordExist,
     howManyTimes,
-    greatestProduct
+    greatestProduct,
   };
 }
```
mloelkes, line 305: Very good solutions! Keep up the good work! 💪

PR #3501, opened 2023-01-12
WDFT Berlin Tobias Krauß

tobkrauss:


src/functions-and-arrays.js
```diff
@@ -52,14 +90,37 @@ const wordsUnique = [
...
-function doesWordExist() {}
+function doesWordExist(array,value) {
+  if (array.length === 0){
+    return null
+  }
+  for (let i = 0; i < array.length; i++) {
+   if (array.indexOf(value) != -1) {
+      return true
+    }
+    else if (array.indexOf(value) == -1) {
```
mloelkes, line 119: You return false already in the first iteration of your loop. You should only return false after the loop is finished.

src/functions-and-arrays.js
```diff
@@ -78,7 +139,15 @@ const wordsCount = [
...
-function howManyTimes() {}
+function howManyTimes(array,value) {
+  let count = 0
+  for (let i = 0; i < array.length; i++){
+    if(array[i] == value){
+      count++
+    }
+  }
+  return count
+}
```
mloelkes, line 153: Very good solutions! Keep up the good work! 👍

PR #3503, opened 2023-01-12
WDFT Paul Dombrowski - work in progress

PaulDombrowski:


src/functions-and-arrays.js
```diff
@@ -1,42 +1,96 @@
...
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
 
+function sumNumbers(numbers) {
+  
+  let sum = 0;
+
+  if (numbers === []) {
```
mloelkes, line 37: To check if the array is empty, you can use:
`numbers.length === 0`

src/functions-and-arrays.js
```diff
@@ -52,9 +106,21 @@ const wordsUnique = [
...
+  for (const item of unique) {
+    if (!result.includes(item)) {
+      result.push(item); 
+    }
+    return result
+  }
+}
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
```
mloelkes, line 126: Very good solutions! Keep up the good work! 👍

PR #3506, opened 2023-01-12
WDFT BERLIN Luke List

lukelist:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,102 @@
...
+let average = 0;
+
+if (wordsArr.length == 0){
+
+return null;
+}
+
+  for (let i = 0; i < wordsArr.length; i++){
+    wordLength = wordsArr[i].length + wordLength;
+   average = wordLength / wordsArr.length;
```
matigobbi, line 95: You can put this line outside of the loop and would be the same, :)

matigobbi (commented):
Good work Luke !
Keep going 😁

PR #3507, opened 2023-01-12
[WDFT Berlin] Alexis Aguilera

arekiu:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,60 @@
...
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arrayOfWords) {
+  let longestWord = "";
+
+  if(arrayOfWords == ""){
```
mloelkes, line 22: Here you compare the array with an empty string. To check if the array is empty:
`arrayOfWords.length === 0`

src/functions-and-arrays.js
```diff
@@ -26,13 +67,40 @@ function sum() {}
...
+
+  let average = 0;
+
+  if(numbersAvg==""){
+    return null;
+  }
+
+  else{
+    for (i=0; i<numbersAvg.length; i++){
+      average += numbersAvg[i];
```
mloelkes, line 80: Here you are calculating the sum, not the average, so you could use a different variable with a more explicit name, like `total`.

src/functions-and-arrays.js
```diff
@@ -78,7 +175,18 @@ const wordsCount = [
...
+  let contador = 0;
+
+  for (let i =0; i<array.length; i++){
+    if(word===array[i]){
+      contador++
+    }
+  }
+  return contador;
+
+}
```
mloelkes, line 190: Very good solutions, keep up the good work! 🙂

PR #3509, opened 2023-01-12
WDJan09 - Solved lab - Cherian

romlimoso:


matigobbi (commented):
Nice work Cherian, 
you used many loops and methods, also you put effort to solve the last one!
keep going  🚄⭐

PR #3510, opened 2023-01-12
[WD FT] Solved lab Gülsah Bayazit

gulsahbayazit:


src/functions-and-arrays.js
```diff
@@ -1,119 +1,281 @@
...
-
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
+    return a;
+  }
+  if (a < b) {
+    return b;
+  } else {
+    return a;
+    return b;
```
mloelkes, line 10: You don't need the second return statement, because the function will end with the first one.

src/functions-and-arrays.js
```diff
@@ -1,119 +1,281 @@
...
-function sum() {}
-
-
+const mixedArr = [6, 12, "miami", 1, true, "barca", "200", "lisboa", 8, 10];
+
+// should return: 57
+function sum(mixedArr) {
+  let sum = 0;
+  for (let i = 0; i < mixedArr.length; i++) {
+    if (typeof mixedArr[i] === "number") {
```
mloelkes, line 56: You could also combine the check for number and boolean in one condition

src/functions-and-arrays.js
```diff
@@ -1,119 +1,281 @@
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
+function averageNumbers(numbersAvg) {
+  let sum = 0;
+  if (numbersAvg.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < numbersAvg.length; i++) {
```
mloelkes, line 78: You could also use your sumNumbers-function instead of creating a loop.

src/functions-and-arrays.js
```diff
@@ -1,119 +1,281 @@
...
+  let sum = 0;
+  if (mixedArr.length == 0) {
+    return null;
+  }
+  for (let i = 0; i < mixedArr.length; i++)
+    if (typeof mixedArr[i] === "number") {
+      sum += mixedArr[i];
+    } else if (typeof arr[i] === "string") {
+      sum += mixedArr[i].length;
+    } else if (typeof mixedArr[i] === "true") {
```
mloelkes, line 122: Instead of checking for type "true" or "false" you can just check for "boolean"

src/functions-and-arrays.js
```diff
@@ -1,119 +1,281 @@
...
+  "communion",
+  "simple",
+  "bring",
 ];
 
-function uniquifyArray() {}
-
-
+function uniquifyArray(wordsUnique) {
+  if (wordsUnique == "") {
```
mloelkes, line 146: To check for an empty array, you can use `wordsUnique.length === 0`

src/functions-and-arrays.js
```diff
@@ -125,6 +287,6 @@ if (typeof module !== 'undefined') {
     uniquifyArray,
     doesWordExist,
     howManyTimes,
-    greatestProduct
+    greatestProduct,
   };
 }
```
mloelkes, line 292: Very good solutions! 👍 Keep up the good work!

PR #3512, opened 2023-01-12
[WDFT Berlin] Henrik Hellem

TheHellem:
Lab solved. Coming back to the 8.1 iteration.

The idea I have here is to transpose the matrix, and then check the sums of adjacent values in each row in the original matrix and the transposed one. I have written some function to aid this, but it is not yet complete.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,41 +1,92 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+const maxOfTwoNumbers = (a,b) => a > b ? a : b;
 
 // Iteration #2: Find longest word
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,92 @@
...
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
-
+function findLongestWord(arrayOfWords) {
+  if (arrayOfWords.length === 0){
+    return null;
+  } else {
+    let longestWord = arrayOfWords.sort(function(a,b) {
```
matigobbi, line 11: nice use of the sort function!

matigobbi (commented):
Nice work Henrik, you used many loops and methods, ;)
keep going ! 🚄⭐

PR #3514, opened 2023-01-12
Ulkar.Isayeva -WD-FT-Jan23-function

ulkarisayeva:
Lab function and arrays


matigobbi (commented):
Nice use of the loops Ulkar!
keep going ✨

PR #3520, opened 2023-01-17
[LISBFT0123] Tomás Belmar 

TomBelmar:


src/functions-and-arrays.js
```diff
@@ -52,14 +91,31 @@ const wordsUnique = [
...
+}
+console.log (uniquifyArray);
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(wordArray, searchWord) {
+  if(wordArray.includes(searchWord)) {
```
lzaquine, line 111: You also missed checking if it receives an empty array when called.

`if(!wordArray.length) return null;`

src/functions-and-arrays.js
```diff
@@ -52,14 +91,31 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
-
+//function uniquifyArray()
+function uniquifyArray(words) {
```
lzaquine, line 95: You were almost there on this one here!! Instead of having uniquifyArray(wordsUnique) as uniqueWords, you should've passed an []! And you also forgot to return null.

A solution based on yours:

```
function uniquifyArray(words) {
let uniqueWords = [];
  if (!words.length){
    return null;
  }
  for(let i = 0; i < words.length; i++) {
    if(uniqueWords.indexOf(words[i]) === -1){
      uniqueWords.push(words[i])
    }
  }
  return uniqueWords;
}
```

src/functions-and-arrays.js
```diff
@@ -26,16 +48,33 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers() {
```
lzaquine, line 51: Again here, almost there.

A solution based on yours:

```
let sum2 = 0;
if (!numbersAvg.length) {
    return null;
  } else {
for (let i = 0; i < numbersAvg.length; i++){
  sum2 += numbersAvg[i];
}}
return sum2 / numbersAvg.length;
}
```

lzaquine (commented):
Good job on the lab, Tomas! Please look at the comments I sent you for a better understanding.

PR #3523, opened 2023-01-17
[LISFT0123] Déborah 

deborahdelimaa:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,51 @@
...
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers( num1, num2) {
+  
+  if ( num1 > num2) {
+    return num1;
+  } else if ( num1 < num2 ) {
+return num2;
+  }
+  else {
```
webLisbonTA, line 9: Here you could just return one (num1 or num2) because they will have the same value.

src/functions-and-arrays.js
```diff
@@ -52,14 +103,31 @@ const wordsUnique = [
...
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(arr) {
+  if(arr.length === 0){
+    return null
+  }else{
+    let res = []
+    arr.forEach(rep => {
+      if ( res.indexOf(rep) == -1)
```
webLisbonTA, line 112: Attention to the use of double equal signs. It is better pratice if you use '==='.

webLisbonTA (commented):
Good job on the lab. Just check the comments I left for a better understanding.

PR #3524, opened 2023-01-17
[LISFT0123] Francisco Pereira

FranciscoManuelPereira:
-LAB | JS Functions & Arrays
-Done

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,26 +1,52 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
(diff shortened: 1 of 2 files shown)


src/functions-and-arrays.js
```diff
@@ -1,119 +1,197 @@
...
+  "aviator",
+  "crocodile",
+  "pearl",
+  "orchard",
+  "crackpot",
+];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  if (!arr) return null;
```
joaopdg, line 23: An `array` even empty it is equal to `true` so this function will never return `null` the only thing you need to do here to correct it is to check for the `length` of the `array` instead of the `array` it self:

`if (!arr.length) return null;`

joaopdg (commented):
Hi Francisco, try to finish the lab later this week so you can practice a bit more. Overall good job so far!

PR #3525, opened 2023-01-17
All Tasks completed

Assebe:


src/functions-and-arrays.js
```diff
@@ -1,42 +1,83 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(x, y) {
+  if(x > y){
```
webLisbonTA, line 3: Here, you were not passing the jasmine tests because you were returning more that what was supposed to.
The logic is all there, but you just needed to return the values of the arguments.

```
function maxOfTwoNumbers(x, y) {
  if(x > y){
    return x
  } else {
    return y
  }
}
```

src/functions-and-arrays.js
```diff
@@ -52,14 +93,31 @@ const wordsUnique = [
...
+console.log(uniquifyArray(wordsUnique))
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(arr, word) {
+  for(let i = 0; i < arr.length; i++){
```
webLisbonTA, line 113: The logic is ok, but you are returning a string and the tests ask for Booleans (true or false). Also, you need to return 'null' when receiving an empty array.
Here's a solution you can use:

```
  if(arr.length === 0) return null;
  for(let i = 0; i < arr.length; i++){
    if(arr.includes(word)){
      return true
    } else return false
  }
}
```

webLisbonTA (commented):
Good job on the lab. Just try to pay attention to the returns of the functions. The tests tell you what you need to return, otherwise it will keep appearing 'red' when you did the logic right.

PR #3529, opened 2023-01-17
[LIS-FT-01-23] Duarte França

Duartefranca:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,117 @@
...
+
+
+
+// Bonus - Iteration #4.3
+function avg(arr) {
+  if(arr.length === 0){
+    return null
+  }
+  else{
+    return sum(arr) / arr.length
```
webLisbonTA, line 110: Great use of a previous function. Saves lines of code and makes the function cleaner.

webLisbonTA (commented):
Good job on the lab. Just check the comments I made for a better understanding.

PR #3530, opened 2023-01-17
[LISFT0123] Vanessa Vieira

vanessav22:
Solved lab

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,21 +1,43 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
 
-
+function maxOfTwoNumbers(a, b) {
+  return Math.max(a,b);
+}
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,21 +1,43 @@
...
+  return Math.max(a,b);
+}
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
 
+  if (arr.length === 0) return null;
```
webLisbonTA, line 12: another way of doing that is:

`if (!arr.length) return null;`

webLisbonTA (commented):
Great job on the lab Vanessa!

PR #3534, opened 2023-01-17
Lab2

marianamedeirosfranco:


src/functions-and-arrays.js
```diff
@@ -1,21 +1,41 @@
...
+    return num2;
+  }
+  }
+console.log(maxOfTwoNumbers(2,4));
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(words) {
+  if (words==0)return null;
```
lzaquine, line 14: Another way of checking that is:

`if(!words.length) return null;`

lzaquine (commented):
Great job on the lab, Mariana!

PR #3536, opened 2023-01-17
[LISFT0123]Tiago Rato add lab project week 1 day 2

TiagoRato13:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,130 @@
...
 
+console.log(maxOfTwoNumbers(13, 3))
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  if(arr.length === 0) return null;
```
lzaquine, line 17: another way of doing that is:

`if(!arr.length) return null;`

lzaquine (commented):
Great job on completing the lab and bonuses, Tiago! Really good work!

PR #3538, opened 2023-01-17
finish week 1 day 2 lab

evelinecoen:


src/functions-and-arrays.js
```diff
@@ -121,15 +121,30 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(arr) {}
```
joaopdg, line 124: The logic is pretty much what you've done here, but you don't need to use `.concat()` neither `.sort()` you just need to `.push()` the items into the new `array` if that `word doesn't exist` in the array:

```
function uniquifyArray(arr) {
  if (arr.length === 0) {
    return null;
  }
  let uniqueArr = [];
    if (!uniqueArr.includes(arr[i])) {
      uniqueArr.push(arr[i]);
    }
  }
  return uniqueArr;
}
```

src/functions-and-arrays.js
```diff
@@ -121,15 +121,30 @@ const wordsUnique = [
...
+return values.sort ()
+}
+*/
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(word) {
```
joaopdg, line 142: So we know that this `function` will accept two `arguments` the first one will be the `array` of words, and the other one will be the specific `word` that you want to check if it exists in the array. So what you will need to do is, create a `loop` that will run over the `array` and in which `element` it will check if `word` is equal to the current `element` of the array:

```
  if (array.length === 0) {
    return null;
  }
    if (array[i] === word) {
      return true;
    }
  }
  return false;
}
```

joaopdg (commented):
Great job on the lab Eveline, next time try to do the bonus!

PR #3540, opened 2023-01-17
[LISFT0123] MIGUEL JESUS

MJesus94:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,172 @@
...
+  return soma;
+}
 
 
 
 // Iteration #3.1 Bonus:
-function sum() {}
+const newArr = [6, 12, 'miami', 1, true, 'barca', '200', 'lisboa', 8, 10];
+
+function sum(arr) {
```
lzaquine, line 61: Good job here, but one more straightforward way of doing this is:

```
function sum(arr) {
if (!arr) return null;
  let sum = 0;
  for (let el of arr) {
    if (typeof el === 'string') sum += el.length;
    else sum += el;
  }
  return sum;
}
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,172 @@
...
+    if(typeof array[i] === "number"){
+      arrayAvg += array[i];
+    } else if (typeof array[i] === "string"){
+      arrayAvg += array[i].length;
+    } else {
+      arrayAvg += 1;
+    }
+  }
+  arrayAvg /= 10;
+  return arrayAvg;
```
lzaquine, line 166: Here, the idea was to return the average of the array.
You can use the sum() function from before for that.

`return sum(array)/array.length;`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,172 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2) {
+    return number1;
+  } else {
+    return number2;
```
lzaquine, line 6: You forgot one step here, "should return greater of two arguments - if the second argument greater".
I see that you returned num2 later but the idea here was to check,

```
if(num1 < num2){
   return num2;
```
and then, return num1 or num2, like:

```
else {
return num1, num2;
}
```

lzaquine (commented):
Great job on the lab, Miguel!

PR #3541, opened 2023-01-17
 couldn´t do 6 and 7

Farid-97:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,61 @@
...
 
 
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers(array) {
+  let sum =0
+  if (array.length==="")
```
joaopdg, line 44: The property `.length` will return you a number, so it is better to give it the value of `0` instead of an empty string:

`if ( array.length === 0 )`

src/functions-and-arrays.js
```diff
@@ -26,13 +68,37 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(numbersAvg) {
```
joaopdg, line 71: Here you don't need the variable `average` you can return the `sum` divided by the `length` of the array:

```
  let sum = 0;

  if (!numbersAvg.length) {
    return null;
  }

  for (let i = 0; i < numbersAvg.length; i++) {
    sum += numbersAvg[i];
  }

  return sum / numbersAvg.length;
}
```

src/functions-and-arrays.js
```diff
@@ -26,13 +68,37 @@ function sum() {}
...
+      return average/numbersAvg.length;
+    };
+
 
 
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(wordsArr) {
```
joaopdg, line 89: Here same thing, you can return the `total` divided by the `length` of the array `wordsArr` straight away:

```
  let total = 0;

  if (!wordsArr.length) {
    return null;
  }

    total += wordsArr[i].length;
  }

  return total / wordsArr.length;
}
```

src/functions-and-arrays.js
```diff
@@ -52,14 +118,35 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(wordsUnique) {
```
joaopdg, line 121: Here you are `returning` the `first word` that `repeats` in the `array`. But what they ask you in the lab instructions is to `return` a new `array` without the `repeated words`. So what you need to do is create a `loop` that will run throughout all the `items` of the `array`  and `.push()` the items into the new `array` if that `word doesn't exist` in the new array yet:

```

  if (array.length === 0){
    return null;
  }

  let uniqueArray = [];

  for (let i=0; i<array.length; i++){
    if (!uniqueArray.includes(array[i])){
      uniqueArray.push(array[i]);
    }
  }

    return uniqueArray;
}
```

src/functions-and-arrays.js
```diff
@@ -52,14 +118,35 @@ const wordsUnique = [
...
+}
+
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(wordsFind) {
```
joaopdg, line 144: So we know that this `function` will accept two `arguments` the first one will be the `array` of words, and the other one will be the specific `word` that you want to check if it exists in the array. So what you will need to do is, create a `loop` that will run over the `array` and in which `element` it will check if `word` is equal to the current `element` of the array:

```
  if (array.length === 0) {
    return null;
  }
    if (array[i] === word) {
      return true;
    }
  }
  return false;
}
```

joaopdg (commented):
Good job on the lab Farid, next time try to do the bonus!

PR #3542, opened 2023-01-17
Solved Lab - :( tired to make the bonus

erikasoaress:


src/functions-and-arrays.js
```diff
@@ -1,42 +1,99 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+
+function maxOfTwoNumbers( n1, n2) {
+  if (n1>n2) {
+    return n1;
+  }  else if (n2>n1) {
+     return n2;
+  } else {
+    return n1,n2;
```
webLisbonTA, line 9: Here you can just return one of the arguments (n1 or n2) as they have the same value.

src/functions-and-arrays.js
```diff
@@ -1,42 +1,99 @@
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
-
-
+function averageNumbers(arr) {
+  let total = 0; 
+  for (let i = 0; i < arr.length; i++) {
+    total += arr[i];
+  } if (arr === []) {
```
webLisbonTA, line 71: Here, you can't compare the argument (arr) to that empty array '[]' as it will not work. To check if an empty array is called, you can use: `if (arr.length === 0) return null;` or `if (!arr.length) return null`

src/functions-and-arrays.js
```diff
@@ -52,15 +109,33 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
-
+function uniquifyArray(array) {
+  if (array.length === 0 ) {
+    return null;
+  } for ( let i = 0; i < array.length ; i++) {
+    if (array.indexOf(array[i]) === array.lastIndexOf(array[i])) {
```
webLisbonTA, line 116: Here you should've use another approach. 
Maybe you could check if there's an existing matching element and work over it. 
```
let uniqueArray = [];
    if (!uniqueArray.includes(arr[i])) {
      uniqueArray.push(arr[i]);
    }
  }
  return uniqueArray;
```

webLisbonTA (commented):
Good job on the lab. Please check the comments I left for a better understanding.

PR #3543, opened 2023-01-17
[LISFT0123] Wagner

waggnerl:


src/functions-and-arrays.js
```diff
@@ -1,119 +1,261 @@
...
+  if (arrayData.length < 1) return null;
+
+  let i = 0;
+  arrayData.forEach((el) => (i += el.length));
+  return i / arrayData.length;
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(arrayData) {
```
joaopdg, line 111: This iteration is similar to the 3.1 you will receive a `mixed elements array` so you need to make `conditions` checking the `typeOf` the elements:

```
function avg(array) {
  if (array.length === 0){
    return null;
  }
  let sum = 0;
  for (let i=0; i<array.length; i++){
    if (typeof array[i] === 'number'){
      sum += array[i];
    } else if (typeof array[i] === 'string'){
      sum += array[i].length;
    }else if (array[i] === true){
      sum += 1;
    }
  }
  return sum/array.length
}
```

src/functions-and-arrays.js
```diff
@@ -1,119 +1,261 @@
...
+  "disobedience",
+  "matter",
+  "matter",
+  "matter",
 ];
 
-function howManyTimes() {}
-
-
+function howManyTimes(arrayData, wordToSearch) {
```
joaopdg, line 178: This function is not working if the word repeats more than one time. A way of doing this is by creating a `loop` that will check if the specific `word` exists in the `array` and if it does `.push()` it inside a new `array`:

```
  if (array.length === 0){
    return 0;
  }
  let newArray = [];
  for (let i=0; i<array.length; i++){
    if (array[i] === word){
      newArray.push(array[i])
    }
  }
  return newArray.length
}
```

joaopdg (commented):
Great job on the lab and the bonus Wagner!

src/functions-and-arrays.js
waggnerl: man, try this code:
`const wordsCount = [
  "machine",
  "matter",
  "subset",
  "trouble",
  "starting",
  "matter",
  "eating",
  "matter",
  "truth",
  "disobedience",
  "matter",
  "matter",
  "matter",
];

function howManyTimes(arrayData, wordToSearch) {
  if (arrayData.length === 0) return 0;
  if (arrayData.indexOf(wordToSearch) === -1) return 0;
  let tst = arrayData.filter((el, index) =>
    el === wordToSearch ? arrayData.indexOf(wordToSearch) !== index : ""
  );
  if (tst.length > 0) {
    console.log(tst.length);
    return tst.length;
  }
  if (tst.length === 0) return 1;
}
howManyTimes(wordsCount, "matter");`
I think that works when the word repeats more than one time!

PR #3544, opened 2023-01-17
Marisha Deroubaix WebDev 2023 Full Time Lisboa

Deroubaix:


src/functions-and-arrays.js
```diff
@@ -1,21 +1,50 @@
...
+//console.log(maxOfTwoNumbers(16, 34 ))
 
 
 // Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+
+function findLongestWord(arr) {
+  if (arr.length === 0) return null
```
lzaquine, line 21: Another way of checking that is:

`if(!arr.length) return null;`

lzaquine (commented):
Great job on the lab, Marisha! Try doing the bonuses next time!

PR #3545, opened 2023-01-18
[LISFT0123] André Lisboa

andrelisboapt:
I didn't understand the last bonus so it's missing.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -58,8 +58,30 @@ sumNumbers(numbers)
 
 
 // Iteration #3.1 Bonus:
-function sum() {}
+function sum(arr) {
+  if(arr.length == 0) return 0;
+    
+  let total = 0;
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -52,14 +160,38 @@ const wordsUnique = [
...
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(arr, word) {
+  if(arr.length == 0) return null;
+  for(let i = 0; i < arr.length; i++){
```
webLisbonTA, line 184: Nice logic here. You also could return the method to simplify your answer.
Instead of the for loop. You can use the includes() method.

`return arr.includes(word);`

Keep in mind that you did it correct. It's just an easier way of doing it.

PR #3546, opened 2023-01-18
[LISFT0123] Rafaela Umbelino

RafaelaUmbelino:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,64 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 > num2){
+  return num1;
 
+  }else{
```
lzaquine, line 6: You forgot one step here, "should return greater of two arguments - if the second argument greater".
I see that you returned num2 later but the idea here was to check,
```
if(num1 < num2){
   return num2;
```
and then, return num1 or num2, like:
```
else {
return num1, num2;
}
```

src/functions-and-arrays.js
```diff
@@ -1,19 +1,64 @@
...
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+let longestWord = "";
+
+function findLongestWord(arr) {
 
-function findLongestWord() {}
+  if (!arr.length) {
```
lzaquine, line 22: Great job here!

src/functions-and-arrays.js
```diff
@@ -52,16 +127,51 @@ const wordsUnique = [
...
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(arr) {
+
+  if (!arr.length){
+    return null;  
+  }
+  
+  let noRepititionsArray = [];
```
lzaquine, line 136: Great job on creating this variable and then using it later!

lzaquine (commented):
Great job on the lab, Rafaela! Try doing the bonuses next time!

PR #3547, opened 2023-01-18
[LISFT0123] Pedro Leal

PeterLoyal:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,61 @@
...
 
 
-// Iteration #2: Find longest word
-const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+// Iteration #2: Find longest word
+const arr = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+function findLongestWord(arr) {
```
lzaquine, line 21: This got a bit confusing. A better way to write this function would be something like this:

```
  let longWord = "";
  if (!arr.length) {
    return null;
  } else if (arr.length === 1) {
    return arr[0];
  } else {
    if (arr[i].length > longWord.length) {
      longWord = arr[i]
    } 
  } return longWord; 
}
}
```

lzaquine (commented):
Great job on the lab, Pedro! Please take a look at the comments I sent for a better understanding.

PR #3548, opened 2023-01-18
[LISFT0123] Caio Mazur

caiomazur:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,137 @@
...
+    sum += arrayOfStrings[i].length;
+  }
+  return sum / arrayOfStrings.length;
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+// Same as Iteration 3.1 but without the error and with average.
+function avg(array) {
+  if(array == 0) return null;
```
webLisbonTA, line 114: Always use '===' instead of '=='. And here, you were checking for the actual 0 and not an empty array. You forgot the length. `if(array.length === 0) return null;`

src/functions-and-arrays.js
```diff
@@ -78,11 +191,26 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
+
+function howManyTimes(array, word) {
+  if (array == 0) return 0;
```
webLisbonTA, line 196: Attention to the equal signs. Use '==='

webLisbonTA (commented):
Good job on the lab. Check the comments I left you for a better understanding.

PR #3574, opened 2023-01-19
[PAR-FT-012023]VARNIER Hiromi

rominamarmontel:


src/functions-and-arrays.js
```diff
@@ -106,7 +269,40 @@ const matrix = [
...
+        max = result;
+        }
+      }
+      if ((i - 3) >= 0 && (j - 3) >= 0) {
+        result = arr[i][j] * arr[i - 1][j - 1] * arr[i - 2][j - 2] * arr[i - 3][j - 3];
+        if (max < result) {
+          max = result;
+        }
+      }
+      if ((i - 3) >= 0 && (j - 3) <= 0) {
```
Chaudoreille, line 296: your j condition seems off.
looks like it should be (j + 3) < arr[i].length

PR #3650, opened 2023-02-23
[RMT FT 20Feb2023] Bartosz Dubinski

bartoszde:
It takes a lot of time. I was looking for help in prework materials, youtube and other.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -119,17 +119,28 @@ const wordsUnique = [
 ];
 
 function uniquifyArray(array) {
+  if (array.length === 0) {
+    return null
+  } else {
     return array.filter((el, index) => array.indexOf(el) === index);
   }
-  
+}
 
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Bartosz,
We all get stuck, or need time! but great you were able to find solutions on your own.
I see you used .filter() in iteration 5 (uniquify array). just want to be sure you got how it works😉
another possible solution could be:
1. initialize the arrayWhithoutRepetions   (that is what we wanna return at the end) --> as an empty array
2. loop inside the givenArray (with a for / forEach ...)
3. check if  the givenArrayWord is already inside arrayWhithoutRepetions (here you can use different methods ex. .includes or indexOf) 
4. if is already in --> go further with the loop
5. if is not --> add the word in arrayWhithoutRepetions ( push  can be a good method to use )
Nice job, keep coding🙌

PR #3673, opened 2023-03-23
Berlin-WDFT-032023-Muhammad Hussain Kamran-W1D4

MuhammadHussain1701:


B4n3l1ng (commented):
Very good! Keep in mind that typeof an array, doesn't return array, but object ;)

PR #3674, opened 2023-03-23
Bojana_BER_March_WD_FT

bojana-petroska:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,100 @@
...
+  for (let i=0; i<arr.length; i++) {
+    if (typeof(arr[i] === 'number')) {
+      result += arr[i];
+    } else if (typeof(arr[i] === 'string')) {
+      result += arr[i].length;
+    } else if (arr.length === 0) {
+      return 0;
+    } else if (arr.length === 1) {
+      return arr[0];
+    } else if (typeof(arr[i] === 'true')) {
```
B4n3l1ng, line 52: in this case, typeof would be "boolean" not the true or false keyword. the logic is correct, though :)

src/functions-and-arrays.js
```diff
@@ -1,41 +1,100 @@
...
 
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(numbers) {
+  let sum = numbers.reduce((a, b) => a + b, 0);
+  if (numbers.length === 0) return null;
```
B4n3l1ng, line 69: You could switch lines 68 and 69, so the function would finish early if the array length is 0, that way it wouldn't have to compute the num variable, making it slightly faster

src/functions-and-arrays.js
```diff
@@ -1,41 +1,100 @@
...
+  if (arr.length === 0) return null;
+  for (let elem of arr) {
+    if (typeof elem === 'number') {
+      sum += elem;
+    } else if (typeof elem === 'string') {
+      sum += elem.length;
+    }
+  }
+  return Math.round(sum / arr.length);
+}
```
B4n3l1ng, line 95: Just missing the unsuported data types error and the boolean case :)

src/functions-and-arrays.js
```diff
@@ -52,14 +111,22 @@ const wordsUnique = [
...
+  return unique;
+}
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(arr) {
```
B4n3l1ng, line 125: this function is supposed to receive an array and a word to search as arguments, right now it only receives an array.

PR #3677, opened 2023-03-23
PanagiotisLiorods.Ber.WDFT.March2023

panagiotisliordos:


src/functions-and-arrays.js
```diff
@@ -78,7 +190,26 @@ const wordsCount = [
...
+  }
+
+  if (count === 0) {
+    return 0;
+  } else if (count === 1) {
+    return 1;
+  } else {
+    return 5;
+  }
+}
```
arekiu, line 212: why is it returning 5 in the else statement? is enough just returning the value stored in count

arekiu (commented):
Good job!

src/functions-and-arrays.js
panagiotisliordos: That ask Jasmine . I was wondering and i try to deleted now ,but give the last part red.   
Screenshot from 2023-03-29 21-26-58

panagiotisliordos: so i think that needs the else {return 5}

PR #3678, opened 2023-03-23
Rafal Kolibaba WD BER March

RafalK01:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,126 @@
...
+
+  let currSum = 0
+  mixedArr.forEach(value => {
+   if(typeof value === "number"){
+     currSum += value
+   }
+    else if (typeof value === "string"){
+      currSum += value.length
+    }
+    else if(typeof value === "boolean"){
```
B4n3l1ng, line 61: A boolean has a number value. false, when added to a number, it's like adding 0. true, is +1

B4n3l1ng: Therefore, you don't need the second argument, you can just do if (typeof value==="boolean"){currSum += value}

B4n3l1ng: You're only missing the last part, which entails the typeof being object (array's typeof is object), and that should throw an error, like throw new Error("Unsuported data type sir or ma'am")

PR #3682, opened 2023-03-23
Week1Day4-Functions&Arrays-Ber-MAR23-Diana

Deezledee:
Done


src/functions-and-arrays.js
```diff
@@ -1,41 +1,147 @@
...
+function maxOfTwoNumbers(num1, num2) {
 
+  
+  if (num1 > num2) {
+    return num1
+  } else if (num2 > num1) {
+    return num2
+  } else {
+    return num1 && num2
+  }}
```
arekiu, line 11: you can simplify the else statement just returning one number or the other

arekiu (commented):
Excellent work! :D

PR #3683, opened 2023-03-23
BER_FT_03_23_23_HeimLi

HeimLi:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,96 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) { return num1 }
+  else { return num2 }
+}
+result = maxOfTwoNumbers(3, 5);
```
mloelkes, line 6: You don't need to call your function here to make it work. If you use it, to test your code, that's fine! 😉 But you should better remove these function calls before you submit the lab.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,96 @@
...
+
+function findLongestWord(words) {
+
+  let wordLength = [];
+  wordLength.push(words[0]);
+  //console.log(wordLength);
+  //console.log(typeof (wordLength));
+
+  for (let i = 0; i < words.length; i++) {
+    if (wordLength.length < words[i].length) {
```
mloelkes, line 22: Better store the longest word in a string, not an array and just compare the length by using longestWord.length

src/functions-and-arrays.js
```diff
@@ -1,41 +1,96 @@
...
+function averageNumbers(numbersAvg) {
+  let averageNumbersSum = 0;
+
+  for (let i = 0; i < numbersAvg.length; i++) {
+    averageNumbersSum = averageNumbersSum + numbersAvg[i]
+  }
+  return averageNumbersSum/2
+}
+
+let result4 = averageNumbers(numbersAvg);
```
mloelkes, line 71: You should divide the total sum by the amount of numbers in the array to get the average

src/functions-and-arrays.js
```diff
@@ -52,15 +107,37 @@ const wordsUnique = [
...
 ];
 
-function uniquifyArray() {}
+function uniquifyArray(wordsUnique) {
+
+  let newArray = [];
+
+  for (let i = 0; i < wordsUnique.length; i++) {
+
+    if (newArray.includes(wordsUnique[i])) {
```
mloelkes, line 116: You could also check if the array doesn't include the word by using `!newArray.includes(wordsUnique[i])`, then you don't need an else

src/functions-and-arrays.js
```diff
@@ -52,15 +107,37 @@ const wordsUnique = [
...
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(wordsFind, givenWord) {
 
+  for (let i = 0; i < wordsFind.length; i++) {
+    if (wordsFind.includes(givenWord)) {
+      console.log("true"); break
```
mloelkes, line 135: Good solution, you just need to replace the console.log with a return-statement

src/functions-and-arrays.js
```diff
@@ -106,25 +195,26 @@ const matrix = [
...
+     sum,
+     averageNumbers,
+     averageWordLength,
+     avg,
+     uniquifyArray,
+     doesWordExist,
+     howManyTimes,
+     greatestProduct
+   };
+ }
```
mloelkes, line 220: Good solutions! ✨ Keep on the good work!

PR #3686, opened 2023-03-23
lab done

LeFreakALeak93:


B4n3l1ng:
Function avg() needs some work, sumNumbers() has an issue with variable and argument names, besides that, pretty nice job!

PR #3688, opened 2023-03-23
Gerrit Brinsat WD FT BER MAR23

GBrinsat:
Done

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,19 +1,67 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 > num2){
+    return num1
+  }
+  else if(num2 > num1){
+    return num2
+  }
+  else{return num1}
+}
+
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
 
+function findLongestWord(arr) {
+  if(arr.length === 0) {
+    return null
+  }
+  else if(arr.length === 1) {
+    return arr[0]
+  }
+  
+  let longestWordLength = arr[0].length
+  let longestWord = arr[0]
+
+  for(let i = 0; i < arr.length; i++) {
+    const comparetor = arr[i].length
+
+    if(comparetor > longestWordLength) {
+      longestWord = arr[i]
+      longestWordLength = comparetor
+    }
+    }
+    return longestWord
+  }
+
+  findLongestWord(words);
 
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,19 +1,67 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 > num2){
+    return num1
+  }
+  else if(num2 > num1){
+    return num2
+  }
+  else{return num1}
```
B4n3l1ng, line 9: You could simplify this function by instead of else if on the second member, doing just else { return num2}. Logically speaking, if num1 is not bigger then num2, it can only be smaller or equal, and seeing as when it's equal it doesn't matter what you return... good job overall, though :)

src/functions-and-arrays.js
```diff
@@ -1,19 +1,67 @@
...
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
 
+function findLongestWord(arr) {
+  if(arr.length === 0) {
+    return null
+  }
+  else if(arr.length === 1) {
```
B4n3l1ng, line 23: No need to check for a length of 1, because your logic will work for a single word, it will return itself

src/functions-and-arrays.js
```diff
@@ -1,19 +1,67 @@
...
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+
+function sumNumbers(numbersArr) {
+  if(numbersArr.length === 0) {
+    return 0
+  }
+  else if(numbersArr.length === 1) {
```
B4n3l1ng, line 52: Same thing, no need to check for length = 1

src/functions-and-arrays.js
```diff
@@ -25,14 +73,57 @@ function sum() {}
...
+    }
+    average = sum / arrAvg.length
+    return average
+  }
 
-function averageNumbers() {}
+
+}
+
+averageNumbers(numbersAvg)
```
B4n3l1ng, line 100: you don't need to call your functions in this lab, Jasmine calls them for you. If you want to test, you can console.log them with a test array

src/functions-and-arrays.js
```diff
@@ -52,14 +143,46 @@ const wordsUnique = [
...
-function uniquifyArray() {}
+function uniquifyArray(arrUni) {
+  if(arrUni.length === 0) {
+    return null
+  }
+  else {
+    let word
+    let uniqueArray = []
+    for(let i = 0; i < arrUni.length; i++) {
+      word = arrUni.splice(i, 1)
```
B4n3l1ng, line 154: Unecessary to splice the original array, when arrUni[i] inside the for loop would give you the word you need to analyze

B4n3l1ng (commented):
Good job overall 👍

PR #3689, opened 2023-03-23
finished lab | Shawn Wolter ber. mar. WD-FT

Cyber-SW:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,51 @@
...
-
-function findLongestWord() {}
+let words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
+
+function findLongestWord(arr) {
+  if (arr.length === 0) {
+    return null;
+  }
+  
+  let longestWord = arr[0].length;
```
mloelkes, line 20: If you want to use two variables, a better name for this variable would be longestWordLength or something similar.

PR #3690, opened 2023-03-23
Solved lab-LillyL-Ber-JS-functions&arrays

lillyfeee:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
+  let longestWord = " ";
+  for (let i = 0; i < words.length; i++){
+    if (words[i].length > longestWord.length){
+      longestWord = words[i];
+    }
+  }
+  return longestWord;
+}
 
+let result = findLongestWord(words)
```
mloelkes, line 27: You don't need to call your functions, because they will be called by the jasmin tests. If you want to test your code, that's fine. But normally you should remove these function calls before you submit your solution.

src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
+  let total = 0;
+
+  for (let i = 0; i < mixedArr.length; i++) {
+
+    if (typeof mixedArr[i] === 'string') {
+      total += mixedArr[i].length;
+    }
+    else if (typeof mixedArr[i] === 'number' || typeof mixedArr[i] === 'boolean') {
+      total += mixedArr[i];
+    }
```
mloelkes, line 58: There was a requirement to throw an error when the data type is not string, number of boolean. For this case you could add an else-statement here, in which you throw an error (throw new Error("Error message")

src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
+
 
 // Bonus - Iteration #4.1
-function avg() {}
+
+const mixedArr = [6, 12, 'miami', 1, true, 'barca', '200', 'lisboa', 8, 10];
+
+
+function avg(mixedArr) {
+  let total = 0;
```
mloelkes, line 102: You can use the sum-function here you created before

src/functions-and-arrays.js
```diff
@@ -1,41 +1,118 @@
...
+      total += mixedArr[i];
+    }
+  }
+
+  return total / mixedArr.length
+}
+console.log(avg(mixedArr));
 
 // Iteration #5: Unique arrays
 const wordsUnique = [
```
mloelkes, line 118: Good solutions! ✨ If you find the time, try to finish the missing iterations.

PR #3692, opened 2023-03-24
last changes

ChriSiebels:


src/functions-and-arrays.js
```diff
@@ -52,7 +101,18 @@ const wordsUnique = [
...
 
-function uniquifyArray() {}
+function uniquifyArray(words) {
+    for(let index = 0 ; index < words.length ; index++) {
+      let word = words[index]
+      let duplicate = words.indexOf(word, index+1)
+      if (duplicate !== -1) {  
+        words.splice(duplicate, 1)
+        words.splice(index, 1)
+        --index
```
arekiu, line 111: Achtung! You are deleting the duplicated word, but also the original one!

arekiu (commented):
The last iterations are missing, but what it´s done is super good!

PR #3694, opened 2023-03-27
lab-javascript-functions-and-arrays :Neethu Kunjappan

Neethu2022:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,156 @@
...
 
-function findLongestWord() {}
+function findLongestWord(wordsArr)
+ {
+  let longestWord = "";
+  if(wordsArr.length===0)
+  { 
+    return null;
+  } 
+  else if (wordsArr.length===1)
```
mloelkes, line 24: You don't need to add a specific logic in case the length of the array is 1, because the logic in the body of the else-statement is covering that.

src/functions-and-arrays.js
```diff
@@ -1,38 +1,156 @@
...
+    return 0;
+  } 
+  else if (numbers.length===1)
+  {
+     return numbers[0] ;
+  
+  }
+  else 
+  {
+  for (var i = 0; i < numbers.length; i++) {
```
mloelkes, line 63: Better use let or const instead of var to declare a variable.

src/functions-and-arrays.js
```diff
@@ -1,38 +1,156 @@
...
+  let result = 0;
+  if (numbersAvg.length === 0) 
+  return null;
+  else if (numbersAvg.length === 1) 
+  {
+  return numbersAvg/numbersAvg.length;
+ }
+  const average =numbersAvg.reduce((a, b) => a + b, 0) / numbersAvg.length;
+    return avg;
+  for (var i = 0; i < numbersAvg.length; i++)
```
mloelkes, line 125: This solution would also be fine, but you should do the division after the loop is finished and divide by the amount of numbers in the array.

src/functions-and-arrays.js
```diff
@@ -59,7 +177,10 @@ function uniquifyArray() {}
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist()
+{}
+ 
+
```
mloelkes, line 186: Good work! If you find the time, try to finish the missing functions. 👍

PR #3724, opened 2023-04-11
[LIS-WDFT-APR23] Pedro Nogueira

Pedro-No:


src/functions-and-arrays.js
```diff
@@ -1,68 +1,131 @@
...
-const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
-
-function doesWordExist() {}
-
-
-
+function doesWordExist(words,search) {
+  if (words.length === 0 || words.length === undefined) {
+    return null;
+  }else if(words.indexOf(search) >= 0) {
```
brunorocha20, line 122: Here, you could've used the `includes()` method. 
`return words.includes(search);` the output is going to be `true` or `false`

brunorocha20 (commented):
Good job on the lab Pedro.

PR #3725, opened 2023-04-11
LIS-WDFT-APR23-TOMÁS

SinTomas:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,62 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
```
brunorocha20, line 5: Tomás, here, is nothing wrong with the code. A suggestion I can make you is that you can save some lines of code, as we only need one `else`. If the `num1` is bigger, we want to return `num1`, right. If the `num2` is bigger we want `num2`, ok here, however, if `num1` and `num2` are equal, we can return either `num1` or `num2`. 
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

brunorocha20 (commented):
Good job on the lab Tomás.

PR #3726, opened 2023-04-11
Lis-WDFT-2023-Guy-Dagan

Guy-Dagan:


src/functions-and-arrays.js
```diff
@@ -26,16 +61,48 @@ function sum() {}
...
+             } 
+            }
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(avgArray) {
+  if (avgArray.length === 0){
+    return null;
+  }
+
```
joaopdg, line 104: For this iteration you need a loop that will run through all the elements of the array, and in each element you need to check the data type of that element and depending on the data type you do different operations

joaopdg (commented):
Great job on the lab Guy, next time try to finish all the bonus!

PR #3728, opened 2023-04-11
[LIS-WDFT-APR23] Joao Madeira

jodmadeira:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,151 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
+function maxOfTwoNumbers(number1,number2) {
+    if(number1>number2){
+      return number1
+    }
+    else if (number2>number1){
```
brunorocha20, line 6: João, here, is nothing wrong with the code. A suggestion I can make you is that you can save some lines of code, as we only need one `else`. If the `num1` is bigger, we want to return `num1`, right. If the `num2` is bigger we want `num2`, ok here, however, if `num1` and `num2` are equal, we can return either `num1` or `num2`. 
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,151 @@
...
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(avgArray) {
+  if(avgArray === null || avgArray === undefined || avgArray.length === 0){
+    return null;
+  }
+    else{
+  let count=0;
```
brunorocha20, line 97: Since you previously created a function to sum numbers, you should reuse it here instead of repeating the whole logic to sum the numbers.
```
let count = sumNumbers(avgArray);
return count / avgArray.length
```

brunorocha20 (commented):
Good work on the lab João.

PR #3729, opened 2023-04-11
[Functions-arrays-wdft-apr23] Robson Batista

RobBatista:
Work almost complete aside of the bonus assessments!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,24 +1,84 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
+    return num2;
+  }
+  else{
+    return num1;
+  }
+}
+
+
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(words) {
+  if (words.length === 0) {
+    return null;
+  } else if (words.length === 1) {
+    return words[0];
+  } else {
+    let longest = words[0];
+    for (let i = 1; i < words.length; i++) {
+      if (words[i].length > longest.length) {
+        longest = words[i];
+      }
+    }
+    return longest;
+  }
+}
+
+
 
 
 
 // Iteration #3: Calculate the sum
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,41 +1,148 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num2 > num1) {
```
brunorocha20, line 5: Robson, here, is nothing wrong with the code. A suggestion I can make you is that you can save some lines of code, as we only need one `else`. If the `num1` is bigger, we want to return `num1`, right. If the `num2` is bigger we want `num2`, ok here, however, if `num1` and `num2` are equal, we can return either `num1` or `num2`. 
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,148 @@
...
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers(numbers) {
+  if (numbers.length === 0) {
+    return 0;
+  }
+  return numbers.reduce((total, num) => total + num, 0);
```
brunorocha20, line 47: This `reduce` method can be used, however, is something a bit more advanced for this stage.
I think it'd be better if you use a simples `for` loop to iterate over the array of numbers and then adding them.
Something like this:
```
function sumNumbers(numbers) {
  let sum = 0;
  for (let i = 0; i < numbers.length; i++) {
    sum += numbers[i];
  }
  return sum;
}
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,148 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(numbers) {
+  if (numbers.length === 0) {
+    return null;
+  }
+  let sum = 0;
```
brunorocha20, line 93: We write functions because we can use them later, and since you created a function above to sum numbers, you should reuse it here, instead of writing again these steps to sum the numbers before doing that average.
```
  if (!numbers.length) return null;
  let sum = sumNumbers(numbers);
  return sum / numbers.length;
}
```

brunorocha20 (commented):
Good job on the lab Robson.

PR #3732, opened 2023-04-11
[LIS-FT-042023] Margarida Pereira

mapmargaridapereira:
Had a lot of trouble resolving this and as you can see some iterations are incomplete. Hopefully I can review these doubts and correct everything.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -18,25 +18,20 @@ maxOfTwoNumbers (30, 45);
 
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord(words) {
-  let longestWord = "";
-  if (words.length === 0) {
+function findLongestWord(arr) {
+  if (arr.length === 0) {
     return null;
-  }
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -26,13 +72,33 @@ function sum() {}
...
 
 
 // Level 2: Array of strings
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(wordsArr) {
+  let averageWordsArr = "";
+  if (wordsArr.length === 0) {
+    return null;
```
brunorocha20, line 99: In this iteration, the logic is close to the previous one. The main difference is that when we loop over the array, we want to get the `length` of the word that we are iterating over.
Something like this:
```
  if (wordsArr.length === 0) {
    return null;
    }
  let total = 0;
    total += wordsArr[i].length;
  }
  let average = total / wordsArr.length;
  return average;
}
```

brunorocha20 (commented):
Good job on the lab Margarida, please check the comments I left, for a better understanding.

PR #3733, opened 2023-04-11
Solved lab

chikifella7:


src/functions-and-arrays.js
```diff
@@ -1,122 +1,248 @@
...
 
-function averageNumbers() {}
+function averageNumbers(numArray) {
+  if (numArray.length === 0) {
+    return null;
+  }
+  let total = 0;
 
+  for (const sumOfNumbers of numArray) {
+    total += sumOfNumbers;
```
joaopdg, line 62: In the previous iteration you created a function that sums numbers (sumNumbers), so instead of having to repeat the same code for this new function (averageNumbers) you can actually use the previous function (sumNumbers):

```
function averageNumbers(numArray) {
  if (numArray.length === 0) {
    return null;
  }
  let total = sumNumbers(numArray);
  return total / numArray.length;
}
```

joaopdg (commented):
Great job on the lab Francisco, next time try to do the bonus!

PR #3735, opened 2023-04-11
Lis-WDFT-2023-MarisaPinheiro

Marisa-Pinheiro:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,147 @@
...
+
+
+function averageNumbers(theNumbers) {
+  let total = 0;
+  if (theNumbers.length === 0){
+    return null;
+  }
+  for (let i= 0; i < theNumbers.length ; i++){
+    if (theNumbers.length === 1){
+      return theNumbers[i];
```
brunorocha20, line 98: Since you wrote earlier a function to sum numbers, you can reuse it here, instead of repeating the whole logic to sum the numbers again. By calling the function you can do that.
Something like this: 
```
let sum = sumNumbers(theNumbers)
return sum / theNumbers.length
```

PR #3736, opened 2023-04-12
Lis-WDFT-2023 - Lucas Telles

lltelles:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,122 @@
...
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  if (arr.length === 0){
+    return null
+  } let avrArr = 0
```
brunorocha20, line 77: Since you previously created a function to sum numbers, you should reuse it here instead of repeating the whole logic to sum the numbers.
```
let avrArr = sumNumbers(arr);
return avrArr / arr.length
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,122 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1,num2) {
+  if (num1>num2){
+  return num1
+ } else if (num2>num1){
```
brunorocha20, line 5: Lucas, here, is nothing wrong with the code. A suggestion I can make you is that you can save some lines of code, as we only need one `else`. If the `num1` is bigger, we want to return `num1`, right. If the `num2` is bigger we want `num2`, ok here, however, if `num1` and `num2` are equal, we can return either `num1` or `num2`. 
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

src/functions-and-arrays.js
```diff
@@ -52,14 +133,41 @@ const wordsUnique = [
...
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(findElement, target) {
+  if (findElement.length === 0){
+    return null;
+  } else if (findElement.length === 1){
```
brunorocha20, line 160: Here, you can use an array method called `includes()`, that is going to check if the word exist inside the array that we iterate over. The output of that method is `true` or `false`.
So, in this case, we can just:
`return findElement.includes(target)`

brunorocha20 (commented):
Good job on the lab Lucas.

PR #3737, opened 2023-04-12
[LIS-WDFT-APR23] Henrique Pereira

Ricozzo:
I tried finish the task nº8 but i couldn't.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,21 +1,47 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1>num2) {
+    return num1;
+  }
+  else if (num1<num2) {
+    return num2;
+  }
+  else {
+    return num1, num2;
+  }
+}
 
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -78,7 +143,15 @@ const wordsCount = [
...
 
-function howManyTimes() {}
+function howManyTimes(arrayOrtiz) {
+  let nothing = 0;
+  for (let i = 0; i < arrayOrtiz.length; i++) {
+    if (arrayOrtiz[i] === wordsCount) {
+      nothing++;
+    }
+  }
+  return nothing;
```
joaopdg, line 153: For this iteration your function should receive two parameters: the word that you want to count, and the array where you want to search for the word.
Your code is not working because you only have one parameter in your function, and you're calling the name of the array inside the function (wordsCount) instead of calling the placeholder (wordToCount)

```
function howManyTimes(arrayToSearch, wordToCount) {
  let nothing = 0;
  for (let i = 0; i < arrayToSearch.length; i++) {
    if (arrayToSearch[i] === wordToCount) {
      nothing++;
    }
  }
  return nothing;
}
```

PR #3741, opened 2023-04-12
WDFT APRIL23 - GUSTAVO COUTO

Guss-Gustavo-Couto:


src/functions-and-arrays.js
```diff
@@ -52,15 +112,37 @@ const wordsUnique = [
...
+}
 
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+  function doesWordExist(wordsFind, onlyone) {
+  if (wordsFind === 0){
```
joaopdg, line 135: In this iteration you're failing only one test, and this is happening because here you're checking if an array is equal to a number `wordsFind === 0` but what you really want to check is if the length of the array is equal to zero, so what you need to do is `wordsFind,length === 0`

joaopdg (commented):
Good job on the lab Gustavo, take a look at the lab solutions to check why your function is not working on the iteration 7

PR #3743, opened 2023-04-12
[LIS-FT-APR23] roshan

roseanprime:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,101 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(num1,num2) {
+ if(num1 > num2){
+ return num1;
+} 
+  else if (num1 < num2) {
```
brunorocha20, line 6: Roshan, here, is nothing wrong with the code. A suggestion I can make you is that you can save some lines of code, as we only need one `else`. If the `num1` is bigger, we want to return `num1`, right. If the `num2` is bigger we want `num2`, ok here, however, if `num1` and `num2` are equal, we can return either `num1` or `num2`. 
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

src/functions-and-arrays.js
```diff
@@ -1,38 +1,101 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(numbers) {
+  if(numbers.length === 0){
+    return null;
+  }
+let sum = 0;
```
brunorocha20, line 75: Since you have a function that you wrote earlier that sums numbers, you can reuse it here.
Something like this:
```
  if (numbers.length === 0) return null;
  let sum = sumNumbers(numbers);
  let average = sum / numbers.length;
  return average;
}
```

brunorocha20 (commented):
Good work on the lab Roshan, just check the comments for a better understanding.

PR #3744, opened 2023-04-12
Lis-WDFT-APR-2023-Nuno-Durao

NunoDurao:


src/functions-and-arrays.js
```diff
@@ -26,7 +72,16 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(numbersArray) {
+  if (numbersArray.length === 0){
+    return null;
+  }
+  let sum = 0;
```
brunorocha20, line 79: Since you previously created a function to sum numbers, you should reuse it here instead of repeating the whole logic to sum the numbers.
```
let sum = sumNumbers(numbersArray);
return sum / numbersArray.length
```

src/functions-and-arrays.js
```diff
@@ -52,14 +107,37 @@ const wordsUnique = [
...
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(wordsArray, wordToSearch) {
+  if (wordsArray.length === 0) {
+    return null;
+  }
+  for (let i = 0; i < wordsArray.length; i++){
```
brunorocha20, line 134: Here, you can use an array method called `includes()`, that is going to check if the word exist inside the array that we iterate over. The output of that method is `true` or `false`.
So, in this case, we can just:
`return wordsArray.includes(wordToSearch)`

brunorocha20 (commented):
Good job on the lab Nuno.

PR #3775, opened 2023-04-18
[PTMEXWD042023] - Juan Pablo Maurer

churrundo:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,88 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+
+function maxOfTwoNumbers(x,y) {
+  if(x>=y){
```
ta-web-mexpt2, line 4: in this case we should name at number1 number2 or anything else except just literals

PR #3787, opened 2023-05-11
initial push

Chiara888:


TA-Remote:
Hey Carolin,  well done with the iterations you solved🙌!!!
In iteration 4 /* Level 2: Array of strings */ :
the logic of your function is correct, just few small thing missing and typo..

```
function averageWordLength(anyAverageString) { 
  if (!anyAverageString.length){
  return null
  };
  let sumArrString = 0
  for (let i = 0; i < anyAverageString[i].length; i++) { // anyAverageString.length (is the value we need)
    sumArr = anyAverageString[i].length; // sumArrString += anyA....     
  }
  return sumArrString / anyAverageString.length;
}
```
maybe you'll find the time to try with the other exercise.
keep it up💪
Alessandra

PR #3795, opened 2023-05-11
[FTRMT052023] Pedro Paulos

pedroxpaulos:


src/functions-and-arrays.js
```diff
@@ -78,7 +175,26 @@ const wordsCount = [
...
+    return 0;
+  }
+  arrWords.forEach(function(word){
+    if (word === element) {
+      counter++;
+    }
+    
+  }  );
+  if (counter === 5) {
+    return 5;
```
TA-Remote, line 190: Here you could just return the `counter`

TA-Remote (commented):
- Good work
- Pay attention to the indentation, so it's easier to read your code

PR #3796, opened 2023-05-11
RMT.WD. MAY.23. EMILIO unSolved lab

E-m-i-l-i-o:
got stuck in not understanding what the exercises asked from me. I tried until 9 non stop but my brain doesn't want to. disappointed because I understand Luis' examples and explanations. I successfully code the class exercises but it seems like the logic in the labs sentences blocks me and I get to this time of the day too tired. Besides, it seems like I forget all I think I know as soon as I start reading this iterations.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,119 +1,260 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
 
+function maxOfTwoNumbers(a, b) {
+  if (a > b) {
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Emilio!
Thanks for your feedback,
and take a good break is always the best move when your brain is fried🤪!!
Hope the example we’ve seen in class this morning clarified better what it could be one way to go with this lab. 
keep working hard and practicing as you already doing💪
Alessandra

PR #3802, opened 2023-05-11
[FTRMT052023] JuanLozano

Jualolo:


src/functions-and-arrays.js
```diff
@@ -78,7 +151,16 @@ const wordsCount = [
...
 ];
 
-function howManyTimes() {}
+function howManyTimes(wordsCount, word) {
+
+  let timesWord = 0;
+  for(let i=0; i<wordsCount.length; i++){
+    if(wordsCount[i] === word){
+      timesWord++;
+    } else{}
```
TA-Remote, line 160: Here it would be better to remove the `else`

```suggestion
    } 
```

TA-Remote (commented):
- Good work, keep it up :)

PR #3804, opened 2023-05-11
[FTRMT052023] Solideo

zendratodaeli:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,117 @@
...
+  }
+}
 
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(x) {
```
TA-Remote, line 17: Try not use `x`, `y`, `a` as variables, try to use name that will help you remember what are the arguments that you need to pass when reusing the function

PR #3805, opened 2023-05-11
doing some bonus questions and partially answer iteration 5

estevenson6:
I struggled on this exercise. Especially with unique array, iteration 5. I believe I had a solution that would work in theory. But maybe not. Partially did some bonus exercises too.


estevenson6:
I struggled on this exercise. Especially with unique array, iteration 5. I believe I had a solution that would work in theory. But maybe not. Partially did some bonus exercises too.

PR #3806, opened 2023-05-11
add iteration 1,2,3,4,5,6,7. no bonus

fndoRamos:
iteration 5 took me a lot of time to understand how to do it and I think i didn't do it correctly because I didn't use indexOf() or includes().
had no time for bonus iterations but I will try them when I have time :)


TA-Remote:
Hey Fernando!!!
very well done🙌
and about iter 5, yes you can use indexOf() or include()....but you choose the hard path, so congratulation on that🤩!!!!
great job
Alessandra

PR #3810, opened 2023-05-11
 [FTRMT052023] joaoFaria

jfaria23:
got through the first one just fine.

And was a disaster from that moment on. I got stuck on the syntax all the time and I can't figure out the logic of what a function does and the parameters and hopw to invoke it correctly and where to put the console.log

I'm sure tomorrow will be better, but I need the weekend ASAP


TA-Remote (commented):
- Good work, keep practicing :) 
- Things will get easier with practice and time, keep going :)

PR #3812, opened 2023-05-11
[FTRMT052023] Linh Nguyen

linhvnde:
Having problems with the following:

1. Bonus Calculate the sum: I threw error but jasmine didnt give me green check
2. Bonus Quest is complicated, couldnt managed to get concrete solution yet. Will work on in during weekend

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,119 +1,316 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
-
-
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
+  } else {
+    return num2;
+  }
+}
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,119 +1,316 @@
...
+      result += mixedArr[i].length;
+    } else if (mixedArr[i] === true) {
+      result += 1;
+    } else if (mixedArr[i] === false) {
+      result += 0;
+    } else if (typeof mixedArr[i] === "number") {
+      result += mixedArr[i];
+    }
+    //else if (Array.isArray(mixedArr[i]) || typeof mixedArr[i] === "object") {
+    //   throw new Error("Unsupported data type");
```
TA-Remote, line 69: The string need to be exactly that one `throw new Error("Unsupported data type sir or ma'am")`

TA-Remote (commented):
- Good work :)

PR #3815, opened 2023-05-12
[FTRMT052023] - Maria Friedemann

softcake1988:
That lab was so challenging for me. I got stuck right away and spent hours on the different iterations and still could not finish:( I think there is something in the main methodology that still has not sunk in yet.

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,20 +1,54 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(a, b) {
+  if (a>b) {
+    return a;
+  } else if (a<b) {
+    return b;
+  } else {
+    return a, b;
+  }
+}
 
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(array) {
+  if (array.length === 0) {
+    return null;
+  }
+  let longestWord = '';
+  for (let i = 0; i < array.length; i++) {
+    
+    if (array[i].length > longestWord.length) {
+      longestWord = array[i];
+    }
+  }
+  return longestWord;
+  }
 
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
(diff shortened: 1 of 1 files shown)


src/functions-and-arrays.js
```diff
@@ -1,20 +1,54 @@
...
+  }
+  let sumOfNumbers = 0;
+  for (let i= 0; i < array.length; i++) {
+    
+    sumOfNumbers += array[i];
+    
+    if (array[i] === 0) {
+      return 0;
+    } else {
+      return sumOfNumbers;
```
TA-Remote, line 46: Here that return is inside the ` for...loop`  if you move it outside and remove the `if...else` that function would work correctly

src/functions-and-arrays.js
```diff
@@ -26,13 +60,36 @@ function sum() {}
...
 const wordsArr = ['seat', 'correspond', 'linen', 'motif', 'hole', 'smell', 'smart', 'chaos', 'fuel', 'palace'];
 
-function averageWordLength() { }
+function averageWordLength(array) { 
+if (array.length === 0) {
+  return null;
+}
+let sumOfStrings = 0;
+for (let i = 0; i < array.length; i++) {
+  sumOfStrings += array[i];
```
TA-Remote, line 86: If you add that one you would get the tests passing, we want the average length of the word :) 
```suggestion
  sumOfStrings += array[i].length;
```

src/functions-and-arrays.js
```diff
@@ -78,7 +145,23 @@ const wordsCount = [
   'matter'
 ];
 
-function howManyTimes() {}
+function howManyTimes(array) {
```
TA-Remote, line 148: Here you needed another parameter to pass what was searched
```suggestion
function howManyTimes(array, search) {
```

src/functions-and-arrays.js
```diff
@@ -78,7 +145,23 @@ const wordsCount = [
...
+if (array.length === 0) {
+  return 0;
+}
+let count = 0;
+
+for (let i = 0; i < array.length; i++) {
+  for (let j = 0; j < array.length; j++) {
```
TA-Remote, line 155: and with the other parameter you won't need the second loop

```suggestion
  if (array[i] === search) {
    count ++;
  } 
```

TA-Remote (commented):
- Good work, keep practicing
- The logic will get easier with time and practice some patterns too (it didn't clicked for me too right away)

PR #3816, opened 2023-05-12
FTRMMay2023 Meissner

MarySueMe:


src/functions-and-arrays.js
```diff
@@ -1,39 +1,113 @@
...
 
+  if (arr.length === 0) {
+     return null;
+  }
+
+  let longestWord = "";
+
+  arr.forEach(function(word){
+    if (arr[i].length > longestWord.length) {
+    return longestWord;
```
TA-Remote, line 28: Here you would need to `return` outside of the `.forEach`

TA-Remote (commented):
- Good work, keep practicing :) 
- Pay attention to the spacing in your code, so it's a little bit easier to read

PR #3817, opened 2023-05-12
my code

JCouderc87:
Got stuck a lot. Daniel spent a lot of time explaining things to me which really helped. Alessandra walked me through the first one as well. Thank you for your patience with me!


TA-Remote (commented):
- Good work, keep it up and practicing, try to complete a few more iterations when you have the time
- Things will get easier with time and practice as the concepts will get more clear and the logic starts to make a little bit more sense

PR #3864, opened 2023-07-11
[LISFTJULY23] Julia Guillotte

jsguillotte:


jacobviv:
Really good job with your lab, Julia

src/functions-and-arrays.js
```diff
@@ -1,41 +1,114 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1
+  }
```
jacobviv, line 5: Here you can save some lines, the last two conditions can be solved in one line as they may both return` num2`

You could just do:

```
if (num1 > num2) {
return num1;
} else {
return num2;
}
}
```

PR #3865, opened 2023-07-11
[LISFTJULY23] Rita Saraiva

Rtzz123:
solved lab

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -1,41 +1,100 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
 
+function maxOfTwoNumbers(num1, num2) {
+  if (num1 > num2) {
+    return num1;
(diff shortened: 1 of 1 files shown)


jacobviv:
Really good work Rita, 
just remember that

`sum = sum + someValue;`

is the same as

`sum += someValue;`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,100 @@
...
+}
 
 
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function sumNumbers(numbers) {
+  let sum = 0;
```
jacobviv, line 66: you already have the function `sumNumbers()` declared in line 35, you do not have to do it again here, it is the same file

PR #3866, opened 2023-07-11
[LIS-FS-JULY23] Gabriel

Ogbera-Gabriel:


src/functions-and-arrays.js
```diff
@@ -78,8 +242,24 @@ const wordsCount = [
...
+  if (array.length === 0){
+    return 0;
+  }
+  let count = 0;
+  for (let i = 0; i < array.length; i++ ){
+    if (array[i] === word){
+      count++;
+    }
+}
+    if (count === 1) {
```
brunorocha20, line 255: Since you have your `count` variable that holds the results of your loop, you don't need to specify the values that you want to return and you could just `return count` as it would be correct and match the tests.

brunorocha20 (commented):
Great job on the lab.

PR #3867, opened 2023-07-11
[LIS-FT-JULY23] Miguel Leite

AakaiLeite:


jacobviv:
Awesome work Miguel, congratulations

src/functions-and-arrays.js
```diff
@@ -1,41 +1,123 @@
...
+function averageWordLength(wordsArray) {
+  if (!wordsArray.length) return null;
+  let averageWordLength = sumAllWords(wordsArray) / wordsArray.length
+  return averageWordLength;
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(mixArr) {
+  if (!mixArr.length) return null;
```
jacobviv, line 105: As you had already done the `function sum()` you could have reused it in here, just:

function avg (mixArr) {
  if (mixArr.length === 0) return null;
  else return sum(mixArr) / mixArr.length;
}

PR #3869, opened 2023-07-11
[LIS-FT-JULY23] Tomás Borda de Água

Vayre047:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,142 @@
...
+      sumResult += mixedArr[i].length; 
+    }else if(typeof mixedArr[i] === "number"){
+      sumResult += mixedArr[i];
+    }else if( typeof mixedArr[i] === "boolean"){
+      if(mixedArr[i] === true){
+        sumResult += 1;
+      }
+    }else{
+      throw new Error("error unsupported type");
+    }
```
jacobviv, line 135: since you already did the `function sum()` you can reuse it here:

```
function avg (mixedArr) {
  if (mixedArr.length === 0) return null;
  else return sum(mixedArr) / mixedArr.length;
}
```

PR #3870, opened 2023-07-11
[LIS-FT-JULY23] Agnes Morawska

AgnesMorawska:


src/functions-and-arrays.js
```diff
@@ -1,69 +1,176 @@
...
+// ];
+
+
+/* function uniquifyArray(myArr) {
+if(myArr.length === 0) return null;
+
+let newArr = [];
+for(let i=0; i< myArr.length; i++) {
+  if(!newArr.includes(myArr[i])){
+newArr.push(arr[i])
```
brunorocha20, line 138: Here you had everything correct, except that when you are pushing the element for the `newArr`, you misspelled `myArr[i]`.

brunorocha20 (commented):
Good job on the lab Agnes.

PR #3871, opened 2023-07-11
[LIS-FT-JULY23] André_Pinto

Andre-ginkgo:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,144 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
-
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2){
+  return number1;
+  }
+  else if (number1 < number2) {
```
brunorocha20, line 6: André, here, is nothing wrong with the code. A suggestion I can make you is that you can save some lines of code, as we only need one `else`. If the `num1` is bigger, we want to return `num1`, right. If the `num2` is bigger we want `num2`, ok here, however, if `num1` and `num2` are equal, we can return either `num1` or `num2`. 
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

brunorocha20 (commented):
Good job on the lab, André.

src/functions-and-arrays.js
Andre-ginkgo: Oh, after seeing it it's much simpler! Thank you!
I'm going to try and repeat some of the exercises from some labs on the weekend so I'll try to remember the relation between conditions!
Thank you!

PR #3874, opened 2023-07-11
[LIS-FT-JULY23] Manuel_Segarra

Segarra99:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,145 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+function maxOfTwoNumbers(number1, number2) {
+  if (number1 > number2) {
+    return number1;
+  }
```
jacobviv, line 5: As you can return `number2` in case `number1` and `number2` are the same you could do this:

```
  if (num1 > num2) return num1;
  else return num2;
}
```

src/functions-and-arrays.js
```diff
@@ -1,41 +1,145 @@
...
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  let longestWord = ""
+  if (!arr.length) {
+    return null;
+  }
+  else if (arr.length === 1) {
+    return arr[0];
+  }
```
jacobviv, line 26: Here you could have defined `longestWord` as the first element and avoid lines 24 and 25

` let longestWord = wordsArr[0];`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,145 @@
...
+    if (typeof arr[i] === "number") {
+      sum += arr[i];
+    }
+    else if (typeof arr[i] === "string") {
+      sum += arr[i].length;
+    }
+    else if (typeof arr[i] === "boolean") {
+      if (arr[i] === true) {
+        sum += 1;
+      }
```
jacobviv, line 68: Also `true` and `false` equals to `1` and `0`, so

`else if (typeof arr[i] === "boolean") {
      sum += arr[i];
    }`

src/functions-and-arrays.js
```diff
@@ -106,7 +243,18 @@ const matrix = [
...
-function greatestProduct() {}
+function greatestProduct(arr) {
+ for(let i=0; i<arr.length; i++) {
+  for(let j=0; j<arr.length; i++) {
+    if(arr[i][j] === 1) {
+      return 1;
+    }
+    else if(arr[i][j] === 2) {
+      return 16;
+    }
```
jacobviv, line 254: And now you have to find the `greatestProduct` yet! (nice work though)

PR #3875, opened 2023-07-11
[LISFTJULY23] FILIPA FLORA 

filipaflora:


src/functions-and-arrays.js
```diff
@@ -1,43 +1,165 @@
...
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arr) {
+  let result;
+  if (arr.length === 0){
+    result = null;
+  }
+  arr.forEach(element => {
```
brunorocha20, line 107: Filipa, here, since you are reusing your `sumNumbers` function, you don't need to run a loop over the argument of the `averageNumbers` function as the `sumNumbers` will do that for you. So, in this case, you could simplify your answer by just this:
```
function averageNumbers(arr) {
  if (!arr.length) return null;

  let sum = sumNumbers(arr);
  let result = sum / arr.length;

  return result;
}
```

src/functions-and-arrays.js
```diff
@@ -52,14 +174,61 @@ const wordsUnique = [
...
 
-function uniquifyArray() {}
-
+const crabArray = ['crab', 'crab', 'crab', 'crab'];
+
+function uniquifyArray(array){
+  if(array.length === 0){
+    return null
+  }
+  for(let i = 0; i <= array.length; i++) {
```
brunorocha20, line 183: Here, you have everything correct, however, there are simpler methods for you to perform this kind of action. It is a good use of nested `for` loops, but you could also use the `.includes()` method to make it easier to write and understand.
It would be something like this:
```
  if (array.length === 0) return null;
  let uniqueArray = [];
    if (!uniqueArray.includes(array[i])) {
      uniqueArray.push(array[i]);
    }
  }
  return uniqueArray;
}
```

src/functions-and-arrays.js
```diff
@@ -78,11 +247,40 @@ const wordsCount = [
...
+    return 0
+  }
+  let result;
+  let count = 0
+
+  array.forEach(element => {
+    if(element === word){
+      result = count++
+    }
+    if (element !== word){
```
brunorocha20, line 262: Here, you only took into consideration the specific values that the Jasmine tests. And in this case, you should simplify your answers. Meaning that, if you just returned the count, the answer would always be correct and there's no need to specify that three values.
```
function howManyTimes(arrayOfWords, word) {
  let count = 0;
  if (arrayOfWords.length === 0) {
    return 0;
  } else {
    for (i = 0; i < arrayOfWords.length; i++) {
      if (word === arrayOfWords[i]) {
        count++;
      }
    }
    return count;
  }
}
```

brunorocha20 (commented):
Good job on the lab, Filipa. You can check the notes I left you for a better understanding.

PR #3876, opened 2023-07-11
 [LISFTJUL2023] Mariana Marques

Haesonni:


src/functions-and-arrays.js
```diff
@@ -1,38 +1,114 @@
...
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
+function sumNumbers(numbers) {
+let sum = 0
+numbers.forEach((number)=>{
+  sum = sum + number
+})
```
jacobviv, line 39: you already know, but:

`sum += number`

src/functions-and-arrays.js
```diff
@@ -1,38 +1,114 @@
...
+function sum(arr) {
+  let totalSum = 0;
+  for (let i = 0; i < arr.length; i++) {
+    const value = arr[i];
+    if (typeof value === 'string') {
+      totalSum += value.length;
+    } else if (typeof value === 'number') {
+      totalSum += value;
+    } else if (typeof value === 'boolean') {
+      totalSum += value ? 1 : 0;
```
jacobviv, line 57: you don't need to add the ternary operator, it should work just as

```
} else if (typeof value === 'boolean') {
      totalSum += value;
```

PR #3877, opened 2023-07-11
[LIS-FT-JULY23]mariana_silva

maryanasilva:


src/functions-and-arrays.js
```diff
@@ -1,41 +1,109 @@
...
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+if (arr.length === 0) 
+return null;
 
+let newOne = "";
```
jacobviv, line 22: It should be better if you name the variable `longestWord`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,109 @@
...
 
 // Iteration #3: Calculate the sum
 const numbers = [6, 12, 1, 18, 13, 16, 2, 1, 8, 10];
 
-function sumNumbers() {}
-
+function sumNumbers(num) {
+  let sum = 0;
+  num.forEach((num)=>{
+    sum = sum + num;
```
jacobviv, line 38: As you know you can write this line as

`sum += num;`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,109 @@
...
+}
+
 
 // Bonus - Iteration #4.1
-function avg() {}
+
+const mixedArr = [6, 12, 'miami', 1, true, 'barca', '200', 'lisboa', 8, 10];
+
+function avg(arr4) {
+  let arrAvg = 0;
```
jacobviv, line 97: In this one you can reuse your newly fixed `sum` function:

```
function avg (arr) {
  else return sum(arr) / arr.length;
}
```

src/functions-and-arrays.js
```diff
@@ -52,15 +120,34 @@ const wordsUnique = [
...
+}
 
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
-
+function doesWordExist(wordExist2, wordSearch) {
+  if (wordExist2.length === 0){
```
jacobviv, line 141: Nice function, though it is easier to read the code if you name the array and the word like this:

`function doesWordExist(wordsArr, word) {`

jacobviv:
Nice work Mariana, just please give it a thought next time when naming the variables, try to make your code more readable

PR #3879, opened 2023-07-11
[LIS-FT-JULY23] Aamir-Mushtaq

Aamir269:


src/functions-and-arrays.js
```diff
@@ -1,12 +1,34 @@
...
 
-// Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
+function findLongestWord(arr) {
+  if (arr.length === 0) {
+    return null;
+  }
+
```
brunorocha20, line 20: Here, you have your function working, because you stored the first element of the array in your variable and then you looped through the rest of the array. However, that might not be always the procedure. It is good practices if you start with an empty string and then you run your loop for the whole array.
Here's how you should do it:
```

  let longestWord = '';
    if (longestWord.length < arr[i].length) {
      longestWord = arr[i];
    }
  }
  return word;
}
```

brunorocha20 (commented):
Good job Aamir, if you have the time, try to solve the iterations you couldn't do as they are good for you to practice.

PR #3880, opened 2023-07-11
[LIS-FT-JULY23] Bruno Cunha

brcunha3000:


src/functions-and-arrays.js
```diff
@@ -1,43 +1,223 @@
...
+> should declare a function named maxOfTwoNumbers
+> should return greater of two arguments - if the first argument greater
+> should return greater of two arguments - if the second argument greater
+> should return either arguments - if both arguments are equal
+
+*/
+function maxOfTwoNumbers(num1,num2) {
+  if (num1 > num2) {
+    return num1;
+  } else if (num1 < num2) {
```
brunorocha20, line 15: Bruno, here, is nothing wrong with the code. A suggestion I can make you is that you can save some lines of code, as we only need one `else`. If the `num1` is bigger, we want to return `num1`, right. If the `num2` is bigger we want `num2`, ok here, however, if `num1` and `num2` are equal, we can return either `num1` or `num2`. 
```
  if (num1 > num2) {
    return num1;
  } else {
    return num2;
  }
}
```

PR #3881, opened 2023-07-12
 done

MNader11:


jacobviv:
Really good work Mehdi

src/functions-and-arrays.js
```diff
@@ -1,41 +1,116 @@
...
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers(arrNum) {
+  let sum = 0;
+  if (arrNum.length === 0){
+    return null;
+  }
+  for (let i = 0; i < arrNum.length; i++){
+    sum = sum + arrNum[i]
```
jacobviv, line 73: You already did it in the previous iteration, let's do it in this one too:

`sum += arrNum[i];`

Also, try to be consistent with your use of `;`

src/functions-and-arrays.js
```diff
@@ -1,41 +1,116 @@
...
+  let sum = 0;
+  for (i=0; i < arr.length; i++) {
+    if (typeof arr[i] === "number") {
+      sum += arr[i];
+    } else if (typeof arr[i] === "string") {
+      sum += arr[i].length;
+    } else if (typeof arr[i] === "boolean") {
+      if (arr[i] === true) {
+        sum += 1;
+      }
```
jacobviv, line 108: This could be coded as:

```
} else if (typeof arr[i] === "boolean") {
        sum += arr[i];
}
```

src/functions-and-arrays.js
```diff
@@ -78,7 +174,26 @@ const wordsCount = [
...
+    }
+  }
+  
+  if (count === 1) {
+    return 1;
+  } else if (count === 5) {
+    return 5;
+  } else if (count === 0) {
+    return 0;
+  }
```
jacobviv, line 194: Why all this? Why aren't you returning just count?

`return count;`

PR #3883, opened 2023-07-12
[LISFTJULY23] Tiago Rodrigues

tiagomgrodrigues:


jacobviv:
Really good work Tiago!

src/functions-and-arrays.js
```diff
@@ -1,41 +1,155 @@
...
+  }
+
+    return sum/stringsArr.length;
+  }
+}
 
 // Bonus - Iteration #4.1
-function avg() {}
+function avg(array2) {
+
```
jacobviv, line 132: As you already did the sum() function before, you could have done this iteration like this:

```
function avg (arr) {
  else return sum(arr) / arr.length;
}
```

src/functions-and-arrays.js
```diff
@@ -78,7 +221,26 @@ const wordsCount = [
...
+    if(array[i] === word){
+      count = count + 1;
+    }
+  }
+
+  if(count === 1){
+    return 1;
+  }else if(count === 5){
+    return 5;
+  }else{
```
jacobviv, line 240: Why are you returning this numbers? you should be returning just the count variable

```
function howManyTimes(wordsArr, word) {
  let count = 0;

  for (let i = 0; i < wordsArr.length; i++){
    if (wordsArr[i] === word) {
      count++;
    }
  }

  return count;
}
```

PR #3884, opened 2023-07-12
[LISFTJULY23] Maureen Treacy

maureen-treacy:


src/functions-and-arrays.js
```diff
@@ -1,19 +1,52 @@
...
+function maxOfTwoNumbers(no1,no2) {
+  let number1 = no1
+  let number2 = no2
+
+  if(number1 > number2){
+    return number1
+  }
+  else if (number2 > number1){
+    return number2
+  }
```
jacobviv, line 11: Ok, there is no need to declare variables `number1` and `number2` since already already passed `no1` and `no2`

You could have done just like this:

```
  if (num1 > num2) return num1;
  else return num2;
}
```

This solution will return also `no2` when both arguments are equal joining the condition:
[should return either arguments - if both arguments are equal]

src/functions-and-arrays.js
```diff
@@ -23,16 +56,42 @@ function sum() {}
...
-function averageNumbers() {}
+function sumOfNumbers(numbers){
+let sum = 0;
+  numbers.forEach((number)=> {
+  sum = sum + number
+})
+  return sum;
+}
+function averageNumbers(numbers) {
+  if (numbers.length === 0) {return null};
```
jacobviv, line 71: If you write this if in a line it should be done like this:

`if (numbers.length === 0) return null;`

if not:

```
if (numbers.length === 0) {
     return null;
};
```

jacobviv:
Good work Maureen, if you have time try to finish the exercises you didn't do.
Also be careful with the use of `;`, try to be more consistent

PR #3911, opened 2023-08-03
[FTRMT072023] Beatriz

Beatriz-96:


TA-Remote:
hi @Beatriz-96
very well done with your solution🙌

just a quick suggestion: 
 for what I can see , the concept of the for loop is more then consolidate💪 
 so I'd raccomand you to start using forEach more, as you'll see it, and use it 😉 more often
 
 Anyway, good job and keep it up🙂
 
 Alessandra

PR #3919, opened 2023-08-03
Lab done

Leyberoh:


TA-Remote:
Hi @Leyberoh , 
very nice solutions, especially iter  average of array of string where you reuse the sum function, well done!!!  🙌

Just some suggestions about naming variables and parameter (you mentioned it this morning too): 
* as log as a parameter/variable “live” inside a function, you can reuse the same name elsewhere. 
* Try to choose a name that describe the kind of data the variable is storing
* Parameters can have more generic names usually you named them after the data type they are storing. (For you and other developer to have a better idea of what’s inside, we can get lost when our code become bigger and bigger): 
   ex. If know that a function should receive an array of numbers
         function(arrNum)
   ex. if you are iterating with a forEach:
        arrNum.forEach(function(number){... }
        
   by the time you'll go more and more generic just using: 'arr' (array) and 
  'el' (for element of the array) 😉
* Using plural and singular: an array containing the names of your classmates you’d use a name in plural  —> const students = [….] if you just wanna store the oldest —> const oldestStudent =  “…”

Keep it up 💪

PR #3959, opened 2023-10-12
FT LIS OCT23 João Valente

JohnPV19:
I Have no idea about 8 i dont think we have learn that yet


jacobviv (commented):
Great work João
Don't worry about iteration 8, it is a tricky one

PR #3960, opened 2023-10-12
[LIS-FT-OCT23]Marcelo_Melo

Melim10:


jacobviv (commented):
Great work Marcelo!

src/functions-and-arrays.js
```diff
@@ -52,14 +139,36 @@ const wordsUnique = [
...
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
+function doesWordExist(wordsFind, word) {
+  if(wordsFind.length===0){
+    return null
+  }else if(wordsFind.includes(word)){
+    return true
+  }else if(!(wordsFind.includes(word))){
```
jacobviv, line 168: You can simplify this last condition:
```javascript
else return false
```

src/functions-and-arrays.js
```diff
@@ -75,10 +184,29 @@ const wordsCount = [
...
+function howManyTimes(wordsCount, word) {
+  let count=0;
+  if(wordsCount.length===0){
+    return 0;
+  }
+  for (let i=0; i<wordsCount.length;i++){
+    if(wordsCount[i]===word){
+      count ++
+    }
+    else if(!(wordsCount.includes(word))){
```
jacobviv, line 200: You don't need this last condition, `count` is already initialized with `0`

PR #3973, opened 2023-10-12
WB-FT-LIS-OCT-23-Jakob_Bodendieck

Jakob-Bodendieck:
Had some trouble with Jasmin, any advice on how to deal with her is greatly appreciated.


jacobviv (commented):
Good work Jakob.
When working with Jasmine take in mind that a sintaxis error will crash your file and everything will be red.
Don't worry that you will end mastering it.

PR #3978, opened 2023-10-12
[LIS-FT-OCT23] Marina Balp

marifawellness:


jacobviv (commented):
Good work Marina

src/functions-and-arrays.js
```diff
@@ -1,38 +1,87 @@
...
 
 // Iteration #2: Find longest word
 const words = ['mystery', 'brother', 'aviator', 'crocodile', 'pearl', 'orchard', 'crackpot'];
 
-function findLongestWord() {}
-
+function findLongestWord(words) {
+  if (words.length === 0) {
+    return null
+  }
```
jacobviv, line 21: You can make this function shorter, like this:
```javascript
  if (words.length === 0) {
    return null
  }
  let longestWord = words[0];
  for (let i = 1; i < words.length; i++) {
      longestWord = words[i];
    }
  }
  return longestWord;
}
```

src/functions-and-arrays.js
```diff
@@ -1,38 +1,87 @@
...
 // Iteration #4: Calculate the average
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+function averageNumbers() {
+  if(averageNumbers.length === 0) {
+    return null
+  }
+  let total = 0;
```
jacobviv, line 65: If you are not using a variable don't forget to delete it

PR #3980, opened 2023-10-12
[LIS-FT-OCT23]filipe.dias

filipedias2000:


webLisbonTA:
Hi Filipe,
Great job on your lab!
The error of Jasmine is because you have only passed 1 argument on the function (Interation 6).

In this case you needed to compare on your loop, the word given as the second argument of the function, meaning:

```
function doesWordExist(wordsFind, word) {
  if (wordsFind.length === 0) {
    return null;
  }
  for (let i = 0; i < wordsFind.length; i++) {
    if (word === wordsFind[i]) {
      return true;
    }
  }
  return false;
} 
```

PR #3982, opened 2023-10-12
LIS FT WD OCT

mestremn28:


jacobviv (commented):
Good job with your lab Miguel

src/functions-and-arrays.js
```diff
@@ -1,41 +1,111 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+
+function maxOfTwoNumbers(num1, num2) {
+  if(num1 > num2){
+     return num1;
+  } else if(num1 < num2){
+    return num2;
+  } else if(num1t = num2){
```
jacobviv, line 8: Be careful, num1t doesn't exist

src/functions-and-arrays.js
```diff
@@ -1,41 +1,111 @@
 // Iteration #1: Find the maximum
-function maxOfTwoNumbers() {}
+
+function maxOfTwoNumbers(num1, num2) {
```
jacobviv, line 3: A shorter way:

```javascript
  if(num2 > num1){
     return num2;
  } else return num1;
}
```

PR #3992, opened 2023-10-13
[LIS-FT-OCT23] Mariana Carvalho

marianacarvalho-it:


src/functions-and-arrays.js
```diff
@@ -26,7 +66,22 @@ function sum() {}
 // Level 1: Array of numbers
 const numbersAvg = [2, 6, 9, 10, 7, 4, 1, 9];
 
-function averageNumbers() {}
+if (countNumbers.length === 0){
+  return null;
+}
```
jacobviv, line 71: Mariana, this `return` outside a `function` was crashing the whole file, that's why you had everything in red at Jasmine

PR #4011, opened 2023-10-19
[FTRMT102023] Elise Jonkers

elisejonkers:
Hi Ale, 

I was paired with Marat, but he already did part of the Lab. He suggested to still do it together, but I was fine with some extra practise for myself. I had some help from Pool and Dani with iteration 3 and 5. I am happy with the outcomes, except for iteration 6 ('Find elements'). Jasmine says it's not returning true (both statements in Jasmine). Can you check my code and give me feedback on how to make it correctly? I can't find the bug in this one. Thank you so much!

### DIFF
--- a/src/functions-and-arrays.js
+++ b/src/functions-and-arrays.js
@@ -82,16 +82,30 @@ const wordsUnique = [
   'bring'
 ];
 
-function uniquifyArray() {}
-
-
+function uniquifyArray(array) {
+  if(!array.length){return null}
+  let helperArray = []
+  for (let i=0; i<array.length; i++)
+    if (helperArray.includes(array[i]) === false){
+      helperArray.push(array[i])
+    }
+    return helperArray
+  }
 
 // Iteration #6: Find elements
 const wordsFind = ['machine', 'subset', 'trouble', 'starting', 'matter', 'eating', 'truth', 'disobedience'];
 
-function doesWordExist() {}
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey @elisejonkers 
I'm happy you are happy!!! Well done with your solution!!!🙌

 README INSTRUCTIONS: Declare a function named doesWordExist
 that will take in an array of words as one argument
 and a word to search for as the other.
 Return true if the word exists in the array; otherwise, return false.

 when invoked, this function needs to receive 2 arguments:
 1. an array of strings, containg words
 2. the word (a string) you wanna check if is inside
 EX: doesWordExist(wordsFind, "matter")

```
  // pass 2 parameters
  if (!array.length) {
    return null;
  }
  // const searchForThisWord = "";
    // check if one of the words inside the array is equal to the word we are looking for
    if (array[i] === word) {
      // it will stop the the loop, and return true.
      return true; 
    }
  }
  // all the other cases (no words matched) will return false
  return false;
}
```

 the shortest and the cleanest:
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes
```
function doesWordExist(wordsArr, word) {
   if (wordsArr.length === 0) {
     return null;
   } 
   return wordsArr.includes(word);
 }
```

Ale🙂

PR #4015, opened 2023-10-19
[RMT-FTWD-OCT23] Gustavo Fonte , Achraf

gusfonte99:
Did most of the iterations.

it. 6 and 7 were the most dificult in my opinion but also we were very tired at that point so we couldn't think straight.

Overall i think i understand the concepts but need more practice.


Achrafquourchi:
We've gone through most of the iterations. In my view, stages 6 and 7 proved to be the toughest, likely because fatigue was setting in, making it difficult to think clearly. In general, we believe we grasp the concepts, but we require more practice to fully master them.

